PR #6, opened 2019-09-13
Update protocols to use paths

bclayman:
We can't easily serialize Dask DataFrames to parquet using file objects.
So this commit updates each protocol's read and write methods to use
paths instead of file objects.

### DIFF
--- a/bionic/tokenization.py
+++ b/bionic/tokenization.py
@@ -6,9 +6,11 @@ from __future__ import absolute_import
 from __future__ import division
 
 from future import standard_library
+from pathlib2 import Path
 standard_library.install_aliases() # NOQA
 from builtins import str, chr, range
-from io import BytesIO
+import shutil
+import tempfile
 
 from .util import hash_to_hex
 
@@ -53,6 +55,7 @@ def clean_str(string):
 HASH_LEN = 5
 
 
+# TODO: add optional directory parameter for where to write/read from
 def tokenize(value, serialize_func=None):
     '''
     Convert an arbitrary value to a nice, unique string that could be used as a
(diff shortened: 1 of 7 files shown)


bionic/tokenization.py
```diff
@@ -53,6 +55,7 @@ def clean_str(string):
 HASH_LEN = 5
 
 
+# to-do: add optional directory parameter for where to write/read from
```
jqmp: nit: the convention in the rest of the project is `TODO`.  The only reason I care is that it makes it easier to search for all the outstanding `TODO`s if we use the same convention everywhere.

bionic/tokenization.py
```diff
@@ -62,10 +65,17 @@ def tokenize(value, serialize_func=None):
     be appended.
     '''
 
+    # to-do: clean up temp_dir after use
```
jqmp: This one is done, right?

bionic/tokenization.py
```diff
@@ -62,10 +65,17 @@ def tokenize(value, serialize_func=None):
...
+    # to-do: update docs
```
jqmp: We should probably do this one in this PR, to keep the docs in a consistent state.

In particular, I think the file you want to change is `docs/api/protocols.rst`, which has an example protocol with `read` and `write` methods.

tests/test_tokenize.py
```diff
@@ -33,8 +33,13 @@ def __init__(self, x, y):
         self.y = y
 
 
+def serialize(value, path):
```
jqmp: super nit: Since this is a top-level function, maybe a slightly more specific name, like `serialize_to_path` or something?

bionic/tokenization.py
bclayman: Yep, removed the comment!

bionic/tokenization.py
bclayman: Updated!

bionic/tokenization.py
bclayman: Sounds good, updated!

tests/test_tokenize.py
bclayman: Sure thing, just updated.

bionic/tokenization.py
jqmp: Excellent, thank you!

docs/api/protocols.rst
```diff
@@ -61,11 +61,11 @@ API changes may break your implementation.)
             """Returns the extension that persisted files should end with."""
             raise NotImplementedError()
 
-        def write(self, value, file_):
+        def write(self, value, path):
             """Write the contents of ``value`` to file object ``file_``."""
```
jqmp: Oh, we should change "file object ``file_``" to "`path`" in both of these, too.

PR #10, opened 2019-09-18
Bclayman/add dask protocol

bclayman:
This PR:
* Updates protocols to use `Path` objects and how `cache.py` makes use of a protocol's `read` and `write` methods
* Adds a Dask protocol, so users can now decorate with `@dask`, which will serialize to Parquet.  We don't currently support Categorical columns but may in the future.

### DIFF
--- a/bionic/cache.py
+++ b/bionic/cache.py
@@ -17,6 +17,7 @@ import yaml
 import six
 from pathlib2 import Path, PurePosixPath
 
+from bionic.exception import UnsupportedSerializedValueError
 from .datatypes import Result
 from .util import (
     check_exactly_one_present, hash_to_hex, get_gcs_client_without_warnings)
@@ -176,6 +177,8 @@ class Translator(object):
 
             try:
                 value = self._query.protocol.read(value_path, extension)
+            except UnsupportedSerializedValueError:
+                raise
             except Exception as e:
                 raise InvalidCacheStateError(
                     "Unable to load value %s due to %s: %s" % (
--- a/bionic/extras.py
+++ b/bionic/extras.py
@@ -28,6 +28,7 @@ extras['viz'] = combine(['hsluv', 'networkx', 'pydot'], extras['image'])
 extras['standard'] = combine(extras['matplotlib'], extras['viz'])
 
 extras['dill'] = ['dill']
+extras['dask'] = ['dask[dataframe]']
 extras['gcp'] = ['google-cloud-storage']
 
 extras['examples'] = combine(extras['standard'], ['scikit-learn'])
--- a/bionic/protocols.py
+++ b/bionic/protocols.py
@@ -15,12 +15,15 @@ import pickle
 import sys
 import tempfile
 import shutil
+import warnings
 
 import numpy as np
 from pyarrow import parquet, Table
 import pandas as pd
 from pathlib2 import Path
+import six
 
+from .exception import UnsupportedSerializedValueError
 from .provider import provider_wrapper, ProtocolUpdateProvider
 from .optdep import import_optional_dependency
 from .util import read_hashable_bytes_from_file_or_dir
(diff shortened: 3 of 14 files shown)


bionic/protocols.py
```diff
@@ -314,6 +317,38 @@ def write(self, array, path):
...
+        return 'dask.pq'
+
+    def validate(self, value):
+        assert isinstance(value, dd.DataFrame)
+
+    def read(self, path, extension):
+        if sys.version_info.major == 2:
+            path = str(path)
+        with warnings.catch_warnings():
+            warnings.filterwarnings('error')
```
jqmp: If we're only concerned about one particular type of warning, how about filtering for that one directly?  Something like `warnings.filterwarnings('error', message=r".*cannot\s+autodetect\s+index.*")
`?

tests/test_flow/test_protocols.py
```diff
@@ -196,6 +199,80 @@ def df():
...
+    @bn.protocol.dask
+    def df():
+        return dask_df
+
+    assert equal_frame_and_index_content(builder.build().get('df').compute(), dask_df.compute())
+    assert builder.build().get('df').compute().dtypes.to_dict() ==\
+        dask_df.compute().dtypes.to_dict()
+
+
+@pytest.mark.skipif(sys.version_info.major == 2, reason="requires python 3")
```
jqmp: Yeah, for whatever reason, this case actually works properly on Python2.

tests/test_flow/test_protocols.py
```diff
@@ -196,6 +199,80 @@ def df():
...
 
 
+def test_simple_dask_dataframe(builder):
+    df_value = df_from_csv_str('''
+    color,number
+    red,1
+    blue,2
+    green,3
+    ''')
+    dask_df = dd.from_pandas(df_value, npartitions=1)
```
jqmp: Should we have a test case with multiple partitions?

bionic/protocols.py
```diff
@@ -15,12 +15,15 @@
 import sys
 import tempfile
 import shutil
+import warnings
 
+import dask.dataframe as dd
```
jqmp: Can we make this work like `ImageProtocol`, where (a) we do an optional import of `dask.dataframe`, so things still work even when Dask is not installed, and (b) include `DaskProtocol` in `flow.DEFAULT_PROTOCOL` so that the user doesn't have to explicitly specify it?

bionic/protocols.py
```diff
@@ -314,6 +317,38 @@ def write(self, array, path):
...
+class DaskProtocol(BaseProtocol):
+    """
+    Decorator indicating that an entity's values always have the
+    ``dask.dataframe.DataFrame`` type.
+
+    These values will be serialized to .dask.pq files.
+    """
+
+    def get_fixed_file_extension(self):
+        return 'dask.pq'
```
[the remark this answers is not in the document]
bclayman: `pq.dask` sgtm!

tests/test_flow/test_protocols.py
bclayman: Definitely, adding!

bionic/extras.py
```diff
@@ -28,6 +28,7 @@ def combine(*dep_lists):
 extras['standard'] = combine(extras['matplotlib'], extras['viz'])
 
 extras['dill'] = ['dill']
+extras['dask'] = ['dask[dataframe]']
```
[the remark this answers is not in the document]
bclayman: Sg, adding!

bionic/protocols.py
bclayman: Much cleaner, adding!

bionic/protocols.py
```diff
@@ -314,6 +317,38 @@ def write(self, array, path):
...
+
+    def validate(self, value):
+        assert isinstance(value, dd.DataFrame)
+
+    def read(self, path, extension):
+        if sys.version_info.major == 2:
```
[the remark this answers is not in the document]
bclayman: Yep, we definitely can.  I'd prefer working with `Path` objects going forward, so would rather set it up like this.  Open to other suggestions if you think there are better options!

bionic/protocols.py
bclayman: Sg, thanks for the help!

bionic/cache.py
```diff
@@ -176,6 +177,8 @@ def _load_result_from_dir(
 
             try:
                 value = self._query.protocol.read(value_path, extension)
+            except DaskMultiIndexError:
```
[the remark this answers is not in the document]
bclayman: `UnsupportedSerializedValueError` has won the day here

bionic/protocols.py
```diff
@@ -314,6 +317,45 @@ def write(self, array, path):
...
 
+dd = import_optional_dependency('dask.dataframe', raise_on_missing=False)
+
+
+class DaskProtocol(BaseProtocol):
+    """
+    Decorator indicating that an entity's values always have the
+    ``dask.dataframe.DataFrame`` type.
+
+    These values will be serialized to .dask.pq files.
```
jqmp: nit: we should probably say that we're serializing to a directory.

tests/test_flow/test_protocols.py
```diff
@@ -196,6 +199,100 @@ def df():
...
+    @bn.protocol.dask
+    def df():
+        return dask_df
+
+    assert equal_frame_and_index_content(builder.build().get('df').compute(), dask_df.compute())
+    assert builder.build().get('df').compute().dtypes.to_dict() ==\
+        dask_df.compute().dtypes.to_dict()
+
+
+@pytest.mark.skipif(sys.version_info.major == 2, reason="requires python 3")
```
jqmp: nit: should we use `six.PY3` here too?

bionic/protocols.py
```diff
@@ -314,6 +317,45 @@ def write(self, array, path):
...
+        # which presumably means this value is not a dask dataframe, and hence should fail
+        # validation.
+        assert dd is not None
+        assert isinstance(value, dd.DataFrame)
+
+    def read(self, path, extension):
+        if six.PY2:
+            path = str(path)
+        with warnings.catch_warnings():
+            dd = import_optional_dependency('dask.dataframe', purpose='the @dask decorator')
```
jqmp: nit: this should probably be outside the `with` block, right?

tests/test_flow/test_protocols.py
bclayman: Definitely (assuming you mean `six.PY2`)!

bionic/protocols.py
bclayman: Adding!

bionic/protocols.py
bclayman: Yeah, prob best to have only the stuff that has to be in the `with`, changing!

PR #11, opened 2019-09-24
Janek/use c yaml dumper

jqmp:
This makes the tests run much faster.


damienrj:
Looks good, what is the before and after time for testing?

jqmp:
@damienrj On my machine, running Python 3, they went from around 12 seconds to 7 seconds.

Issue #14, opened 2019-10-03
Flow.export() doesn't support multi-valued entities


jqmp:
Currently you can retrieve in-memory values with `Flow.get`, which can optionally return multi-valued entities as a sequence; and you can retrieve serialized files with `Flow.export`, which can only return single-valued entities.  We should have a way to export serialized files for multi-valued entities as well.

In the short term, I think the best way to do this is to add another parameter to `get` which indicates that we should return a serialized file instead of an object.  Then we can deprecate `export`.

In the long term, I'd like to have a unified abstraction for retrieving entities which supports:
* single-valued and multi-valued entities
* objects, files, cloud blobs, etc.
* entities and tuples of entities
* `@gather`-type operations
* access from both `Flow.get` and from entity functions

jqmp:
This was resolved by https://github.com/square/bionic/pull/32

PR #32, opened 2019-10-15
Modify Flow.get to support multi-valued entities

bclayman:
This PR addresses this issue: https://github.com/square/bionic/issues/14

A couple of things worth noting here:
* We've created a `BlessedPath` class which wraps a `Path` and represents a path to the persisted file for a computed entity.  If the entity in question takes advantage of multiplicity, we return a list of `BlessedPath`s.  This class also exposes a `copy` method, which replaces `export`.
* In an ideal world, our `BlessedPath` class would inherit from `Path` so we get all the `Path` functionality for free.  It turns out this is a bit complicated (see here: https://codereview.stackexchange.com/questions/162426/subclassing-pathlib-path).  As a result, we've stored the `Path` as a field on `BlessedPath` as a workaround for now.
* `BlessedPath` is not a serious name and we'll find/replace with something even less silly before merging.


bionic/flow.py
```diff
@@ -844,46 +907,60 @@ def to_builder(self):
 
         return FlowBuilder._from_state(self._state)
 
-    def get(self, name, fmt=None):
+    def get(self, name, collection=None, object_type=object):
```
jqmp: I like the name change of `collection`, but it might be nice to keep the original `fmt` parameter but deprecate it (i.e., keep it working but emit a warning) -- just in case anyone has code that uses the keyword explicitly.

bionic/flow.py
```diff
@@ -844,46 +907,60 @@ def to_builder(self):
...
+        indicates how to handle them.  It can have any of the following values:
 
-        * ``object``: return a single value or throw an exception
+        * ``None``: return a single value or throw an exception
         * ``list`` or ``'list'``: return a list of values
         * ``set`` or ``'set'``: return a set of values
         * ``pandas.Series`` or ``'series'``: return a series whose index is
           the root cases distinguishing the different values
 
+        The user can specify the type of object to return in the collection
```
jqmp: We should be explicit about the complete list of acceptable value here.

bionic/flow.py
```diff
@@ -844,46 +907,60 @@ def to_builder(self):
...
 
         The value of the entity, or a collection containing its values.
         """
 
         result_group = self._deriver.derive(name)
-        if fmt is None or fmt is object:
-            if len(result_group) == 0:
+        if object_type is object:
+            results = [result.value for result in result_group]
+        elif object_type == 'BlessedPath':
```
jqmp: What do you think about supporting `Path` and `'filename'` as alternative options here?

bionic/flow.py
```diff
@@ -844,46 +907,60 @@ def to_builder(self):
...
         -------
 
         The value of the entity, or a collection containing its values.
         """
 
         result_group = self._deriver.derive(name)
-        if fmt is None or fmt is object:
-            if len(result_group) == 0:
+        if object_type is object:
+            results = [result.value for result in result_group]
```
jqmp: nit: Let's use a different name than `results` for the new values, since they're no longer `Result` objects.

bionic/flow.py
```diff
@@ -894,71 +971,11 @@ def get(self, name, fmt=None):
...
-
-        dst_file_path_str = str(dst_file_path)
-
-        if dst_file_path_str.startswith('gs:/'):
-            # The path object combines // into /, so we revert it here
-            copy_to_gcs(
-                str(src_file_path), dst_file_path_str.replace('gs:/', 'gs://'))
-        else:
-            shutil.copyfile(str(src_file_path), dst_file_path_str)
+            raise ValueError("Unrecognized format %r" % collection)
```
jqmp: Nit: we should replace "format" with something like "collection type".

bionic/flow.py
```diff
@@ -795,6 +795,69 @@ def then_set(self, name, value):
         return self
 
 
+class BlessedPath(object):
+    """
+    A wrapper for a Path object
```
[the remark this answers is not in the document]
jqmp: Who, sorry, I totally missed the comment where you explained why you didn't do this!

With that in mind, maybe it's better to name this something like `Copyable` or `Copier`, not expose the underlying path, and let the user request a `Path` instead if that's what they want.

bionic/flow.py
bclayman: Ah, so just to clarify: are you suggesting we temporarily allow both but emit a deprecation warning if `fmt` is used?  This ensures code relying on `fmt` will still work for now and we can remove in the future, fully moving to `collection`?

bionic/flow.py
bclayman: I'm open!  I'm assuming you mean changing the `elif` line to something like `elif object_type is Path or object_type == 'filename'`?

bionic/flow.py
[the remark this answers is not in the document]
bclayman: I'm on board with `FileCopier` since all it does is provide access to a `Path` and expose `copy` functionality!

bionic/flow.py
jqmp: Yep, exactly!

bionic/flow.py
jqmp: I was thinking more like adding new `elif` entries, like
```
elif object_type is Path or object_type == 'path':
    results = [result.file_path for result in result_group]
elif object_type == 'filename':
    results = [str(result.file_path) for result in result_group]
...
```

bionic/flow.py
jqmp: `values` works for me!

bionic/flow.py
```diff
@@ -795,6 +795,69 @@ def then_set(self, name, value):
         return self
 
 
+class BlessedPath(object):
```
jqmp: Oh, BTW, I wonder if this class should live in some other module, like `util.py`?

bionic/flow.py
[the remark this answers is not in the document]
bclayman: Hmm, def tricky.  I'd vote against `format` since old versions had `fmt` (now deprecated in favor of `collection`).  Let's go with `mode`!

bionic/flow.py
bclayman: Yep, definitely should.  Moving!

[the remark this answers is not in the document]
bclayman: Based on below comments, going with `FileCopier` :)

bionic/flow.py
```diff
@@ -795,6 +795,69 @@ def then_set(self, name, value):
...
+        The name of the computed entity.
+
+    src_file_path: Path
+        A path to the persisted file.
+    """
+    def __init__(self, name, src_file_path):
+        self.name = name
+        self.src_file_path = src_file_path
+
+    def copy(self, file_path=None, dir_path=None):
```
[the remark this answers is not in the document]
bclayman: To address your first comment: Yeah, definitely agree.  I think `copy` can take in a `destination` and we'll check if it's an existing directory with `os.path.isdir(destination)`.  If not, we'll treat it as a file and make the parent folders as needed.  Sound good?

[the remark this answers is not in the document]
jqmp: Sounds good!  Although it's possible (I haven't checked in detail) that `shutil.copy()` already does what you need.

bionic/util.py
```diff
@@ -312,6 +313,47 @@ def exception(self, msg, *args, **kwargs):
             logging.ERROR, msg, *args, exc_info=True, **kwargs)
 
 
+class FileCopier(object):
+    """
+    A wrapper for a Path object
```
jqmp: I think we can write a better description here.  Why does this class exist?

bionic/util.py
```diff
@@ -312,6 +313,47 @@ def exception(self, msg, *args, **kwargs):
...
+
+
+    Parameters
+    ----------
+    name: String
```
jqmp: We don't need this anymore, do we?

bionic/util.py
```diff
@@ -312,6 +313,47 @@ def exception(self, msg, *args, **kwargs):
...
+    src_file_path: Path
+        A path to the persisted file.
+    """
+    def __init__(self, name, src_file_path):
+        self.name = name
+        self.src_file_path = src_file_path
+
+    def copy(self, destination):
+        """
+        Copies file at `self.src_file_path` to `destination`
```
jqmp: nit: The reader of these docs probably doesn't care about `self.src_file_path`.  Depending on what we write for the class-level documentation, there's probably a clearer way to express this.

tests/test_flow/test_copy.py
```diff
@@ -0,0 +1,138 @@
...
+    @builder
+    @bn.protocol.dask
+    def dask_df():
+        return expected_dask_df
+
+    return builder.build()
+
+
+def test_copy_filename(flow):
+    assert pickle.loads(flow.get('f', mode='FileCopier')
```
jqmp: Sorry, I'm confused about what this is testing.  No copying is happening here, right?

bionic/flow.py
```diff
@@ -844,46 +844,76 @@ def to_builder(self):
...
 
-        * ``object``: return a single value or throw an exception
+        * ``None``: return a single value or throw an exception
         * ``list`` or ``'list'``: return a list of values
         * ``set`` or ``'set'``: return a set of values
         * ``pandas.Series`` or ``'series'``: return a series whose index is
           the root cases distinguishing the different values
 
+        The user can specify the type of object to return in the collection using the
+        ``mode`` parameter.  It can have any of the following values:
```
jqmp: Do we have tests for these other parameters?

bionic/flow.py
```diff
@@ -844,46 +844,76 @@ def to_builder(self):
...
+        indicates how to handle them.  It can have any of the following values:
 
-        * ``object``: return a single value or throw an exception
+        * ``None``: return a single value or throw an exception
         * ``list`` or ``'list'``: return a list of values
         * ``set`` or ``'set'``: return a set of values
         * ``pandas.Series`` or ``'series'``: return a series whose index is
           the root cases distinguishing the different values
 
+        The user can specify the type of object to return in the collection using the
```
jqmp: I wonder if we should be leading with the distinction between in-memory and persisted data, rather than just saying "type of object"?

tests/test_flow/test_copy.py
```diff
@@ -0,0 +1,138 @@
...
+def test_copy_file_to_local_file_using_str(flow, tmp_path):
+    file_path = tmp_path / 'data.pkl'
+    file_path_str = str(file_path)
+    flow.get('f', mode='FileCopier').copy(destination=file_path_str)
+    assert pickle.loads(file_path.read_bytes()) == 5
+
+
+@skip_unless_gcs
+def test_copy_file_to_gcs_dir(flow, tmp_path):
+    flow.get('f', mode='FileCopier').copy(destination='gs://' + GCS_TEST_BUCKET)
```
jqmp: Hmm, I realize this code was already here, so this probably shouldn't block this PR, but it'd probably be better to have separate setup and teardown steps like in `test_persistence_gcs` (see `tmp_object_path`).  Maybe we could add a TODO?

tests/test_flow/test_copy.py
```diff
@@ -0,0 +1,138 @@
...
+
+    dask_flow.get('dask_df', mode='FileCopier').copy(destination=destination)
+
+    actual = dd.read_parquet(expected_dir_path)
+    assert equal_frame_and_index_content(actual.compute(), expected_dask_df.compute())
+
+
+@skip_unless_gcs
+def test_copy_dask_to_gcs_dir(tmp_path, expected_dask_df, dask_flow):
+
```
jqmp: nit: why the extra newline here?

tests/test_flow/test_copy.py
```diff
@@ -0,0 +1,138 @@
...
+    @builder
+    def multi_entity(val):
+        return val
+
+    flow = builder.build()
+    results = flow.get('multi_entity', collection=set, mode=Path)
+    results = {pickle.loads(res.read_bytes()) for res in results}
+
+    assert results == my_set, "Expected {}, got {}".format(results, my_set)
```
jqmp: I'm not sure you need the explanatory string for this assert -- I think pytest will construct a useful message here automatically.

bionic/flow.py
```diff
@@ -894,30 +924,24 @@ def get(self, name, fmt=None):
...
                 index=index,
             )
         else:
-            raise ValueError("Unrecognized format %r" % fmt)
+            raise ValueError("Unrecognized collection type %r" % collection)
 
     # TODO Maybe this wants to be two different functions?
     def export(self, name, file_path=None, dir_path=None):
         """
         Provides access to the persisted file corresponding to an entity.
```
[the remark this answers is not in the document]
bclayman: I'm likely missing something but could you elaborate on this comment?

bionic/util.py
bclayman: Oh why not?  I think it's helpful to have a name associated with a `FileCopier` so the user can easily see which entity it refers to?

bionic/util.py
bclayman: Updating class-level docstring to be `A wrapper for a Path object, exposing a copy method that will copy the underlying file to a local or cloud destination`, so will likely change this comment to avoid referring to `self.src_file_path`.  Lmk if you disagree tho!

bionic/flow.py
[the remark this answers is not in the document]
jqmp: Sorry, what I meant was: in this diff, it looks like several whitespace lines have been removed.  Was this intentional?  (Or maybe there's something weird going on with my browser?)

tests/test_flow/test_copy.py
bclayman: Yeah, this test is not copying and unnecessary since we have all the tests for local copying anyway.  Gonna delete!

bionic/util.py
jqmp: Works for me!  Only suggestion I would have (*) would be maybe `file path` instead of `path object`?

(*) Not counting the previous 41 suggestions, obv.

bionic/util.py
jqmp: Isn't that fairly apparent from the file path itself now?

But also, doesn't this feel more like a generic utility object than something that knows about entities, etc.?

tests/test_flow/test_copy.py
bclayman: Just a mistake, deleting and doing same for extra line in `test_get_multi_value_entity`

bionic/flow.py
```diff
@@ -844,46 +844,81 @@ def to_builder(self):
...
         Returns
         -------
 
         The value of the entity, or a collection containing its values.
         """
 
         result_group = self._deriver.derive(name)
-        if fmt is None or fmt is object:
-            if len(result_group) == 0:
+        if mode is object:
```
cathyatsquare: should we handle the string version of this as well, e.g. `if mode is object or mode == 'object'`?

bionic/flow.py
```diff
@@ -844,46 +844,81 @@ def to_builder(self):
...
         * ``list`` or ``'list'``: return a list of values
         * ``set`` or ``'set'``: return a set of values
         * ``pandas.Series`` or ``'series'``: return a series whose index is
           the root cases distinguishing the different values
 
+        The user can specify the type of object (implicitly specifying in-memory vs. persisted data)
+        to return in the collection using the ``mode`` parameter.  It can have any of the
+        following values:
+        * ``object`` for a value in-memory
+        * ``FileCopier`` for a wrapper for a path to the persisted file for the computed entity
```
jqmp: Shouldn't this have quotes around it, since we expect a string?

(Sorry for missing this the first time.)

tests/test_flow/test_copy.py
```diff
@@ -0,0 +1,131 @@
...
+    red,1
+    blue,2
+    green,3
+    ''')
+    return dd.from_pandas(df_value, npartitions=1)
+
+
+@pytest.fixture(scope='function')
+def dask_flow(builder, expected_dask_df):
+
```
jqmp: nit: stray newline

bionic/flow.py
```diff
@@ -844,46 +844,81 @@ def to_builder(self):
...
+        * ``Path`` or ``'path'`` for a path to persisted file
+        * ``'filename'`` for a string representing a path to a persisted file
+
         Parameters
         ----------
 
         name: String
             The name of a entity.
```
jqmp: This predates your PR, but looks like we should s/a entity/an entity/ here.

bionic/util.py
jqmp: Bump.

tests/test_flow/test_api.py
```diff
@@ -376,6 +381,36 @@ def test_get_formats(preset_flow):
...
 
 
+def test_get_modes_exclude_object(preset_flow):
+    # TODO: should we raise a warning for ValueProvider entities when
+    # queried for a mode that implies the existence of a persisted file?
+    flow = preset_flow
+
+    for mode in [Path, 'path']:
+        y_path = flow.get('y', mode=mode)
+        assert y_path is None
```
jqmp: Hmm, is this what we want?  Wouldn't it make more sense to throw an exception here?

tests/test_flow/test_api.py
```diff
@@ -376,6 +381,36 @@ def test_get_formats(preset_flow):
...
+    assert isinstance(y_filecopier, bn.util.FileCopier)
+    assert y_filecopier.src_file_path is None
+
+    y_fxn_filecopier = flow.get('y_fxn', mode='FileCopier')
+    assert isinstance(y_fxn_filecopier, bn.util.FileCopier)
+    assert y_fxn_filecopier.src_file_path.is_file()
+
+    y_filename = flow.get('y', mode='filename')
+    assert y_filename == 'None'
+    # TODO: maybe don't stringify None if there is no filename to return?
```
jqmp: Yeah, returning a string here doesn't seem like what I'd expect as a user.  We could return `None` itself but I think an exception would make more sense.

tests/test_flow/test_api.py
```diff
@@ -376,6 +381,36 @@ def test_get_formats(preset_flow):
...
+        assert isinstance(y_fxn_path, Path)
+        assert y_fxn_path.is_file()
+
+    y_filecopier = flow.get('y', mode='FileCopier')
+    assert isinstance(y_filecopier, bn.util.FileCopier)
+    assert y_filecopier.src_file_path is None
+
+    y_fxn_filecopier = flow.get('y_fxn', mode='FileCopier')
+    assert isinstance(y_fxn_filecopier, bn.util.FileCopier)
+    assert y_fxn_filecopier.src_file_path.is_file()
```
jqmp: Is the idea that we don't need to test the actual value because we're doing the "real" test in `test_copy`?  Maybe we could do the full test here (i.e., load the value and check that it's correct) but keep the more complicated combinations in `test_copy`?

bionic/flow.py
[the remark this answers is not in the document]
cathyatsquare: 👍  adding whitespace back in

bionic/util.py
cathyatsquare: Good point! removed the `name` attribute and accompanying reference in docstring to "entity"

tests/test_flow/test_api.py
cathyatsquare: Sure, added: `pickle.loads(y_fxn_filecopier.src_file_path.read_bytes()) == 1`

tests/test_flow/test_api.py
cathyatsquare: Added the following at the top of `get` to raise a `ValueError` for this situation:
```
        result_group = self._deriver.derive(name)
        if mode is object or mode == 'object':
            values = [result.value for result in result_group]
        else:
            # all other modes expect the entity to be persisted
            result_file_paths = [result.file_path for result in result_group]

            if None in result_file_paths:
                raise ValueError(
                    "Entity %r is not persisted but persisted file is expected by mode %r"
                    % (name, mode))
```

tests/test_flow/test_api.py
jqmp: If it were me I would just copy the file (so the test matches the expected user behavior) but I think this is fine too.

jqmp:
Looks like the tests still need to be updated, but otherwise this looks pretty good to me!

PR #35, opened 2019-10-21
Cathy/yamlprotocol

cathyatsquare:
Support yaml protocol

### DIFF
--- a/bionic/protocols.py
+++ b/bionic/protocols.py
@@ -18,6 +18,7 @@ import tempfile
 import shutil
 import warnings
 
+import yaml
 import numpy as np
 from pyarrow import parquet, Table
 import pandas as pd
@@ -366,6 +367,32 @@ class DaskProtocol(BaseProtocol):
         if six.PY2:
             path = str(path)
         dd.to_parquet(df, path, write_index=True)
+
+
+class YamlProtocol(BaseProtocol):
+    """
+    Decorator indicating that an entity's values can be serialized using the
+    ``PyYAML`` library.
+
+    Parameters
+    ----------
+    **kwargs: keyword args for ``yaml.dump``
+        E.g. ``default_flow_style``, ``encoding``, etc.
+    """
+    def __init__(self, **kwargs):
+        super(YamlProtocol, self).__init__()
+        self._kwargs = kwargs
+
+    def get_fixed_file_extension(self):
+        return 'yaml'
+
+    def write(self, value, path):
+        with path.open('w') as file_:
+            yaml.dump(value, file_, **self._kwargs)
+
+    def read(self, path, extension):
+        with path.open('r') as file_:
+            return yaml.safe_load(file_)
 
 
 class CombinedProtocol(BaseProtocol):
--- a/tests/test_flow/test_protocols.py
+++ b/tests/test_flow/test_protocols.py
@@ -372,6 +372,37 @@ def test_numpy_protocol_high_d(builder):
     assert(np.array_equal(arr, desired))
 
 
+def test_yaml_protocol_list(builder):
+    @builder
+    @bn.protocol.yaml
+    def alist():
+        return ['red', 'blue']
+
+    assert builder.build().get('alist') == ['red', 'blue']
+
+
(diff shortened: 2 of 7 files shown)


bionic/protocols.py
```diff
@@ -368,6 +369,32 @@ def write(self, df, path):
...
 
 
+class YamlProtocol(BaseProtocol):
+    """
+    Decorator indicating that an entity's values can be serialized using the
+    ``PyYAML`` library.
+
+    Parameters
+    ----------
+    kwargs: keyword args for ``yaml.dump``
```
jqmp: super nit: We should probably write `` `**kwargs` `` instead of `kwargs` since these are keyword arguments.

tests/test_flow/test_protocols.py
```diff
@@ -372,6 +372,26 @@ def np_array_high_d():
     assert(np.array_equal(arr, desired))
 
 
+def test_yaml_protocol_list(builder):
+    x = ['red', 'blue']
+    @builder
+    @bn.protocol.yaml(default_flow_style=False)
```
jqmp: If `default_flow_style` is an important part of the test, should we validate that it's actually taking effect?  For example, we could export `alist` and check that the YAML looks the way we expect.  This might make sense as a separate  test.

tests/test_flow/test_protocols.py
```diff
@@ -372,6 +372,26 @@ def np_array_high_d():
     assert(np.array_equal(arr, desired))
 
 
+def test_yaml_protocol_list(builder):
+    x = ['red', 'blue']
```
jqmp: nit: Is there a better name we can use here than `x`?

tests/test_flow/test_protocols.py
```diff
@@ -372,6 +372,26 @@ def np_array_high_d():
...
+    @builder
+    @bn.protocol.yaml(default_flow_style=False)
+    def alist():
+        return x
+
+    assert builder.build().get('alist') == x
+
+
+def test_yaml_protocol_dict(builder):
+    x = {'dumbo': {'token': 'payment_token', 'signals': ['abc', 'def']}}
```
jqmp: nit: Maybe use a less company-specific test value?

tests/test_flow/test_protocols.py
```diff
@@ -372,6 +372,26 @@ def np_array_high_d():
...
+    @builder
+    @bn.protocol.yaml(encoding='utf-8')
```
jqmp: Similarly, if we want to test `encoding`, we probably want to either check that it actually took effect.

tests/test_flow/test_protocols.py
cathyatsquare: how about getting rid of `x` entirely?  it means duplicating `['red', 'blue']` (i.e. in the entity and the assertion), but seems very minor

jqmp: Yeah, I think that makes sense here.

tests/test_flow/test_protocols.py
```diff
@@ -372,6 +372,42 @@ def np_array_high_d():
...
+- blue
+"""
+    with open(str(tmpdir / 'value.yaml'), 'r') as file_:
+        observed = file_.read()
+
+    assert observed == expected
+
+
+def test_yaml_protocol_dict(builder):
+    record = {'record': {'name': 'Oscar', 'favorite_things': ['trash', 'green']}}
```
bclayman: I believe they call this "adding personality" 👍

tests/test_flow/test_protocols.py
```diff
@@ -372,6 +372,42 @@ def np_array_high_d():
...
+
+
+def test_yaml_protocol_dump_kwargs(builder, tmpdir):
+    @builder
+    @bn.protocol.yaml(default_flow_style=False)
+    def alist():
+        return ['red', 'blue']
+
+    flow = builder.build()
+    flow.export('alist', tmpdir / 'value.yaml')
```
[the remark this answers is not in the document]
cathyatsquare: cool, didn't know about `read_text` for paths

PR #42, opened 2019-10-29
(old, new) -> (left, right) in merge

juanmahvelez:
Replace "old" and "new" with "left" and "right" in flow.merge().

### DIFF
--- a/bionic/flow.py
+++ b/bionic/flow.py
@@ -161,8 +161,8 @@ class FlowState(pyrs.PClass):
     def mark_all_providers_default(self):
         state = self
 
-        new_default_names = pyrs.pset(state.providers_by_name.keys())
-        state = state.set(default_provider_names=new_default_names)
+        right_default_names = pyrs.pset(state.providers_by_name.keys())
+        state = state.set(default_provider_names=right_default_names)
 
         return state.touch()
 
@@ -487,12 +487,12 @@ class FlowBuilder(object):
         flow: Flow
             Any Bionic Flow.
 
-        keep: 'error', 'old', or 'new' (default: 'error')
+        keep: 'error', 'left', or 'right' (default: 'error')
             How to handle conflicting entity names.  Options:
 
             * 'error': throw an ``AlreadyDefinedEntityError``
-            * 'old': use the definition from this builder
-            * 'new': use the definition from ``flow``
+            * 'left': use the definition from this builder
+            * 'right': use the definition from ``flow``
 
         allow_name_match: boolean (default: False)
             Allows the incoming flow to have the same name as this builder.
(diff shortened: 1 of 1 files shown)


bionic/flow.py
```diff
@@ -487,12 +487,12 @@ def merge(self, flow, keep='error', allow_name_match=False):
...
 
-        keep: 'error', 'old', or 'new' (default: 'error')
+        keep: 'error', 'self', or 'arg' (default: 'error')
             How to handle conflicting entity names.  Options:
 
             * 'error': throw an ``AlreadyDefinedEntityError``
-            * 'old': use the definition from this builder
-            * 'new': use the definition from ``flow``
+            * 'self': use the definition from this builder
+            * 'arg': use the definition from ``flow``
```
jqmp: Should we mention the deprecated values of `old` and `new`?

bionic/flow.py
```diff
@@ -541,62 +541,70 @@ def merge(self, flow, keep='error', allow_name_match=False):
...
-                "Value of ``keep`` must be one of {'error', 'old', 'new'}; "
+                "Value of ``keep`` must be one of {'error', 'self', 'arg'}; "
                 "got %r" % keep)
 
         # For both states, check that each jointly-defined name group is kept
         # or discarded as a whole.
         for state_name, state in [
-                    ('old', old_state),
-                    ('new', new_state),
+                    ('self', old_state),
```
jqmp: WDYT about keeping `old` and `new` for the internal strings, to match `old_state` and `new_state`?  I can see why you don't want `self_is_default`, etc.

bionic/flow.py
pcallier: is it deprecated or just plain unsupported?

jqmp: I think we ought to continue supporting `old` and `new`, at least for the next few releases, just to stay backwards compatible.  Since the old tests are still passing, it looks like this code does that.

juanmahvelez: still supports `old`, the docstring doesn't encourage it

bionic/flow.py
juanmahvelez: yeah good point

bionic/flow.py
juanmahvelez: @jqmp for the tests, I'm thinking keep `test_keep_new(merge_tester)` and add `test_keep_arg(merge_tester)` to continue testing the supported `old` syntax

jqmp: Yeah, that makes sense to me.  You could also use `@pytest.mark.parametrize` to run the test for both values.

bionic/flow.py
```diff
@@ -603,8 +603,8 @@ def merge(self, flow, keep='error', allow_name_match=False):
         # For both states, check that each jointly-defined name group is kept
         # or discarded as a whole.
         for state_name, state in [
-                    ('self', old_state),
```
juanmahvelez: @jqmp I reverted the state_names to the originals, since ultimately this is just an api level change.

jqmp:
For some reason the CLA assistant wasn't accepting Juan's signature.  I don't think the CLA is required since Juan already works for Square, so I just merged it myself.

Issue #17, opened 2019-10-03
Entities should have docstrings


jqmp:
Entities should have docstrings which are visible to users of the flow object. If an entity is defined by a function, it should inherit that function's docstring; otherwise the docstring should probably be settable when the entity is declared or assigned.

jqmp:
This was addressed by https://github.com/square/bionic/pull/50.

PR #50, opened 2019-11-07
Make docstrings available for entities derived from functions

bclayman:
This PR addresses issue #17 but doesn't support `ValueProvider`s.

### DIFF
--- a/bionic/flow.py
+++ b/bionic/flow.py
@@ -20,8 +20,8 @@ from .datatypes import CaseKey, VersioningPolicy
 from .exception import (
     UndefinedEntityError, AlreadyDefinedEntityError, IncompatibleEntityError)
 from .provider import (
-    ValueProvider, multi_index_from_case_keys, as_provider, provider_wrapper,
-    AttrUpdateProvider)
+    ValueProvider, multi_index_from_case_keys, as_provider,
+    provider_wrapper, AttrUpdateProvider)
 from .deriver import EntityDeriver
 from . import decorators
 from .util import (
@@ -88,11 +88,11 @@ class FlowState(pyrs.PClass):
     def has_provider(self, name):
         return name in self.providers_by_name
 
-    def create_provider(self, name, protocol):
+    def create_provider(self, name, protocol, docstring):
         if name in self.providers_by_name:
             raise AlreadyDefinedEntityError.for_name(name)
 
-        provider = ValueProvider(name, protocol)
+        provider = ValueProvider(name, protocol, docstring)
         return self._set_provider(provider).touch()
 
     def install_provider(self, provider):
@@ -114,22 +114,21 @@ class FlowState(pyrs.PClass):
     def clear_providers(self, names):
         state = self
 
-        # Remember the original protocol for each provider.
-        protocols_by_entity_name = {}
+        # Remember the original provider attributes per entity.
+        attrs_by_entity_name = {}
         for name in names:
             if not state.has_provider(name):
                 continue
             provider = self.get_provider(name)
-            for res_name, res_protocol in zip(
-                    provider.attrs.names, provider.attrs.protocols):
-                protocols_by_entity_name[res_name] = res_protocol
+            attrs_by_entity_name[name] = provider.attrs
 
         # Delete the providers (or fail if not possible).
         state = state.delete_providers(names)
 
         # Recreate an empty version of each provider.
-        for res_name, res_protocol in protocols_by_entity_name.items():
-            state = state.create_provider(res_name, res_protocol)
+        for name, attrs in attrs_by_entity_name.items():
+            for res_name, res_protocol in zip(attrs.names, attrs.protocols):
+                state = state.create_provider(res_name, res_protocol, attrs.docstring)
 
         return state.touch()
 
@@ -230,7 +229,7 @@ class FlowBuilder(object):
         self._state = state.touch()
         return flow
 
-    def declare(self, name, protocol=None):
+    def declare(self, name, protocol=None, docstring=None):
         """
         Creates a new entity but does not assign it a value.
 
@@ -244,14 +243,20 @@ class FlowBuilder(object):
         protocol: Protocol, optional
             The entity's protocol.  The default is a smart type-detecting
             protocol.
+        docstring: String, optional
+            Description of the new entity.
         """
 
         if protocol is None:
             protocol = DEFAULT_PROTOCOL
 
-        self._state = self._state.create_provider(name, protocol)
+        self._state = self._state.create_provider(name, protocol, docstring)
 
-    def assign(self, name, value=None, values=None, protocol=None):
+    def assign(self, name,
+               value=None,
+               values=None,
+               protocol=None,
+               docstring=None):
         """
         Creates a new entity and assigns it a value.
 
@@ -270,6 +275,8 @@ class FlowBuilder(object):
         protocol: Protocol, optional
             The entity's protocol.  The default is a smart type-detecting
             protocol.
+        docstring: String, optional
+            Description of the new entity.
         """
 
         check_at_most_one_present(value=value, values=values)
@@ -284,7 +291,7 @@ class FlowBuilder(object):
 
         state = self._state
 
-        state = state.create_provider(name, protocol)
+        state = state.create_provider(name, protocol, docstring)
         for value in values:
             case_key = CaseKey([(name, value, protocol.tokenize(value))])
             state = state.add_case(name, case_key, value)
@@ -843,10 +850,23 @@ class Flow(object):
         ----------
 
         name: String
-            The name of a entity.
+            The name of an entity.
         """
 
         return self._state.get_provider(name).protocol_for_name(name)
+
+    def entity_docstring(self, name):
+        """
+        Returns the docstring for the named entity if a docstring is defined,
+        otherwise return None.
+
+        Parameters
+        ----------
+
+        name: String
+            The name of an entity.
+        """
+        return self._state.get_provider(name).attrs.docstring
 
     def to_builder(self):
         """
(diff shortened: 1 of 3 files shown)


bionic/flow.py
```diff
@@ -959,6 +959,23 @@ def get(self, name, collection=None, fmt=None, mode=object):
         else:
             raise ValueError("Unrecognized collection type %r" % collection)
 
+    def get_docstring_for_entity(self, name):
+        """
+        Returns the docstring for function underlying entity ``name``.  Only supported for
+        FunctionProviders (not ValueProviders).  If no docstring is present, returns ``None``.
```
jqmp: `FunctionProvider` and `ValueProvider` are internal classes which the user probably won't be familiar with -- we should phrase this in the same language the documentation uses.  (E.g., "entities defined with functions".)  However, probably we could just say "if one is defined", since function-provided entities may also be missing docstrings, and presumably we will add the ability to put docstrings on value-provided entities later.

bionic/flow.py
```diff
@@ -959,6 +959,23 @@ def get(self, name, collection=None, fmt=None, mode=object):
         else:
             raise ValueError("Unrecognized collection type %r" % collection)
 
+    def get_docstring_for_entity(self, name):
```
jqmp: We might as well match the naming convention of `Flow.entity_protocol`, which does something similar.

bionic/flow.py
```diff
@@ -959,6 +959,23 @@ def get(self, name, collection=None, fmt=None, mode=object):
...
+        """
+        Returns the docstring for function underlying entity ``name``.  Only supported for
```
jqmp: nit: these files all use a 79-character line length, as in PEP8.

bionic/flow.py
```diff
@@ -959,6 +959,23 @@ def get(self, name, collection=None, fmt=None, mode=object):
...
+        FunctionProviders (not ValueProviders).  If no docstring is present, returns ``None``.
+
+        Parameters
+        ----------
+
+        name: String
+            The name of an entity.
+        """
+        state = self._state
+        provider = state.providers_by_name[name]
```
jqmp: If you use `state.get_provider(name)`, it will throw a more helpful exception when the entity name doesn't exist.

bionic/flow.py
```diff
@@ -959,6 +959,23 @@ def get(self, name, collection=None, fmt=None, mode=object):
...
+        if type(provider) is ValueProvider:
```
jqmp: Again, why not just return `None` here?

bionic/flow.py
```diff
@@ -843,11 +854,40 @@ def entity_protocol(self, name):
...
+
+        Parameters
+        ----------
+
+        name: String
+            The name of an entity.
+        """
+        return self._state.get_provider(name).docstring
+
+    def set_entity_docstring(self, name, docstring):
```
jqmp: In general we don't have setters for entity properties like this.  If we did want to provide this, we'd need to make sure to do it in a way that doesn't mutate any existing flow objects, which are supposed to be immutable.  E.g.,

```
flow = builder.build()
# This would also modify the flow we just created, right?
builder.set_entity_docstring('entity', 'new docstring')
```

However, as far as I can tell there's not much reason to let people mutate properties on existing entities.  (Disagree?)  That's why we only allow flow mutation in fairly constrained ways -- and also why a lot of our internal objects are immutable.

bionic/provider.py
```diff
@@ -96,6 +98,14 @@ def protocol_for_name(self, name):
                 % (name, tuple(self.attrs.names)))
         return self.attrs.protocols[name_ix]
 
+    @property
```
jqmp: The other attributes are accessed directly (like `provider.attrs.names`, etc.) so for consistency I'd be inclined to do the same here.

bionic/flow.py
```diff
@@ -1209,6 +1249,7 @@ def __init__(self, state, _official=False):
 
         self.get = ShortcutProxy(self.get)
         self.setting = ShortcutProxy(self.setting)
+        self.entity_docstring = ShortcutProxy(self.entity_docstring)
```
jqmp: Actually I was thinking we could put this docstring on the `ShortcutProxy` itself, so that when you type `flow.get.my_entity?` or `flow.get.my_entity(<Tab>` you can see it in your notebook/IDE.

In general, I'm not sure many people will read the docstrings by explicitly calling `flow.entity_docstring()`.  However, we can attach them to places where user already knows to look for information, like on these methods, in the DAG visualization, perhaps in auto-generated Sphinx docs for a flow, etc.

bionic/flow.py
```diff
@@ -114,22 +114,25 @@ def add_case(self, name, case_key, value):
...
+        attrs_by_entity_name = {}
         for name in names:
             if not state.has_provider(name):
                 continue
             provider = self.get_provider(name)
-            for res_name, res_protocol in zip(
-                    provider.attrs.names, provider.attrs.protocols):
-                protocols_by_entity_name[res_name] = res_protocol
+
+            for res_name, res_protocol, res_docstring in zip(
```
jqmp: What happens if `names` has multiple values?

bionic/flow.py
```diff
@@ -114,22 +114,25 @@ def add_case(self, name, case_key, value):
...
+                    provider.attrs.names,
+                    provider.attrs.protocols,
+                    [provider.attrs.docstring]):
+                attrs_by_entity_name[res_name] = (res_protocol, res_docstring)
```
jqmp: Maybe it would be more straightforward to just store the entire `attrs` object.  That would make the name of this dict more precise, and would also make the code on line 135 clearer.

bionic/flow.py
```diff
@@ -843,11 +854,40 @@ def entity_protocol(self, name):
         ----------
 
         name: String
-            The name of a entity.
+            The name of an entity.
```
jqmp: 😃

bionic/flow.py
```diff
@@ -843,11 +854,40 @@ def entity_protocol(self, name):
...
         """
 
         return self._state.get_provider(name).protocol_for_name(name)
 
+    def entity_docstring(self, name):
+        """
+        Returns the docstring for entity ``name`` if a docstring is defined,
```
jqmp: nit: "for entity `name`" reads a little weird to me.  WDYT about "for the named entity"?

bionic/provider.py
```diff
@@ -142,7 +152,7 @@ def __init__(self, wrapped_provider, attr_name, attr_value):
 
 
 class ProtocolUpdateProvider(WrappingProvider):
-    def __init__(self, wrapped_provider, protocol=None, protools=None):
+    def __init__(self, wrapped_provider, protocol=None, protocols=None):
```
jqmp: Nice catch!  But it looks like we don't even use this argument.  I think it's vestigial and can be removed.

bionic/flow.py
```diff
@@ -230,7 +233,7 @@ def build(self):
         self._state = state.touch()
         return flow
 
-    def declare(self, name, protocol=None):
+    def declare(self, name, protocol=None, docstring=None):
```
jqmp: Do you think it's better to say `docstring` or `doc`?  The former is clearer but the latter is shorter and still pretty clear.  Genuine question, I could go either way.  WDYT?

bionic/flow.py
bclayman: Yeah, this makes total sense, especially in light of your comments below about `ShortcutProxy`!

bionic/flow.py
bclayman: I would vote `docstring` here :) but it's an easy change either way

bionic/flow.py
bclayman: Addressing in incoming update (so we won't zip the docstring)!

bionic/flow.py
bclayman: Makes sense, we're updating!

bionic/flow.py
```diff
@@ -114,22 +114,21 @@ def add_case(self, name, case_key, value):
...
+            attrs_by_entity_name[name] = provider.attrs
 
         # Delete the providers (or fail if not possible).
         state = state.delete_providers(names)
 
         # Recreate an empty version of each provider.
-        for res_name, res_protocol in protocols_by_entity_name.items():
-            state = state.create_provider(res_name, res_protocol)
+        for name, attrs in attrs_by_entity_name.items():
+            for res_name, res_protocol in zip(attrs.names, attrs.protocols):
```
jqmp: If a provider has multiple names, will we end up processing it twice?  Will that cause `create_provider` to throw an exception the second time?

bionic/flow.py
```diff
@@ -114,22 +113,27 @@ def add_case(self, name, case_key, value):
...
         # Delete the providers (or fail if not possible).
         state = state.delete_providers(names)
 
+        seen_attrs = set()
+
         # Recreate an empty version of each provider.
-        for res_name, res_protocol in protocols_by_entity_name.items():
-            state = state.create_provider(res_name, res_protocol)
+        for name, attrs in attrs_by_entity_name.items():
+            for res_name, res_protocol in zip(attrs.names, attrs.protocols):
```
jqmp: This seems like it will work, but it's not clear to me why we need an inner loop at all.  Could we just do this?

 ```
for name, attrs in attrs_by_entity_name.items():
    protocol = attrs.protocols[attrs.names.index(name)]
    state = state.create_provider(name, protocol, attrs.docstring)
```

But if I'm missing something and we do need an inner loop, does `seen_attrs` need to contain anything but the name?  I'm not certain we want to be hashing a Protocol object....

cathyatsquare: oh my, we certainly did overcomplicate this.  Thanks, your suggestion makes sense.

PR #55, opened 2019-11-08
Update documentation for multi-val export

bclayman:
Also updates Flow#export's docstring to say it's deprecated.

### DIFF
--- a/docs/api/util.rst
+++ b/docs/api/util.rst
@@ -1,0 +1,19 @@
+====================
+Utilities
+====================
+
+FileCopier
+------------
+When called with the ``mode='FileCopier'`` argument,
+:meth:`Flow.get <bionic.Flow.get>` can return a
+:class:`FileCopier <bionic.util.FileCopier>` instance.  This is simply a
+utility class that exposes a
+:meth:`copy <bionic.util.FileCopier.copy>` method, enabling the
+user to copy files around without knowing any internal details about where
+Bionic stores them.
+
+FileCopier API
+---------------
+
+.. autoclass:: bionic.util.FileCopier
+    :members:
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -415,13 +415,20 @@ explictly specify a serialization strategy for an entity by attaching a
 
 .. _Protocol: api/protocols.rst
 
-Exporting Persisted Files
-.........................
+Retrieving Persisted Files
+..........................
+In some cases, you'll want to directly access the persisted file(s) for an
+entity rather than its in-memory representation.  (For example, if you're
+writing a paper or report, you may want to access the files containing the
+plots.) This can be achieved with the ``mode`` argument to
+:meth:`Flow.get <bionic.Flow.get>` method.  For example:
 
-In some cases, you'll want to directly access the persisted file for an entity,
-rather than its in-memory representation.  (For example, if you're writing a
-paper or report, you may want to access the files containing the plots.) This
-can be accomplished using the :meth:`Flow.export <bionic.Flow.export>` method.
+.. code-block:: python
+
+    flow = builder.build().setting('subject', 'Alice')
+    flow.get('subject', mode='path')
+
+This would return a ``Path`` object for the ``subject`` entity.
 
 Multiplicity
 ------------
--- a/bionic/flow.py
+++ b/bionic/flow.py
@@ -961,7 +961,8 @@ class Flow(object):
     # TODO Maybe this wants to be two different functions?
     def export(self, name, file_path=None, dir_path=None):
         """
-        Provides access to the persisted file corresponding to an entity.
+        Provides access to the persisted file corresponding to an entity.  Note:
+        this method is deprecated and the same functionality is available through Flow#get.
 
         Can be called in three ways:
 
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -9,3 +9,4 @@ These are the APIs provided by Bionic.
     flow
     decorators
     protocols
+    util


docs/concepts.rst
```diff
@@ -415,13 +415,44 @@ explictly specify a serialization strategy for an entity by attaching a
 
 .. _Protocol: api/protocols.rst
 
-Exporting Persisted Files
-.........................
+Retrieving Persisted Files
```
jqmp: This prose is very nice and clear -- however, I feel like some of it can be left to other parts of the documentation.  It feels a bit like a self-contained explanation of all of `Flow.get`, whereas I think we can just focus on the key idea that it's possible to access persisted files.

In particular: at this point in the docs, the reader should already be aware of the existence of `get`, but may not have encountered multiplicity.  Also, we don't need to explain the full range of `mode`'s arguments and capabilities here -- we just need to make the reader aware of possibilities and link to `get`'s documentation, which should cover all the details.

What do you think?

docs/concepts.rst
```diff
@@ -415,13 +415,44 @@ explictly specify a serialization strategy for an entity by attaching a
...
+method. Its ``collection`` parameter specifies whether to return a single value
+or, if more than one, the wrapping data structure.  Its ``mode`` parameter
+indicates the type of object to represent each persisted file.  For example:
+
+.. code-block:: python
+
+    flow = builder.build().setting('subject', values=['Alice', 'Bob'])
+    flow.get('subject', collection=list, mode=Path)
+
+Would return a ``List[Path]``, one ``Path`` per persisted file (in this case, 2).
```
jqmp: style nit: Treating the code block as part of the sentence suggests that we shouldn't capitalize "would".  Maybe the cleanest approach is to write "This would ..."?

docs/concepts.rst
bclayman: Yeah, happy to remove the parts about `collection` and `mode`'s arguments / capabilities.  Do you think including documentation about `FileCopier` is excessive?

docs/concepts.rst
bclayman: SGTM!

docs/concepts.rst
jqmp: In this particular section of the docs, I think it can make sense to show that it's possible (like `flow.get('subject', mode='FileCopier').copy(...)`).  For the details on `FileCopier`, maybe the class should have docs in the API Reference section.

bclayman: Cool, will add.  Just to follow the naming conventions for docs in /docs/api, I'll make a `util.rst`.  Will just have `FileCopier` in it for now.

docs/concepts.rst
```diff
@@ -415,13 +415,31 @@ explictly specify a serialization strategy for an entity by attaching a
...
+
+.. code-block:: python
 
-In some cases, you'll want to directly access the persisted file for an entity,
-rather than its in-memory representation.  (For example, if you're writing a
-paper or report, you may want to access the files containing the plots.) This
-can be accomplished using the :meth:`Flow.export <bionic.Flow.export>` method.
+    file_copiers = flow.get('subject', collection=list, mode='FileCopier')
+    for file_copier in file_copiers:
+        # copy all of the entity's files to the current working dir
```
jqmp: Super super nit: Just to be consistent with the rest of the examples in these docs, I'd capitalize the first character and add a period at the end.

docs/concepts.rst
```diff
@@ -415,13 +415,31 @@ explictly specify a serialization strategy for an entity by attaching a
...
 .. _Protocol: api/protocols.rst
 
-Exporting Persisted Files
-.........................
+Retrieving Persisted Files
+..........................
+In some cases, you'll want to directly access the persisted file(s) for an
+entity rather than its in-memory representation.  (For example, if you're
+writing a paper or report, you may want to access the files containing the
+plots.) This can be achieved using the :meth:`Flow.get <bionic.Flow.get>`
```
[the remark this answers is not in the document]
bclayman: Hmm, it's a bit awkward either way, I think. It seems a bit strange to me not to include info about collection here, but then we'd have to place this after the multiplicity section. Will def make the above change, but makes me wonder if the ordering in this part of the docs warrants a minor reordering.

docs/api/util.rst
```diff
@@ -0,0 +1,19 @@
+====================
+FileCopier
+====================
+
+Introduction
+------------
+
+:meth:`FileCopier.copy <bionic.util.FileCopier.copy>` method.  This lets you
```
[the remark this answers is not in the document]
bclayman: Yeah, that's a good point.  Will update!

docs/api/util.rst
```diff
@@ -0,0 +1,19 @@
+====================
+FileCopier
```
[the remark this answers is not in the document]
bclayman: Yep!

docs/api/util.rst
```diff
@@ -0,0 +1,18 @@
+====================
+Utilities
+====================
+
+FileCopier
+------------
+:meth:`Flow.get <bionic.Flow.get>` can return a
```
jqmp: nit: I might add "When called with the `mode='FileCopier'` argument", or similar.

docs/concepts.rst
```diff
@@ -415,13 +415,20 @@ explictly specify a serialization strategy for an entity by attaching a
...
+In some cases, you'll want to directly access the persisted file(s) for an
+entity rather than its in-memory representation.  (For example, if you're
+writing a paper or report, you may want to access the files containing the
+plots.) This can be achieved with the ``mode`` argument to
+:meth:`Flow.get <bionic.Flow.get>` method.  For example:
+
+.. code-block:: python
+
+    flow = builder.build().setting('subject', 'Alice')
+    flow.get('subject', mode=Path)
```
jqmp: nit: Do you think it's better to use `Path` or `'path'` in the example here?  I sliiiightly lean towards the latter but could go either way.

docs/api/util.rst
bclayman: Adding!

docs/concepts.rst
bclayman: I'm open to either, so `'path'` it is!

PR #63, opened 2019-12-11
Enable and disable caching globally

squangelo:
Add `core__persistent_cache__enabled` and `core__memory_cache__enabled` flags to globally enable/disable persistent and in-memory caching for a flow.

### DIFF
--- a/bionic/deriver.py
+++ b/bionic/deriver.py
@@ -38,6 +38,8 @@ class EntityDeriver(object):
         # This state allows us to do full resolution for external callers.
         self._is_ready_for_full_resolution = False
         self._persistent_cache = None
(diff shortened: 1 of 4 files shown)


squangelo:
Disabling in-memory caching seems to work, but if I add `self._memory_cache_enabled` also on line 529 of provider.py (which I _think_ I should do) I get errors `task_state.cache_accessors is None` in `_get_results_for_complete_task_state`.
image

namanjain:
I'm closing this PR now that we merged #181. The other PR only allows you to disable in-memory caching globally, but we can add an option to turn off persistent caching globally, too, if there is an ask for it.

PR #66, opened 2019-12-13
Bump version: 0.6.1 → 0.6.2

bclayman:


docs/conf.py
```diff
@@ -20,7 +20,7 @@
 author = u'Janek Klawe'
```
cathyatsquare: I guess we can leave this be

PR #68, opened 2019-12-18
Janek/allow multiple docstrings

jqmp:
This adds support for multiple docstrings (when using multi-output functions) and also makes the docstrings of `flow.get.ENTITY_NAME` and `flow.setting.ENTITY_NAME` a little nicer.

Finally, now that I've tried using docstrings in the `ml_workflow` example, I'm tempted to change the signature from

    builder.assign(
        'test_split_fraction', 0.3,
        docstring='Fraction of data to include in test set.')

to

    builder.assign(
        'test_split_fraction', 0.3,
        doc='Fraction of data to include in test set.')

in order to leave more space for the actual string.  @bclayman @cathyatsquare what do you think?

### DIFF
--- a/bionic/flow.py
+++ b/bionic/flow.py
@@ -88,11 +88,11 @@ class FlowState(pyrs.PClass):
     def has_provider(self, name):
         return name in self.providers_by_name
 
-    def create_provider(self, name, protocol, docstring):
+    def create_provider(self, name, protocol, doc):
         if name in self.providers_by_name:
             raise AlreadyDefinedEntityError.for_name(name)
 
-        provider = ValueProvider(name, protocol, docstring)
+        provider = ValueProvider(name, protocol, doc)
         return self._set_provider(provider).touch()
 
     def install_provider(self, provider):
@@ -129,8 +129,8 @@ class FlowState(pyrs.PClass):
         for name, attrs in attrs_by_entity_name.items():
             name_ix = attrs.names.index(name)
             protocol = attrs.protocols[name_ix]
-            docstring = attrs.docstrings[name_ix]
-            state = state.create_provider(name, protocol, docstring)
+            doc = attrs.docs[name_ix]
+            state = state.create_provider(name, protocol, doc)
 
         return state.touch()
 
@@ -231,7 +231,7 @@ class FlowBuilder(object):
         self._state = state.touch()
         return flow
 
-    def declare(self, name, protocol=None, docstring=None):
+    def declare(self, name, protocol=None, doc=None, docstring=None):
         """
         Creates a new entity but does not assign it a value.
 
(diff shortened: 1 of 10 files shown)


example/ml_workflow.py
```diff
@@ -15,15 +15,26 @@
 builder = bn.FlowBuilder('ml_workflow')
 
 # Define some basic parameters.
-builder.assign('random_seed', 0)
-builder.assign('test_split_fraction', 0.3)
-builder.assign('hyperparams_dict', {'C': 1})
-builder.assign('feature_inclusion_regex', '.*')
+builder.assign(
+    'random_seed', 0,
+    docstring='Arbitrary seed for all random decisions in the flow.')
```
cathyatsquare: I think `doc` instead of `docstring` is fine if that's how you'd like to go!

bionic/flow.py
```diff
@@ -704,15 +707,48 @@ def __call__(self, func_or_provider):
...
+                got {len(names)} names {tuple(names)!r} and
+                {len(protocols)} protocols {tuple(protocols)!r}'''))
+        if len(provider.attrs.docstrings) != len(provider.attrs.names):
+            names = provider.attrs.names
+            docstrings = provider.attrs.docstrings
+            message = oneline(f'''
+                Number of docstrings must match the number of names;
+                got {len(names)} names {tuple(names)!r} and
+                {len(docstrings)} docstrings {tuple(docstrings)!r}''')
+            if len(provider.attrs.docstrings) == 1:
```
cathyatsquare: wow, pretty subtle handling for backwards compatibility

bionic/flow.py
```diff
@@ -1271,7 +1335,25 @@ def __dir__(self):
...
+            self._docstring_prefix_template.format(name=name)
+        entity_docstring = self._flow.entity_docstring(name)
+        if entity_docstring is None:
+            main_docstring = doc_prefix + "."
+        else:
+            clean_entity_docstring = dedent(entity_docstring).strip()
+            main_docstring = f"{doc_prefix}:\n{clean_entity_docstring}"
+        method_name = self._wrapped_method.__name__
+
+        partial.__doc__ = (
```
cathyatsquare: The template looks great! 

nit: I confess it is hard for me to understand what the intention is just reading this code, so it might not hurt to make that extremely obvious to mere mortals with a test for the `help` output like https://github.com/square/bionic/blob/master/tests/test_flow/test_api.py#L652

jqmp: Thanks, great idea! I'll do that.

jqmp: I added slightly more detailed tests and some comments here. Did that help?

example/ml_workflow.py
jqmp: Done!

tests/test_util.py
```diff
@@ -96,3 +96,60 @@ def test_oneline():
 
        three
        ''') == 'one two three'
+
+
+# These functions are not in util.py but it's convenient to test them here too.
+def test_longest_regex_prefix():
```
bclayman: super thorough!

tests/helpers.py
```diff
@@ -114,3 +115,129 @@ def write(self, value, path):
 
     def read(self, path, extension):
         return float(path.read_bytes())
+
+
+def assert_re_matches(regex, string, flags=0):
```
bclayman: this looks really great!

docs/concepts.rst
```diff
@@ -145,6 +145,47 @@ interpreted and converted to an entity (or entities).  In the `future
...
+
+Each Bionic entity can optionally have a documentation string associated with
+it.  Entities defined by functions can use the regular Python docstring
+syntax:
+
+.. code-block:: python
+
+    @builder
+    def message(greeting, subject):
+        'A nice thing to say to someone.'
```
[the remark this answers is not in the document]
jqmp: Thanks, I wasn't aware of that recommendation, but it makes sense. I've fixed the docstrings in this PR; there are probably some other ones but I'll do those in a separate PR at some point.

bionic/flow.py
frangipane: "slightly more" is an understatement :grin:

Issue #56, opened 2019-11-11
Add tooltips to DAG visualization


jqmp:
Currently the `Flow.render_dag()` method returns a PNG image, which doesn't support tooltips or any other kind of interaction.  However, it would be fairly straightforward to return an SVG image instead, in which case we could include tooltips for each node in the DAG.  These tooltips would be a good place to show the docstring of the corresponding entity (once [docstring integration](https://github.com/square/bionic/issues/17) is complete).

lexi-squid-ding:
@jqmp I started looking into this and looks like as of today, [Pillow does not support SVG](
https://github.com/python-pillow/Pillow/issues/1146). But it's straightforward to get a SVG from a pydot graph, so we can evaluate additional packages like Cairo and svglib to actually render the image. 

Let me know if you have a preference

jqmp:
Oh, good point. I'm cool with adding a dependency if necessary. That said, it seems to me that we want to support two main use cases:

    # Saving to a file.
    flow.render_dag(format='svg').save('my_dag.svg')

    # Inside a Jupyter notebook, this should render the image.
    flow.render_dag(format='svg')

I don't know offhand if those libraries provide a class that supports both of those. However, since Jupyter [knows how to render SVGs already](https://ipython.readthedocs.io/en/stable/config/integrating.html#rich-display), it would probably be easy to define our own `SvgImage` class that implements `save` and `_repr_svg_` methods. What do you think?

lexi-squid-ding:
Will look into whether either library does this natively, otherwise adding the new class makes sense to me.

PR #77, opened 2020-01-12
Show entity docstrings as SVG tooltips

lexi-squid-ding:
Addresses #56 : Allow FlowImage to be rendered as SVG and show entity docstrings as SVG tooltips. Continue to support PIL image types and PIL.show(). Save as SVG (to be viewed by system viewer) or display in IPython notebooks (Jupyter).

Testing:
* Ran `example/ml_workflow.py` and visualized DAG as both SVG and PNG.
* See Pytests

### DIFF
--- a/bionic/dagviz.py
+++ b/bionic/dagviz.py
@@ -4,8 +4,9 @@ pull in several optional dependencies as well.  The external Graphviz library
 is also required.
 """
 
+from pathlib import Path
 from collections import defaultdict
-from io import BytesIO
+from io import BytesIO, IOBase
 
 from .optdep import import_optional_dependency
 module_purpose = 'rendering the flow DAG'
@@ -14,12 +15,51 @@ pydot = import_optional_dependency('pydot', purpose=module_purpose)
 Image = import_optional_dependency('PIL.Image', purpose=module_purpose)
 
 
+class FlowImage:
+    def __init__(self, pydot_graph):
+        """
+        Given a pydot graph object, create Pillow Image or SVG represented as XML text.
+        Replicates the PIL APIs for save and show, for both PIL-supported and SVG formats.
+        """
+        self._pil_image = Image.open(BytesIO(pydot_graph.create_png()))
+        self._xml_text = pydot_graph.create_svg()
+
+    def save(self, fp, format=None, **params):
+        """
+        Save flow visualization to filename, Path, or file object. If file object is passed,
+        must also pass format. Pass additional keyword options supported by PIL using params.
+        Args:
+            fp: Filename (string), pathlib.Path object, or file object
+            format: format parameters
+            **params: additional keyword options supported by PIL
+        """
+        is_file_object = isinstance(fp, IOBase)
+        use_svg = (format == 'svg') or \
+            (format is None and not is_file_object and Path(fp).suffix == '.svg')
+        if use_svg:
+            if is_file_object:
+                fp.write(self._xml_text)
+            else:
+                with open(fp, 'wb') as file:
+                    file.write(self._xml_text)
+        else:
+            self._pil_image.save(fp, format, **params)
+
+    def show(self):
+        """Show image using PIL"""
+        self._pil_image.show()
+
+    def _repr_svg_(self):
+        """Rich display image as SVG in IPython notebook or Qt console."""
+        return str(self._xml_text)
+
+
 def hpluv_color_dict(keys, saturation, lightness):
-    '''
+    """
     Given a list of arbitary keys, generates a dict mapping those keys to a set
     of evenly-spaced, perceptually uniform colors with the specified saturation
     and lightness.
-    '''
+    """
 
     n = len(keys)
     color_strs = [
@@ -30,10 +70,10 @@ def hpluv_color_dict(keys, saturation, lightness):
 
 
 def dot_from_graph(graph, vertical=False, curvy_lines=False):
-    '''
+    """
     Given a NetworkX directed acyclic graph, returns a Pydot object which can
     be visualized using GraphViz.
-    '''
+    """
 
     dot = pydot.Dot(
         graph_type='digraph',
@@ -57,6 +97,9 @@ def dot_from_graph(graph, vertical=False, curvy_lines=False):
     def name_from_node(node):
         return graph.nodes[node]['name']
 
+    def doc_from_node(node):
+        return graph.nodes[node].get('doc')
+
     for cluster, node_list in node_lists_by_cluster.items():
         sorted_nodes = list(sorted(
             node_list, key=lambda node: graph.nodes[node]['task_ix']))
@@ -65,12 +108,16 @@ def dot_from_graph(graph, vertical=False, curvy_lines=False):
 
         for node in sorted_nodes:
             entity_name = graph.nodes[node]['entity_name']
-            subdot.add_node(pydot.Node(
+            entity_tooltip = doc_from_node(node)
+            dot_node = pydot.Node(
                 name_from_node(node),
                 style='filled',
                 fillcolor=color_strs_by_entity_name[entity_name],
                 shape='box',
-            ))
+            )
+            if entity_tooltip:
+                dot_node.set('tooltip', entity_tooltip)
+            subdot.add_node(dot_node)
 
         dot.add_subgraph(subdot)
 
@@ -84,10 +131,3 @@ def dot_from_graph(graph, vertical=False, curvy_lines=False):
             ))
 
     return dot
-
-
-def image_from_dot(dot):
-    '''
-    Given a pydot graph object, renders it into a Pillow Image.
-    '''
-    return Image.open(BytesIO(dot.create_png()))
--- a/bionic/deriver.py
+++ b/bionic/deriver.py
@@ -35,6 +35,7 @@ class EntityDeriver(object):
         self._is_ready_for_bootstrap_resolution = False
         self._task_lists_by_entity_name = None
         self._task_states_by_key = None
+        self._doc_by_entity_name = {}
 
         # This state allows us to do full resolution for external callers.
         self._is_ready_for_full_resolution = False
@@ -104,6 +105,7 @@ class EntityDeriver(object):
                     entity_name=entity_name,
                     case_key=task_key.case_key,
                     task_ix=task_ix,
+                    doc=self._doc_by_entity_name.get(entity_name),
                 )
 
                 for child_state in state.children:
@@ -192,7 +194,6 @@ class EntityDeriver(object):
     def _populate_entity_info(self, entity_name):
         if entity_name in self._task_lists_by_entity_name:
             return
-
         provider = self._flow_state.get_provider(entity_name)
 
         dep_names = provider.get_dependency_names()
@@ -218,6 +219,8 @@ class EntityDeriver(object):
         self._task_lists_by_entity_name[entity_name] = provider.get_tasks(
             dep_key_spaces_by_name,
             dep_task_key_lists_by_name)
+
+        self._doc_by_entity_name[entity_name] = provider.doc_for_name(entity_name)
 
     def _bootstrap_singleton(self, entity_name):
         result_group = self._compute_result_group_for_entity_name(
--- a/bionic/flow.py
+++ b/bionic/flow.py
@@ -1169,7 +1169,7 @@ class Flow(object):
 
     def render_dag(self, include_core=False, vertical=False, curvy_lines=False):
         """
-        Returns an image with a visualization of this flow's DAG.
+        Returns a FlowImage with a visualization of this flow's DAG.
 
         Will fail if Graphviz is not installed on the system.
         """
@@ -1178,8 +1178,7 @@ class Flow(object):
 
         graph = self._deriver.export_dag(include_core)
         dot = dagviz.dot_from_graph(graph, vertical, curvy_lines)
-        image = dagviz.image_from_dot(dot)
-        return image
+        return dagviz.FlowImage(dot)
 
     # TODO Should we offer an in-place version of this?  It's contrary to the
     # idea of an immutable API, but it might be more natural for the user, and
--- a/tests/test_dagviz.py
+++ b/tests/test_dagviz.py
@@ -1,0 +1,87 @@
+'''
+Tests for dagviz and FlowImage class.
+'''
+
+import pytest
+from xml.etree import ElementTree as ET
+import pydot
+import networkx as nx
+import bionic
+import bionic.dagviz as dagviz
+from PIL import Image
+
+
+@pytest.fixture
+def flow():
+    """Create FlowImage fixture for testing"""
+    builder = bionic.FlowBuilder('hello_world')
+    builder.assign('greeting', 'hello world', doc='a friendly greeting')
+    return builder.build()
+
+
+@pytest.fixture
+def flow_image(flow):
+    return flow.render_dag()
+
+
+def get_pydot_attributes(index, dot):
+    """Helper function to get attributes from pydot graph given index"""
+    return dot.get_subgraphs()[index].get_nodes()[0].get_attributes()
+
+
+def test_save_flowimage_file_path(tmp_path, flow_image):
+    """When a file path is given as input, and type is supported by PIL
+    check that output image format is preserved."""
+    filepath = tmp_path / "test.png"
+    flow_image.save(filepath)
+    output = Image.open(filepath)
+    assert output.format == 'PNG'
+
+
+def test_save_flowimage_file_path_svg(tmp_path, flow_image):
+    """When a file path is given as input and svg as the format"""
+    filepath = tmp_path / "test.svg"
+    flow_image.save(filepath)
+    output_text = (tmp_path / "test.svg").read_text()
+    try:
+        ET.fromstring(output_text)
+    except ET.ParseError:
+        pytest.fail("output from saving SVG to file object not well formed XML {}"
+                    .format(output_text))
+
+
+def test_save_flowimage_file_object(tmp_path, flow_image):
+    """When a file object is given as input, use PIL interface to save"""
+    with open(tmp_path / "test.png", 'wb') as file_object:
+        flow_image.save(file_object, format='png')
+    output = Image.open(tmp_path/"test.png")
+    assert output.format == 'PNG'
+
+
+def test_save_flowimage_file_object_svg(tmp_path, flow_image):
+    """When a file object is given as input and file is svg, use builtin interface to save"""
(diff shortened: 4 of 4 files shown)


bionic/dagviz.py
```diff
@@ -14,6 +14,26 @@
 Image = import_optional_dependency('PIL.Image', purpose=module_purpose)
 
 
+class SvgImage(object):
```
bclayman: Since Bionic only supports Python 3 now, I _think_ you can get away with "new-style" classes:
`class SvgImage` (inheriting from `object` implicitly).

bionic/deriver.py
```diff
@@ -56,7 +56,7 @@ def derive(self, entity_name):
         self.get_ready()
         return self._compute_result_group_for_entity_name(entity_name)
 
-    def export_dag(self, include_core=False):
+    def export_dag(self,  include_core=False):
```
bclayman: nit: delete extra space

bionic/dagviz.py
```diff
@@ -14,6 +14,26 @@
...
+        with open(fp, 'wb') as file:
+            file.write(self.xml_string)
+
+    def _repr_svg(self):
+        '''
+        Rich display SVG in IPython notebook or Qt console
+        '''
+        return self.xml_string
+
+    def show(self):
```
jqmp: Like you said, it's not clear how badly we need this. (I guess people using text-based shells like IPython might use this?) I wouldn't go out of your way to make this work.

bionic/dagviz.py
```diff
@@ -14,6 +14,26 @@
...
+        self.xml_string = xml_string
+
+    def save(self, fp):
+        '''
+        Save SVG to a file
+        '''
+        with open(fp, 'wb') as file:
+            file.write(self.xml_string)
+
+    def _repr_svg(self):
```
jqmp: I think you may need another underscore at the end, like `_repr_svg_`?

bionic/dagviz.py
```diff
@@ -57,6 +77,9 @@ def dot_from_graph(graph, vertical=False, curvy_lines=False):
     def name_from_node(node):
         return graph.nodes[node]['name']
 
+    def docstring_fron_node(node_name):
```
jqmp: nit: s/fron/from/

bionic/deriver.py
```diff
@@ -218,6 +213,7 @@ def _populate_entity_info(self, entity_name):
         self._task_lists_by_entity_name[entity_name] = provider.get_tasks(
             dep_key_spaces_by_name,
             dep_task_key_lists_by_name)
+        self._docstring_by_entity_name[entity_name] = provider.attrs.docstring
```
jqmp: I merged a PR recently that changed the single `provider.attrs.docstring` to multiple `provider.attrs.docs`, so you'll probably want to use something like `provider.doc_for_name(entity_name)` here now.

(I also changed most internal uses of `docstring` to `doc`, to make it clearer whether we're referring to a Bionic entity documentation string (`doc`), or an actual Python function docstring (`docstring`).)

bionic/flow.py
```diff
@@ -1167,19 +1167,18 @@ def name(self):
 
         return self.get('core__flow_name')
 
-    def render_dag(self, include_core=False, vertical=False, curvy_lines=False):
+    def render_dag(self, include_core=False, vertical=False, curvy_lines=False, svg_format=False):
```
jqmp: You know, thinking more about this, I wonder if the user should need to specify a format here at all. If they're rendering the output in Jupyter, then we can choose the best output for them, which is SVG -- I don't see much reason someone would want to insist on PNG in Jupyter. And if they want to write to a file, they can provide more information when they call `save(...)`. So maybe it'd be better if we returned our own image class that implements `_repr_svg_` and `save(...)` (which can handle SVG itself and forward other file formats to PIL). What do you think?

If you do decide to have a format parameter where, it might be better to have a string parameter (like `format='svg'`) in case we want to (a) introduce more formats, or more likely (b) transition to making SVG the default, which probably means going through a deprecation phase where you get a warning if you don't explicitly specify a format.

bionic/dagviz.py
```diff
@@ -14,6 +14,26 @@
 Image = import_optional_dependency('PIL.Image', purpose=module_purpose)
 
 
+class SvgImage(object):
+    def __init__(self, xml_string):
+        self.xml_string = xml_string
+
+    def save(self, fp):
+        '''
+        Save SVG to a file
```
jqmp: We should probably document what types are supported for `fp`. It looks like we support filenames and Paths, but not file objects (unlike `Image.save`)?

Also, a nit: the convention in the project is to still to complete sentences with periods in docstrings.

bionic/deriver.py
lexi-squid-ding: Saw this when I rebased on master, and will make the changes- thanks!

bionic/flow.py
lexi-squid-ding: For saving, if we allow for SVGs it's currently hard to decouple image rendering and saving. With PIL I think you can take bytes from a PNG and save as (any other supported format) but it looks like SVGs need to be created and saved specifically as SVGs in both stages. So I think the format parameter is required at both steps, not just in `.save()`. 

I tried pseudocoding several designs that might fulfill this. Haven't tested any for real yet, Let me know if I understood the intention correctly! 

1) create new image class that forwards file formats to PIL or SVG: For PIL, use the built in .save() and .show(). For SvgImage, use the methods (.save, ._repr_svg_) specific to that class. Looks kind of redundant but maybe there's some shorthand for this? 
```
class FlowImage:
    def __init__(self, pydot_graph, format=None):
        if format == 'png':
            self.image = Image(BytesIO(pydot_graph.create_png())
	else: # default
            self.image = SvgImage(pydot_graph.create_svg())
	
    def save(fp='foo/bar/baz'):
        self.image.save(fp)
	
    def show # Using any PIL format, but it'll error out on SVG
        self.image.show()
```

2) Having a FlowImage class without the additional SvgImage class, all conditional logic is handled in this class directly
```
class FlowImage:
    def __init__(self, pydot_graph, format='svg'):
        self.format = format
        self.image_data = pydot_graph.create_{format}()
    if self.format == 'svg':
        self.image = self.image_data
    else:
        self.image = Image(BytesIO(self.image_data))

    def show(self):
        if self.format == 'svg':
            # raise some warning
        else:
            self.image.show()
    def _repr_svg_(self):
        return self.image

    def save(self):
        if self.format == 'svg':
            # save self.image_data
        else:
            self.image.save()
```
3) Rewrite the preexisting `image_from_dot` to return either a PIL `Image` or `SvgImage`. This is closest to what we currently have and doesn't require a more general class that encapsulates both PIL and Svg. 
```
def image_from_dot(pydot_graph, format='svg'):
	if format == 'png':
            return Image(BytesIO(pydot_graph.create_png())
	return SvgImage(pydot_graph.create_svg())
```
Between all the above I think #3 is the simplest. What are your thoughts?

Re: Jupyter, it makes sense that SVG is the preferred format! We can introduce a check for the Python platform like [this](https://stackoverflow.com/questions/15411967/how-can-i-check-if-code-is-executed-in-the-ipython-notebook)).

docs/api/protocols.rst
```diff
@@ -22,7 +22,7 @@ entity being defined.  For example:
```
lexi-squid-ding: TODO: I don't know what this is or why it changed, but I'm reverting to master for this file 🤷‍♀

bionic/flow.py
jqmp: I think all of those are decent approaches -- the one I was thinking of was closest to 2. But it's still not clear to me that we need to pass an explicit `format` argument to the constructor, because we can figure out which format to use later depending on what the user asks for.

* `_repr_svg_`: We should use the best format available, which (thanks to you!) is SVG.
* `show`: SVG is not available so we might as well use PNG. If we implement support for SVG later I think we can safely switch to that.
* `save`: I think this is the only place where the user will have a preference, and we can figure out what it is from the arguments they pass in.

The main reason I'd like to avoid taking the `format` argument is so that users immediately get the improved SVG view without having to do anything extra. (You could make the default value `'svg'`, but then any existing code using `flow.render_dag().save('flow.png')` will break.)

What do you think?

lexi-squid-ding: 💡 I think I figured it out

bionic/deriver.py
```diff
@@ -94,18 +94,17 @@ def should_include_entity_name(name):
...
                     entity_name=entity_name, task_ix=task_ix)
-
                 graph.add_node(
                     task_key,
                     name=node_name,
                     entity_name=entity_name,
                     case_key=task_key.case_key,
                     task_ix=task_ix,
+                    # TODO: Get from provider
+                    docstring=self._docstring_by_entity_name.get(entity_name) or "",
```
jqmp: Two small nits here:

1. For consistency, let's reserve `docstring` for actual Python docstrings and use `doc` when describing entity documentation (which might or might not come from an actual docstring). This isn't obvious from the code but it's a pattern I tried to follow when I updated the docstring stuff.

2. It doesn't make much difference here, but as a general rule I'd avoid replacing `None` with `""` until the point where it's actually necessary. In this particular case, you have an attribute called `docstring`, and the entity doesn't actually have a docstring, so the most appropriate value here would seem to be `None`. When we get to the point of converting the docstring into a tooltip, then we may want to substitute in placeholder values like `""`. (Although -- is `""` better than not having a tooltip at all?)

bionic/deriver.py
```diff
@@ -219,6 +216,10 @@ def _populate_entity_info(self, entity_name):
             dep_key_spaces_by_name,
             dep_task_key_lists_by_name)
 
+        ds = provider.attrs.docs
+        if ds[0]:
```
jqmp: I know you have a `TODO` to replace this code, so maybe you already know this, but just in case: the first element of `docs` will not necessarily be the doc corresponding to `entity_name`. That's why you want to use `provider.doc_for_name(entity_name)` to select the correct one.

bionic/deriver.py
lexi-squid-ding: I moved the string handling to dagviz.py when we actually get the docs for entities- also wrote a test for it. If we want to do any additional string cleaning it could happen here too

tests/test_dagviz.py
```diff
@@ -0,0 +1,100 @@
...
+        flow_image.save('flow.svg', format='formatting options')
+        assert len(w) == 1
+
+
+@pytest.mark.parametrize("fp_input",
+                         ['flow.png',
+                          '/path/to/folder/flow.png',
+                          'filename.png',
+                          'machine_learning.jpeg'])
+@mock.patch.object(PIL.Image.Image, 'save')
```
jqmp: The use of mocks sounds like a great idea for testing `show`. For `save`, though, wouldn't it be a similar level of effort to actually write to a file or buffer and test that it actually contains a PNG/SVG? This would additionally test that (a) we passed the right arguments to `save` and (b) that `save` actually does what the user expects.

tests/test_dagviz.py
```diff
@@ -0,0 +1,100 @@
...
+    '''Create a minimal networkx DiGraph and check that docs are propagated to tooltips'''
+    G = nx.DiGraph()
+    G.add_node(0, name='foo', doc=doc, task_ix=0, entity_name='buzz')
+    dot = dagviz.dot_from_graph(G)
+    assert isinstance(dot, pydot.Dot)
+    assert dot.get_subgraphs()[0].get_nodes()[0].get_attributes()['tooltip'] == doc
+
+
+def test_missing_doc_empty_tooltip():
+    '''When doc is missing, tooltip is empty string'''
```
jqmp: Just curious: why is it better to have an empty tooltip than no tooltip?

tests/test_dagviz.py
```diff
@@ -0,0 +1,100 @@
...
+    return dagviz.FlowImage(pydot_graph)
+
+
+def test_flowimage_typing(flow_image):
+    '''Check types for attributes of flowImage'''
+    display_str = flow_image._repr_svg_()
+    try:
+        ET.fromstring(flow_image.xml_text)
+    except xml.etree.ElementTree.ParseError:
+        pytest.fail("FlowImage.xml_text not well formed XML")
```
jqmp: It might be helpful to include both the parsing exception and the XML text in the failure message here. Maybe something like this?

```
except xml.etree.ElementTree.ParseError as e:
    raise AssertionError(f"FlowImage.xml_text not well formed XML {flow_image.xml_text!r}") from e
```

bionic/dagviz.py
```diff
@@ -4,16 +4,53 @@
...
 module_purpose = 'rendering the flow DAG'
 hsluv = import_optional_dependency('hsluv', purpose=module_purpose)
 pydot = import_optional_dependency('pydot', purpose=module_purpose)
 Image = import_optional_dependency('PIL.Image', purpose=module_purpose)
 
 
+class FlowImage:
+    def __init__(self, pydot_graph):
+        '''
+        Given a pydot graph object, create Pillow Image or SVG represented as XML text.
```
jqmp: It would probably be more useful to have a class-level docstring, targeted at the end user, which explains what this class is and how you're expected to use it.

bionic/dagviz.py
```diff
@@ -4,16 +4,53 @@
...
 """
 
+import warnings
+from pathlib import Path
 from collections import defaultdict
-from io import BytesIO
+from io import BytesIO, IOBase
 
-from .optdep import import_optional_dependency
+from .optdep import import_optional_dependency, oneline
```
jqmp: super nit: let's import `oneline` from `util` since that's where it's "supposed" to be.

tests/test_dagviz.py
```diff
@@ -0,0 +1,100 @@
+'''
+Tests for dagviz and FlowImage class.
+'''
+
+import pytest
+import io
+import warnings
+import unittest.mock as mock
+import xml
```
jqmp: Is this used?

tests/test_dagviz.py
```diff
@@ -0,0 +1,100 @@
...
+from xml.etree import ElementTree as ET
+import pydot
+import networkx as nx
+import PIL
+import bionic.dagviz as dagviz
+
+
+@pytest.fixture
+def pydot_graph():
```
jqmp: Maybe it would be better to use a non-empty graph for most of the tests?

bionic/dagviz.py
```diff
@@ -4,16 +4,53 @@
...
+class FlowImage:
+    def __init__(self, pydot_graph):
+        '''
+        Given a pydot graph object, create Pillow Image or SVG represented as XML text.
+        '''
+        self.pil_image = Image.open(BytesIO(pydot_graph.create_png()))
+        self.xml_text = pydot_graph.create_svg()
+
+    def save(self, fp, format=None, **params):
+        '''Save flow visualization under the given filename. For SVG, must pass a filepath and does not
```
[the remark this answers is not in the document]
lexi-squid-ding: Will add descriptive parameters to the functions that need them, and will reformat the docstrings in `dagviz.py`, but will hold off on making formatting changes in other files until we enable `black`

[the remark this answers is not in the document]
lexi-squid-ding: The `format` parameter is specifically for formats supported by PIL. (for example: `quality`, which doesn't have a clear analog for the built in python save). I changed `format` >`pil_format` to clarify this.

Regarding saving SVG to a file object, we could probably add a 2nd format parameter like `file_format` in addition to `pil_format` to be able to specify this? I understand trying to make the API as simple as possible, but it's not clear to me how many teams are saving these visualizations to file objects? Feel free to push back though

example/ml_workflow.py
```diff
@@ -140,6 +140,10 @@ def all_hyperparams_pr_plot(gathered_frame, plt):
 # Compute and print the precision-recall dataframe.
 if __name__ == '__main__':
     bn.util.init_basic_logging()
+    # Show flow diagram
```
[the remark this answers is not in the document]
lexi-squid-ding: Removed!

tests/test_dagviz.py
lexi-squid-ding: Added string interpolation to the `pytest.fail`

bionic/dagviz.py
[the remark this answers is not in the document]
jqmp: Sorry, I'm not sure I'm following what you mean about the `format` parameter. It just tells PIL which of the supported formats you want, right? `quality` would be something you pass to `params`, which is indeed format-specific.

Given that we're using PIL as a starting point for our API (and given that we want to be backwards-compatible) it seems like the simplest API we can present is "just like PIL, but we support one extra format called `'svg'`".

In particular, our current API (which is just PIL) already supports the following:
* `image.save('myfile.png')`
* `image.save('myfile.whatever', format='png')`
* `image.save(some_file_object, format='png')`

If we're going to continue supporting all of those, then why not support all the equivalents for SVG?
* `image.save('myfile.svg')`
* `image.save('myfile.whatever', format='svg')`
* `image.save(some_file_object, format='svg')`

If we do something different, then the user has to learn and remember that there's a different set of rules for SVG vs all other formats, and what they are. In this case, that seems to me to outweigh the extra lines code needed to support PIL's conventions. (In fact, I suspect it would take even more lines of extra documentation to explain our more complex API.)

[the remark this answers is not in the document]
jqmp: This makes sense to me.

tests/test_dagviz.py
lexi-squid-ding: TODO: look into not having tooltip at all

bionic/dagviz.py
jqmp: Just to expand on this previous comment: it's more useful to have a docstring for the entire class than for the constructor, since Bionic users will never need to construct this class themselves. Maybe something like this?

```
class FlowImage:
    """
    An image visualizing a Bionic flow. This object provides a API similar to a
    Pillow Image, but supports an additional `svg` format in addition to the
    Pillow formats. You can use ``save`` to write this image to disk or
    ``show`` to display the image in a UI. This object will also be
    automatically rendered in a Jupyter notebook as an SVG image.
    """
```

tests/test_dagviz.py
```diff
@@ -0,0 +1,98 @@
...
+
+def get_pydot_attributes(index, dot):
+    """Helper function to get attributes from pydot graph given index"""
+    return dot.get_subgraphs()[index].get_nodes()[0].get_attributes()
+
+
+@pytest.mark.parametrize("fp_input",
+                         ['flow.svg', '/path/to/folder/flow.svg', 'filename.svg']
+                         )
+def test_save_flowimage_svg(fp_input, flow_image):
```
jqmp: Given the SVG tests below, do we still need this one?

PR #86, opened 2020-02-12
add persistence backward compatibility tests

namanjain:


.gitignore
```diff
@@ -90,4 +90,7 @@ bndata
 .vscode
 
 # Python virtual environment
-.venv
\ No newline at end of file
+.venv
+
+# Test data
+!tests/test_flow/test_persistence_compatability/artifacts/*/*/*.pkl
```
jqmp: I think you can also use `artifacts/**/.pkl` if you want to be less dependent on the cache's file hierarchy.

jqmp: Oh, also, nit: s/compatability/compatibility/.

tests/test_flow/test_persistence_compatability.py
```diff
@@ -0,0 +1,25 @@
+import pytest
+
+from .generate_test_compatability_cache import Harness
+
+
+@pytest.fixture
+def older_serialized_cache_harness(builder, datadir):
```
jqmp: So... at this point, is `datadir` doing anything except for copying the cached files into a temp directory? If not, it seems like it's only saving us about 1 line of code.

tests/test_flow/test_persistence_compatability.py
```diff
@@ -0,0 +1,25 @@
...
+    harness = Harness()
+    harness.builder.set('core__persistent_cache__flow_dir', datadir)
+    return harness
+
+
+# to regenerate cache, run
```
jqmp: I think we might want a bit more text here explaining what this test is doing and why someone would or wouldn't want to regenerate the cache.

Also, super super nit, but just for consistency, could we use capitalization and periods in comments?

tests/test_flow/generate_test_compatability_cache.py
```diff
@@ -0,0 +1,50 @@
+import os
```
jqmp: Like below: this script probably deserves a little bit of documentation, maybe in the form of a docstring at the top?

tests/test_flow/test_persistence_compatability.py
```diff
@@ -0,0 +1,25 @@
+import pytest
+
+from .generate_test_compatability_cache import Harness
+
+
+@pytest.fixture
+def older_serialized_cache_harness(builder, datadir):
+    harness = Harness()
+    harness.builder.set('core__persistent_cache__flow_dir', datadir)
```
jqmp: Very minor thing, but WDYT about passing the directory as an argument to `Harness` rather than updating the builder object?

tests/test_flow/generate_test_compatability_cache.py
```diff
@@ -0,0 +1,50 @@
...
+
+from ..helpers import count_calls
+
+import bionic as bn
+
+
+def get_test_dir():
```
jqmp: Since `__file__` is basically a constant, could/should this just be a top-level constant rather than a function?

tests/test_flow/test_persistence_compatability.py
namanjain: Yes but I'm also using datadir in relative cache path test so it's already checked in with bionic now

tests/test_flow/generate_test_compatability_cache.py
namanjain: Makes sense to me

tests/test_flow/test_persistence_compatability.py
namanjain: This makes sense to me. The test evolved a lot and some of the things are not really needed. I'm also going to get rid of pytest-datadir as we discussed in person last week since that also does not seem to be needed anymore.

namanjain:
Addressed all comments, PTAL

tests/test_flow/test_persistence_compatibility.py
```diff
@@ -0,0 +1,34 @@
...
+    # shutil.copytree dest should not exist
+    tmp_cache_path = tmp_path.joinpath('test_cache')
+    shutil.copytree(CACHE_TEST_DIR, tmp_cache_path)
+    harness = Harness(tmp_cache_path)
+    return harness
+
+
+# Tests caching backwards compatibility by loading and deserializaing
+# an old snapshot of the cache. Test failure indicates that the changes
+# made to the caching layer is backwards incompatible.
```
jqmp: nit: s/is/are/

tests/test_flow/test_persistence_compatibility.py
```diff
@@ -0,0 +1,34 @@
...
+# In case of a failure, either
+# a) fix the caching logic so it's backwards compatible or
+# b) update the cache schema version and generate a new cache snapshot.
```
jqmp: We should probably point out exactly where and how to update the cache schema version.

tests/test_flow/generate_test_compatibility_cache.py
```diff
@@ -0,0 +1,68 @@
...
+
+import os
+
+from ..helpers import count_calls
+
+import bionic as bn
+
+
+# the folder structure should be supported by pytest-datadir
+CACHE_TEST_DIR = os.path.dirname(__file__) + '/test_persistence_compatibility'
```
jqmp: nit: Technically it's safer to use `os.path.join(..., 'test_...')` instead of `... + '/test_...'` in case we ever want to run this on Windows.

tests/test_flow/generate_test_compatibility_cache.py
```diff
@@ -0,0 +1,68 @@
...
+# "Test data" section in .gitignore.
+
+import os
+
+from ..helpers import count_calls
+
+import bionic as bn
+
+
+# the folder structure should be supported by pytest-datadir
```
jqmp: This comment is out of date now, right?

namanjain: ah yes, removing the comment but I'm going to keep this folder structure

PR #87, opened 2020-02-12
make artifact url relative to metadata url for local files

namanjain:


bionic/cache.py
```diff
@@ -494,25 +494,47 @@ def _exact_metadata_url_for_query(self, query):
     def _load_metadata_from_url(self, url):
         try:
             metadata_yaml = self._fs.read_bytes(url).decode('utf8')
-            return ArtifactMetadataRecord.from_yaml(metadata_yaml)
+            record = ArtifactMetadataRecord.from_yaml(metadata_yaml)
```
jqmp: This is pretty reasonable. I wonder if it would be cleaner for this logic to happen inside the `ArtifactMetadataRecord` code? (Like within `from_yaml` and `from_content`?) It's a little annoying to have to pass in the metadata URL but it also keeps the record object immutable and saves us from having to remember to do this if we load a record anywhere else.

bionic/cache.py
```diff
@@ -494,25 +494,47 @@ def _exact_metadata_url_for_query(self, query):
...
-        metadata_url = self._exact_metadata_url_for_query(query)
-
         self._fs.write_bytes(
             metadata_record.to_yaml().encode('utf8'), metadata_url)
 
         return metadata_url
 
+    # when both urls start with file://, make the artifact url relative to
+    # metadata url
+    @staticmethod
```
jqmp: nit: Is there a reason to make these static methods instead of just top-level functions?

bionic/cache.py
```diff
@@ -494,25 +494,47 @@ def _exact_metadata_url_for_query(self, query):
...
+    def _absolute_or_relative_url(artifact_url, metadata_url):
```
jqmp: nit: I wonder if these would be clearer as something like `relativize_url`/`derelativize_url`? Doesn't actually make their behavior more obvious but it does call out the relationship between them. Could go either way though.

bionic/cache.py
```diff
@@ -494,25 +494,47 @@ def _exact_metadata_url_for_query(self, query):
...
         )
 
-        metadata_url = self._exact_metadata_url_for_query(query)
-
         self._fs.write_bytes(
             metadata_record.to_yaml().encode('utf8'), metadata_url)
 
         return metadata_url
 
+    # when both urls start with file://, make the artifact url relative to
```
jqmp: This would be good as a docstring.

bionic/cache.py
```diff
@@ -494,25 +494,47 @@ def _exact_metadata_url_for_query(self, query):
...
+    # metadata url
+    @staticmethod
+    def _absolute_or_relative_url(artifact_url, metadata_url):
+        if not is_file_url(artifact_url) or not is_file_url(metadata_url):
+            return artifact_url
+        return os.path.relpath(artifact_url, os.path.dirname(metadata_url))
```
jqmp: Just in case you weren't aware, you can also do this with `pathlib` Paths, but whichever way you think is cleaner is fine.

bionic/cache.py
namanjain: Good point about immutability. I missed that since immutability isn't enforced here.

bionic/cache.py
namanjain: I initially thought I'll keep everything at one place which was inventory in this case. But now that I'm moving the code around by putting the logic into metadata record, I think I'll just make them top-level functions

bionic/cache.py
namanjain: Ah I did not see `pathlib` Paths before. I tried it but it's a lot harder to use `relative_to` in `pathlib`. Even the [documentation](https://docs.python.org/3.7/library/pathlib.html#correspondence-to-tools-in-the-os-module) calls out that `os.path.relpath()` and `PurePath.relative_to()` are different enough to not consider them equivalent. So I'm going to keep it like this unless you feel strongly otherwise.

namanjain:
addressed comments and added a test

bionic/cache.py
```diff
@@ -1017,3 +1024,20 @@ def update_hash(hash_, obj):
             update_hash(hash_, value)
     else:
         raise ValueError(f"Unable to hash object {obj!r} of type {type(obj)!r}")
+
+
+def relativize_url(artifact_url, metadata_url):
+    """when both urls start with file://, make the artifact url relative to
```
jqmp: nit: Docstrings should be complete sentences (capitalized with periods).

tests/test_flow/test_relative_cache_path.py
```diff
@@ -0,0 +1,32 @@
...
+    def xy(x, y):
+        return x * y
+
+    return builder
+
+
+def test_move_cache_files(builder, datadir):
+    curDir = os.path.join(datadir, 'current')
+    newDir = os.path.join(datadir, 'new')
+    print(curDir, newDir)
```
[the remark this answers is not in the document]
namanjain: whoops, will remove it

tests/test_flow/test_relative_cache_path.py
```diff
@@ -0,0 +1,32 @@
+import os
+import pytest
+from distutils.dir_util import copy_tree
```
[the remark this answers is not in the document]
namanjain: Done

PR #88, opened 2020-02-20
use urlparse for parts instead of string manipulation

namanjain:


bionic/cache.py
```diff
@@ -746,16 +747,15 @@ def _validated_object_name_from_url(self, url):
...
-            bucket_name, object_prefix = url_parts
-
-        return bucket_name, object_prefix
+        if not is_gcs_url(url):
+            raise ValueError(f'url must have schema "{GCS_SCHEME}"')
+        result = urlparse(url)
+        object_name = result.path
+        # urlparse parses the url with a leading slash in path
+        # we don't want to write to the root dir in gcs bucket
+        if object_name.startswith('/'):
```
jqmp: I'm confused by this part. `urlparse` only includes a leading slash if there's an extra slash in the URL, right? Like `gs:///bucket/dir/data.txt`? But isn't this actually a different (and maybe invalid) URL compared to `gs://bucket/dir/data.txt`?

Also, when I run `urlparse('gs:///bucket/dir/data.txt')`, it leaves `netloc` empty, so that seems like it'd be a problem too.

bionic/cache.py
```diff
@@ -961,20 +961,33 @@ def dependencies_exactly_match(self, prov):
...
+    result = urlparse(url)
+    return result.scheme == FILE_SCHEME
+
+
+def is_gcs_url(url):
+    result = urlparse(url)
+    return result.scheme == GCS_SCHEME
+
+
+def is_absolute_url(url):
```
jqmp: If we see an unrecognized scheme here (like `s3:`), this will return `False`, right? Is that what we want?

bionic/cache.py
```diff
@@ -1028,7 +1041,7 @@ def update_hash(hash_, obj):
 
 def relativize_url(artifact_url, metadata_url):
```
jqmp: To avoid the kind of bug we had before, maybe it would be safer to only perform path operations on path objects, not on URLs? I.e., convert the input URLs to paths, perform the transformation, then convert the output back to a URL?

jqmp:
Oh, one more nit: It would be good for the commit message to include the reason for this fix.

bionic/cache.py
namanjain: I don't think that's true. IIUC, a url like `scheme://netloc/path` are absolute urls so the path can never be relative. I think that's why the url always contains a leading slash. But for gcs, it seems like Bionic assumes that the path is always relative to the cache dir. Hence I'm removing the leading slash. Maybe I can add this to the comment for clarity.

If you see the [syntax spec](https://cloud.google.com/storage/docs/gsutil#syntax), it is `gs://[BUCKET_NAME]/[OBJECT_NAME]`. So I guess having 3 slashes means there is no bucket. Is that a valid url or something you have seen before?

namanjain:
> Are there any tests we can add here to confirm that we fixed the bug? (Other than the ones requiring GCS, of course, since they don't run on Travis.)

Hmm, not sure. We'll need a cache that contains absolute url in metadata. Maybe we can add another cache in `test_persistence_compatibility`. What do you think?

bionic/cache.py
namanjain: ah right, I don't think we want that. How do you feel about throwing an error here if it's not in `SUPPORTED_SCHEMES` list?

jqmp: That sounds reasonable to me, since the set of schemes in use is under our control.

bionic/cache.py
jqmp: Sorry, as we discussed, I was confused here. I think the behavior makes sense, although the comment may be misleading.

Oh, but: if `urlparse` *always* returns a leading slash, maybe it would be better to have an `assert` rather than an `if`?

jqmp:
> Hmm, not sure. We'll need a cache that contains absolute url in metadata. Maybe we can add another cache in `test_persistence_compatibility`. What do you think?

If there's a straightforward way to add an end-to-end test, that sounds good. But otherwise we could also just add some unit tests for some of the helper functions.

(It would be nice if we could totally mock out our GCS calls, but that's probably way outside the scope of this PR.)

bionic/cache.py
```diff
@@ -746,16 +747,14 @@ def _validated_object_name_from_url(self, url):
...
-            bucket_name, = url_parts
-            object_prefix = ''
-        else:
-            bucket_name, object_prefix = url_parts
-
-        return bucket_name, object_prefix
+        if not is_gcs_url(url):
+            raise ValueError(f'url must have schema "{GCS_SCHEME}"')
+        result = urlparse(url)
+        object_name = result.path
```
jqmp: super nit: this isn't an `object_name`, it's either a `result_path` or an `object_name_with_leading_slash`. :)

bionic/cache.py
```diff
@@ -961,20 +960,42 @@ def dependencies_exactly_match(self, prov):
...
+    result = urlparse(url)
+    return result.scheme == GCS_SCHEME
+
+
+def is_absolute_url(url):
+    result = urlparse(url)
+    if not result.scheme:
+        return False
+    if result.scheme not in SUPPORTED_SCHEMES:
+        raise UnsupportedSchemeError(
```
jqmp: I think either `AssertionError` or `ValueError` would be also fine here.

bionic/cache.py
```diff
@@ -961,20 +960,42 @@ def dependencies_exactly_match(self, prov):
...
+            f'Found a URL with unsupported scheme {result.scheme}.')
```
jqmp: BTW, if you use `{result.scheme!r}`, it will put quotes around the scheme (by applying `repr`), which may make the message less confusing in some circumstances.

`Found a URL with unsupported scheme data.`
vs
`Found a URL with unsupported scheme 'data'.`

bionic/cache.py
```diff
@@ -1028,17 +1049,19 @@ def update_hash(hash_, obj):
...
     """Returns the absolute artifact url when it is relative wrt metadata url.
     Otherwise returns the original url."""
-    if os.path.isabs(artifact_url):
+    if is_absolute_url(artifact_url):
         return artifact_url
-    abspath = os.path.normpath(os.path.join(os.path.dirname(metadata_url), artifact_url))
-    # normpath removes redundant `/`. We need to add them back from the url prefix
-    return abspath.replace("file:", FILE_URL_PREFIX)
+    metadata_path = path_from_url(metadata_url)
+    abspath = os.path.normpath(metadata_path.parent.joinpath(artifact_url))
```
jqmp: I know it doesn't change the behavior, but I feel like converting `artifact_url` to `artifact_path` first would make it more obvious that this code is safe.

namanjain: sure, if it looks safer

jqmp:
🎊

PR #90, opened 2020-03-03
Use dependency output hash (if persistable) when triggering cache invalidation.

namanjain:
This prevents Bionic to recompute children tasks of an entity whose provenance
changed but output remained the same.


bionic/deriver.py
```diff
@@ -265,13 +262,18 @@ def _compute_result_group_for_entity_name(self, entity_name):
                         logged_task_keys.add(task_key)
                 continue
 
-            # First, see if we can load it from the cache.
-            self._attempt_to_load_task_state_results(state)
+            # First, see if we can compute the provenance
```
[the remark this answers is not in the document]
namanjain: done

bionic/datatypes.py
```diff
@@ -72,15 +69,33 @@ class Result(object):
     Represents one value for one entity.
     '''
     def __init__(
-            self, query, value, file_path=None):
+            self, query, value, value_hash=None, file_path=None):
         self.query = query
         self.value = value
         self.file_path = file_path
+        # (For now) Only valid when value should persist.
```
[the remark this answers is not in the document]
namanjain: - Oh yes, this applies to `file_path` too. I'll fix that. 
- I put for now since I thought we will replace it with another decorator later (hashable / digestible). I can either remove it or expand on it. 
- `present` sounds better to me too.

bionic/deriver.py
```diff
@@ -437,33 +451,35 @@ def _attempt_to_load_task_state_results(self, task_state):
...
+            task_state.hashed_results_by_name = {
+                name: result.to_hashed_result()
+                for name, result in results_by_name.items()
             }
-        task_state.is_complete = True
 
-    def _get_results_for_complete_task_state(self, task_state):
+        return results_by_name
+
+    def _get_hashed_results_for_complete_task_state(self, task_state):
```
jqmp: Is there any special reason `hashed_results_by_name` needs to be set lazily?

bionic/deriver.py
```diff
@@ -542,23 +553,21 @@ def __init__(self, task):
...
-        # These are set by
-        # EntityDeriver._set_up_cache_accessors_for_task_state(), just before
-        # we attempt to find cached values for these tasks.
         self.cache_accessors = None
-        self.has_cached_values = None
 
         # This can be set by
-        # EntityDeriver._attempt_to_load_task_state_results() or
-        # EntityDeriver._compute_task_state().
+        # EntityDeriver._get_results_for_complete_task_state().
```
jqmp: This can also be set by `_compute_task_state`, right?

Maybe we should also make it clear that this should only be accessed via `_get_results_for_complete_task_state()`, and not directly?

bionic/deriver.py
```diff
@@ -542,23 +553,21 @@ def __init__(self, task):
         self.case_key = None
         self.provider = None
 
-        # These are set by
-        # EntityDeriver._set_up_provenance_for_task_state(), just
-        # before the task state becomes eligible for computation.
+        # These are set by EntityDeriver._complete_task_state(), just
+        # before the task state becomes eligible for cache lookup / computation.
```
jqmp: Maybe we could add `They will be present if and only if is_complete is True.`

bionic/cache.py
```diff
@@ -862,10 +893,10 @@ class Provenance(object):
 
     @classmethod
     def from_computation(
-            cls, code_fingerprint, case_key, dep_provenances_by_task_key,
+            cls, code_fingerprint, case_key, dep_provenance_digest_by_task_key,
```
jqmp: Super nit: the convention here is to make the values plural, like `dep_provenance_digests_by_task_key`.

bionic/datatypes.py
```diff
@@ -72,15 +69,50 @@ class Result(object):
...
 
     def __repr__(self):
         return f'Result({self.query!r}, {self.value!r})'
 
 
+class HashedResult(object):
+    '''
+    Represents hashed part of the result.
+    '''
+    def __init__(
```
jqmp: super nit: do we need the extra newline here?

bionic/datatypes.py
```diff
@@ -45,7 +45,7 @@ def key_for_entity_name(self, name):
         return key
 
     def __repr__(self):
-        return f'Task({self.key!r}, {self.dep_keys!r})'
+        return f'Task({self.keys!r}, {self.dep_keys!r})'
```
jqmp: Nice catch!

bionic/datatypes.py
```diff
@@ -72,15 +69,50 @@ class Result(object):
...
+        self.value_hash = value_hash
+
+    def to_hashed_result(self):
+        return HashedResult(self.query, self.value_hash)
 
     def __repr__(self):
         return f'Result({self.query!r}, {self.value!r})'
 
 
+class HashedResult(object):
```
jqmp: So, given the way the code looks now: does this need to be a class? Could we just pass around the value hashes themselves?

bionic/datatypes.py
```diff
@@ -72,15 +69,50 @@ class Result(object):
...
+    def __init__(self, functional_hash, exact_hash):
+        self.functional_hash = functional_hash
+        self.exact_hash = exact_hash
+
+    @classmethod
+    def from_provenance(cls, provenance):
+        return cls(provenance.functional_hash, provenance.exact_hash)
+
+    @classmethod
+    def from_value(cls, value):
```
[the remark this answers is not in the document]
namanjain: ah I can see the confusion. I'll clean up value / hash naming throughout the PR.

bionic/deriver.py
```diff
@@ -302,19 +291,45 @@ def _compute_result_group_for_entity_name(self, entity_name):
...
+
         task = task_state.task
 
-        dep_states = [
-            self._task_states_by_key[dep_key]
-            for dep_key in task.dep_keys
-        ]
+        # First, see if we can load it from the cache.
+        if task_state.provider.attrs.should_persist and \
+                not task.is_simple_lookup and \
```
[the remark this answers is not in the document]
namanjain: You're right. `is_simple_lookup` is only ever `True` for `ValueProvider` which is never persisted. I can get rid of `is_simple_lookup` check and only rely on `should_persist` bit. This also looks cleaner to me.

bionic/deriver.py
```diff
@@ -302,19 +291,45 @@ def _compute_result_group_for_entity_name(self, entity_name):
...
 
-    def _set_up_provenance_for_task_state(self, task_state):
-        if task_state.provenance is not None:
+    def _complete_task_state(self, task_state):
+        assert not task_state.is_blocked()
+        if task_state.is_complete:
             return
 
+        # Set up provenance, queries and cache accessors.
+        self._set_up_provenance_for_task_state(task_state)
```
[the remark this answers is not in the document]
namanjain: I initially inlined everything in this method but it became too big. Even moving all the setup in one single method made it really big and hard to read so I broke it into multiple smaller methods. I agree we probably won't setup one thing without the others though.

bionic/deriver.py
```diff
@@ -302,19 +291,45 @@ def _compute_result_group_for_entity_name(self, entity_name):
             key_space=self._key_spaces_by_entity_name[entity_name],
         )
 
-    def _set_up_provenance_for_task_state(self, task_state):
-        if task_state.provenance is not None:
+    def _complete_task_state(self, task_state):
+        assert not task_state.is_blocked()
+        if task_state.is_complete:
```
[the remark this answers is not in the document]
namanjain: Yeah, I don't see why we can't make that an assert. For some context, this went through multiple refactoring attempts and might contain remnants of the previous attempts.

bionic/deriver.py
namanjain: Now that I think about it, we don't need to set it lazily.

bionic/deriver.py
namanjain: ah this can totally be set by `_compute_task_state`. I'll change the comments for both results and hashed results by name to include more info (like what you said for results by name and that hashed results by name will always be present for completed tasks if it is persisted).

namanjain: Wdyt about renaming the variable to `_` `results_by_name` so that it's clear that no one should access it directly?

bionic/deriver.py
namanjain: Yeah this makes sense. I said a similar thing about `hashed_results_by_name` in my reply to previous comment!

bionic/cache.py
namanjain: I'll change it if it's the convention but I find it a little misleading, gives me the impression that there are multiple values for a task key.

bionic/datatypes.py
namanjain: Nope, probably copy-pasta from above classes

bionic/cache.py
jqmp: That's a fair response. Let's chat more about the convention. For now let's stick to it and consider revising things all at once in another PR?

bionic/deriver.py
```diff
@@ -302,19 +291,45 @@ def _compute_result_group_for_entity_name(self, entity_name):
...
+        # First, see if we can load it from the cache.
+        if task_state.provider.attrs.should_persist and \
+                not task.is_simple_lookup and \
+                all(axr.can_load() for axr in task_state.cache_accessors):
+            # We only load the hashed result while completing task state
+            # and lazily load the entire result when needed later.
+            results = []
+            for accessor in task_state.cache_accessors:
+                result = accessor.load_hashed_result()
+                self._log(
```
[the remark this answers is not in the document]
namanjain: Thanks for the context. I'll remove these logs for now as they don't give the complete picture of why something was loaded vs recomputed. We should revisit this when we work on increasing visibility in the system.

bionic/deriver.py
jqmp: I like that idea!

bionic/datatypes.py
namanjain: Well, that's how I started it but had to make it a class. I see no reason to keep the class now with the refactor since I can get the queries (for map key name) from other entities loading the hash and creating the map.

bionic/deriver.py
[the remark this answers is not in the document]
jqmp: I see your point about it making the function large -- although I think that bothers me less than it does you, especially if you use comments to delineate the different logical parts of the function.

What does bother me here is that it's not obvious that these three functions are intended to be called together, and indeed they still have some early-return code to handle being called multiple times, which I think is vestigial? Maybe we could remove the early-return code and add more asserts (like `assert task_state.queries is not None` near the beginning of `_set_up_cache_accessors_for_task_state`) to make it clear what the expectations are?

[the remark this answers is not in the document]
namanjain: Converting if conditions to asserts and removing some temporary variables did make this method manageable. However, I did keep the versioning problem check in a separate method. It doesn't look crazy anymore.

bionic/datatypes.py
```diff
@@ -72,15 +68,36 @@ class Result(object):
...
+        self.value_hash = value_hash
 
     def __repr__(self):
         return f'Result({self.query!r}, {self.value!r})'
 
 
+class ProvenanceDigest(object):
+    '''
+    A collection of values used by Provenance for different chained hashes.
+    This hashes depend on the entities and can come from either another
```
jqmp: nit:
* s/This hashes/These hashes/ 
* s/or value hash in result/or the value hash of a result/

bionic/deriver.py
```diff
@@ -324,58 +306,77 @@ def _set_up_provenance_for_task_state(self, task_state):
...
-                but the cache is not available yet because core bootstrap
-                entities depend on this one;
-                you should decorate entity {name!r} with `@persist(False)`
-                or `@immediate` to indicate that it can't be cached.'''))
-
-        accessors = [
-            self._persistent_cache.get_accessor(query)
-            for query in task_state.queries
-        ]
+        # Lastly, set up cache accessors
```
jqmp: Super nit: period at end of sentence.

bionic/deriver.py
```diff
@@ -542,22 +512,29 @@ def __init__(self, task):
...
-
-        # These are set by
-        # EntityDeriver._set_up_cache_accessors_for_task_state(), just before
-        # we attempt to find cached values for these tasks.
         self.cache_accessors = None
-        self.has_cached_values = None
 
         # This can be set by
-        # EntityDeriver._attempt_to_load_task_state_results() or
+        # EntityDervier._complete_task_state() or
```
jqmp: nit: s/EntityDervier/EntityDeriver/

bionic/deriver.py
```diff
@@ -542,22 +512,29 @@ def __init__(self, task):
...
+        # This will be present if and only if both is_complete and
+        # provider.attrs.should_persist are True.
+        self.result_value_hashes_by_name = None
+
+        # This can be set by
+        # EntityDeriver._get_results_for_complete_task_state() or
         # EntityDeriver._compute_task_state().
-        self.results_by_name = None
+        #
+        # This should never be accessed direclty, instead use
```
jqmp: Nit: s/direclty/directly

PR #91, opened 2020-03-09
Update docs to support ZSH

jqmp:
Replaced all instances of

    pip install bionic[SOMETHING]

with

    pip install 'bionic[SOMETHING]'

The former doesn't work on Zsh, since that shell assigns a special
meaning to square brackets. Having to add the quotes is kind of
annoying, but now Zsh is the default shell on new Macbooks and it
generates a cryptic error if you don't add the quotes, so it seems
worthwhile to try to help newcomers avoid this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ python:
 - '3.7'
 before_install:
 - sudo apt-get install graphviz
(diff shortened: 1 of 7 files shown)


jqmp:
cc @damienrj

damienrj:
LGTM

Issue #92, opened 2020-03-13
Improve error message when protocols fail


jqmp:
When Bionic fails to save or load a cached value, the resulting stack trace is confusing for users because it doesn't indicate (a) that the problem is with the serialization protocol or (b) which entity was responsible. We should catch these exceptions and wrap them with a more helpful message.

PR #93, opened 2020-03-13
Janek/improve messaging for user code errors

jqmp:
Addresses #92

When we fail to compute or serialize an entity value, this adds a
clearer exception that indicates what the problem was and which
entity was the cause.

Also fixes a bug causing a crash when we try to compute a failing entity twice.

### DIFF
--- a/bionic/cache.py
+++ b/bionic/cache.py
@@ -15,7 +15,8 @@ from uuid import uuid4
 from pathlib import Path
 from urllib.parse import urlparse
 
-from bionic.exception import UnsupportedSerializedValueError
+from bionic.exception import (
+    EntitySerializationError, UnsupportedSerializedValueError)
 from .datatypes import Result
 from .util import (
     get_gcs_client_without_warnings, ensure_parent_dir_exists, oneline)
@@ -317,7 +318,14 @@ class CacheAccessor(object):
         value_path = dir_path / value_filename
 
         ensure_parent_dir_exists(value_path)
-        self.query.protocol.write(value, value_path)
+        try:
+            self.query.protocol.write(value, value_path)
+        except Exception as e:
+            raise EntitySerializationError(oneline(
+                f'''
+                Value of entity {self.query.entity_name!r}
+                could not be serialized to disk
+                ''')) from e
 
         return value_path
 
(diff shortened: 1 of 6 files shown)


bionic/cache.py
```diff
@@ -15,7 +15,8 @@
 from pathlib import Path
 from urllib.parse import urlparse
 
-from bionic.exception import UnsupportedSerializedValueError
+from bionic.exception import (
+    UnsupportedSerializedValueError, EntitySerializationError)
```
namanjain: super nit: I don't know if we follow this code style but let's sort the imports by name

bionic/cache.py
```diff
@@ -317,7 +318,14 @@ def _file_from_value(self, value):
...
         ensure_parent_dir_exists(value_path)
-        self.query.protocol.write(value, value_path)
+        try:
+            self.query.protocol.write(value, value_path)
+        except Exception as e:
+            raise EntitySerializationError(oneline(
+                f'''
+                Value of entity {self.query.entity_name!r}
+                could not be serialized to disk
+                ''')) from e
```
namanjain: what does `from e` do here?

tests/test_flow/test_api.py
```diff
@@ -623,6 +624,43 @@ def xs_sum(xs):
...
+        assert isinstance(e.__cause__, ZeroDivisionError)
+
+
+def test_multiple_compute_attempts(builder):
+    @builder
+    def uncomputable_value():
+        return 1 / 0
+
+    flow = builder.build()
+    with pytest.raises(EntityComputationError):
```
namanjain: nit: can you add a quick comment here why this is tested twice

bionic/cache.py
jqmp: I haven't been doing that but it sounds reasonable to me.

tests/test_flow/test_api.py
jqmp: Good idea!

bionic/cache.py
jqmp: It creates a "chained exception" which includes both the current stack trace and `e`'s original stack trace. It's analogous to exception wrapping in Java:

    } catch (Exception e) {
      throw new OtherException("message", e);
    }

namanjain: this is neat!

PR #95, opened 2020-03-18
add volatile decorator

### LINKED ISSUE #27: Some entities should be recomputed every time (or with some cadence)
Currently, an entity's value is only recomputed if its inputs change or its code changes.  However, sometimes people want entities to be recomputed whenever some amount of time has passed.  For example, we may want to download new data from a database each day.  Right now the best workaround is like this:

```
builder.assign('date', datetime.today().strftime('%Y-%m-%d'))

@builder
@builder
def current_data(date):
    return download_data()
```

This is a bit hacky and doesn't work if the flow is long-lived, because `date` only gets updated when the flow is defined.

Note that the following does not work:

```
@builder
@bn.persist(False)
@builder
def current_data():
    return download_data()

@builder
def summary(current_data):
    return summarize(current_data)
```

This fails because if we call `flow.get('summary')` and it finds a cached value, it will return it because it doesn't know that `current_data` ought to be recomputed.  We need to invalidate not just `current_data`, but all downstream entities as well.

A better approach would be a way to tell Bionic that (a) an entity should be recomputed every time, and (b) downstream entities should only be recomputed if this entity's output changes.  Maybe something like:

```
@builder
@bn.volatile    # Always recompute
@bn.digestible  # Hash the output to determine whether to recompute children
def date():
    return datetime.today().strftime('%Y-%m-%d')

@builder
def current_data(date):
    return download_data()
```

I'm not sure about the names, or about needing two separate decorators (although I do think they would be useful independently).

namanjain:
The decorator recomputes the entity between different flows and triggers cache invalidation on downstream entities.

Closes #27

### DIFF
--- a/bionic/cache.py
+++ b/bionic/cache.py
@@ -899,7 +899,8 @@ class Provenance(object):
     @classmethod
     def from_computation(
             cls, code_fingerprint, case_key, dep_provenance_digests_by_task_key,
-            treat_bytecode_as_functional):
+            treat_bytecode_as_functional, can_functionally_change_per_run,
+            flow_instance_uuid):
         dep_task_key_provenance_digest_pairs = sorted(
             dep_provenance_digests_by_task_key.items())
 
@@ -919,6 +920,12 @@ class Provenance(object):
             functional_code_dict['bytecode_hash'] = bytecode_hash
         else:
             nonfunctional_code_dict['bytecode_hash'] = bytecode_hash
+
+        # The function's output changes with each run; to reflect that,
+        # we add the flow uuid to the hash so that it will be different
+        # each time.
+        if can_functionally_change_per_run:
+            functional_code_dict['flow_instance_uuid'] = flow_instance_uuid
 
         full_code_dict = dict(
             functional=functional_code_dict,
--- a/bionic/decorators.py
+++ b/bionic/decorators.py
@@ -81,7 +81,7 @@ def persist(enabled):
     if not isinstance(enabled, bool):
         raise ValueError(f"Argument must be a boolean; got {enabled!r}")
 
-    return provider_wrapper(AttrUpdateProvider, 'should_persist', enabled)
+    return provider_wrapper(AttrUpdateProvider, '_can_persist', enabled)
 
 
 def memoize(enabled):
@@ -92,7 +92,7 @@ def memoize(enabled):
     ----------
 
     enabled: Boolean
-        Whether this entity's values should be memoized
+        Whether this entity's values should be memoized.
 
     Returns
     -------
@@ -103,7 +103,89 @@ def memoize(enabled):
     if not isinstance(enabled, bool):
         raise ValueError(f"Argument must be a boolean; got {enabled!r}")
 
-    return provider_wrapper(AttrUpdateProvider, 'should_memoize', enabled)
+    return provider_wrapper(AttrUpdateProvider, '_can_memoize', enabled)
+
+
+def changes_per_run(enabled=None):
+    """
+    Indicates whether this function is non-deterministic: i.e., if it’s called multiple
+    times with the same inputs, can it return different outputs?
+
+    When ``enabled`` is true, Bionic will recompute this function's value (and
+    potentially the values of anything depending on it) each time this flow is
+    instantiated, rather than reusing a value cached on disk. For example, if the
+    function queries data from an external database, the results may be different each
+    time even if the query stays the same.
+
+    However, for practical reasons, Bionic won't compute a new value more than once
+    within a single run. That is, once it's been computed for a particular Flow
+    instance, that value will be saved in memory and reused. This is a compromise:
+    logically it makes sense to recompute it every time, but it's much simpler to have
+    a single fixed value for each entity within a given flow instance. For this reason,
+    when this decorator is enabled, memoization must not be disabled for this entity.
+
+    Note that ``@changes_per_run`` is not the same as ``@persist(False)``. For example,
+    the following code will not necessarily query the database each time:
+
+    .. code-block:: python
+
+        @builder
+        @bn.persist(False)
+        @builder
+        def current_data():
+            return download_data()
+
+        @builder
+        def summary(current_data):
+            return summarize(current_data)
+
+    This fails because if we call ``flow.get('summary')`` and Bionic finds a cached
+    value, it will return the cached value because it doesn't know that current_data
+    ought to be recomputed. On the other hand, if we replace ``persist(False)`` with
+    ``changes_per_run`` -- as in the example below -- then ``current_data`` will be
+    recomputed each time (and ``summary`` will be recomputed if ``current_data``
+    changes).
+
+    Parameters
+    ----------
+
+    enabled: Boolean, optional (default ``True``)
+        Whether this function's output changes per run.
+
+    Returns
+    -------
+    Function:
+        A decorator which can be applied to an entity function.
+
+
+    Example usage:
+
+    .. code-block:: python
+
+        @builder
+        @bn.changes_per_run
+        @builder
+        def current_data():
+            return download_data()
+
+        @builder
+        def summary(current_data):
+            return summarize(current_data)
+    """
+
+    DEFAULT_VALUE = True
+    if callable(enabled):
+        func_or_provider = enabled
+        wrapper = provider_wrapper(AttrUpdateProvider, 'changes_per_run', DEFAULT_VALUE)
+        return wrapper(func_or_provider)
+
+    if enabled is None:
+        enabled = DEFAULT_VALUE
+
+    if not isinstance(enabled, bool):
+        raise ValueError(f"Argument must be a boolean; got {enabled!r}")
+
+    return provider_wrapper(AttrUpdateProvider, 'changes_per_run', enabled)
 
 
 def output(name):
--- a/bionic/deriver.py
+++ b/bionic/deriver.py
@@ -26,8 +26,9 @@ class EntityDeriver(object):
 
     # --- Public API.
 
-    def __init__(self, flow_state):
+    def __init__(self, flow_state, flow_instance_uuid):
         self._flow_state = flow_state
+        self._flow_instance_uuid = flow_instance_uuid
 
         # This state is needed to do any resolution at all.  Once it's
         # initialized, we can use it to bootstrap the requirements for "full"
@@ -308,7 +309,7 @@ class EntityDeriver(object):
         dep_provenance_digests_by_task_key = {}
         for dep_key, dep_state in zip(task_state.task.dep_keys, task_state.parents):
             # Use value hash of persistable values.
-            if dep_state.provider.attrs.should_persist:
+            if dep_state.provider.attrs.should_persist():
                 value_hash = dep_state.result_value_hashes_by_name[dep_key.entity_name]
                 dep_provenance_digests_by_task_key[dep_key] =\
                     ProvenanceDigest.from_value_hash(value_hash)
@@ -323,6 +324,8 @@ class EntityDeriver(object):
             case_key=task_state.case_key,
             dep_provenance_digests_by_task_key=dep_provenance_digests_by_task_key,
             treat_bytecode_as_functional=treat_bytecode_as_functional,
+            can_functionally_change_per_run=task_state.provider.attrs.changes_per_run,
+            flow_instance_uuid=self._flow_instance_uuid,
         )
 
         # Then set up queries.
@@ -336,7 +339,7 @@ class EntityDeriver(object):
         ]
 
         # Lastly, set up cache accessors.
-        if task_state.provider.attrs.should_persist:
+        if task_state.provider.attrs.should_persist():
             if not self._is_ready_for_full_resolution:
                 name = task_state.task.keys[0].entity_name
                 raise AssertionError(oneline(f'''
@@ -355,7 +358,7 @@ class EntityDeriver(object):
                 self._check_accessors_for_version_problems(task_state)
 
         # See if we can load it from the cache.
-        if task_state.provider.attrs.should_persist and \
+        if task_state.provider.attrs.should_persist() and \
                 all(axr.can_load() for axr in task_state.cache_accessors):
             # We only load the hashed result while completing task state
             # and lazily load the entire result when needed later.
@@ -422,7 +425,7 @@ class EntityDeriver(object):
 
             results_by_name[result.query.entity_name] = result
 
-        if task_state.provider.attrs.should_memoize:
+        if task_state.provider.attrs.should_memoize():
             task_state._results_by_name = results_by_name
 
         return results_by_name
@@ -463,7 +466,7 @@ class EntityDeriver(object):
                 value=value,
             )
 
-            if provider.attrs.should_persist:
+            if provider.attrs.should_persist():
                 accessor = task_state.cache_accessors[ix]
                 accessor.save_result(result)
 
@@ -476,11 +479,11 @@ class EntityDeriver(object):
         # Otherwise, load it lazily later so that if the serialized/deserialized
         # value is not exactly the same as the original, we still
         # always return the same value.
-        if provider.attrs.should_memoize and not provider.attrs.should_persist:
+        if provider.attrs.should_memoize() and not provider.attrs.should_persist():
             task_state._results_by_name = results_by_name
 
         # But we cache the hashed values eagerly since they are cheap to load.
-        if provider.attrs.should_persist:
+        if provider.attrs.should_persist():
             task_state.result_value_hashes_by_name = result_value_hashes_by_name
 
     def _log(self, message, *args):
@@ -522,7 +525,7 @@ class TaskState(object):
         # EntityDeriver._compute_task_state().
         #
         # This will be present if and only if both is_complete and
-        # provider.attrs.should_persist are True.
+        # provider.attrs.should_persist() are True.
         self.result_value_hashes_by_name = None
 
         # This can be set by
--- a/bionic/provider.py
+++ b/bionic/provider.py
@@ -17,7 +17,7 @@ import pandas as pd
 
 from .datatypes import (
     Task, TaskKey, CaseKey, CaseKeySpace, CodeFingerprint, CodeVersion)
-from .exception import EntityComputationError
+from .exception import (AttributeValidationError, EntityComputationError)
 from .bytecode import canonical_bytecode_bytes_from_func
 from .util import groups_dict, hash_to_hex, oneline
 from .optdep import import_optional_dependency
@@ -30,16 +30,29 @@ class ProviderAttributes(object):
     def __init__(
             self, names,
             protocols=None, code_version=None, orig_flow_name=None,
-            should_persist=None, should_memoize=None, is_default_value=None,
-            docs=None):
+            can_persist=None, can_memoize=None, is_default_value=None,
+            docs=None, changes_per_run=None):
         self.names = names
         self.protocols = protocols
         self.docs = docs
         self.code_version = code_version
         self.orig_flow_name = orig_flow_name
-        self.should_persist = should_persist
-        self.should_memoize = should_memoize
+        self._can_persist = can_persist
+        self._can_memoize = can_memoize
+        self.changes_per_run = changes_per_run
         self.is_default_value = is_default_value
+
+    def should_persist(self):
+        return self._can_persist
+
+    def should_memoize(self):
+        if not self._can_memoize and self.changes_per_run:
+            raise AttributeValidationError(oneline('''
+                Entity with names {names!r} uses @changes_per_run with @memoize(False),
+                which is not allowed. @changes_per_run computes once in a flow instance
+                and memoizes the value. Memoization cannot be disabled for this entity.'''))
+
+        return self._can_memoize
 
 
 class BaseProvider(object):
@@ -268,8 +281,9 @@ class ValueProvider(BaseProvider):
                 names=[name],
                 protocols=[protocol],
                 docs=[doc],
-                should_persist=False,
-                should_memoize=True,
+                can_persist=False,
+                can_memoize=True,
+                changes_per_run=False,
             ),
             is_mutable=True,
         )
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -159,7 +159,7 @@ syntax:
         """A nice thing to say to someone."""
         return f'{greeting} {subject}!'
 
-If the function defines multiple entities, the :meth:`decorators.docs`
+If the function defines multiple entities, the :meth:`@docs <bionic.docs>`
 decorator can be used to specify documentation for each one:
 
 .. code-block:: python
@@ -200,10 +200,10 @@ Caching and Protocols
 ---------------------
 
 Whenever Bionic computes an entity's value, it automatically caches that value
-in memory (in case you access it again in from the same ``Flow`` object) and in
-persistent storage (in case you want to access it later, perhaps after
-restarting your script or notebook).  Currently the only supported location for
-persistent storage is your computer's hard disk.
+in memory (in case you access it again in from the same ``Flow`` object) and to
+disk (in case you want to access it later, perhaps after restarting your script
+or notebook).  Bionic can also be configured to cache to
+:ref:`Google Cloud Storage <google_cloud_storage_anchor>`.
 
 Bionic's caching can be seen in action in the `ML tutorial`_.
 
@@ -355,6 +355,9 @@ we can disable persistent caching altogether:
     def message(subject):
         return f'Hello {subject}.'
 
+If your goal is just to force an entity to be recomputed more frequently, you may
+want :ref:`@changes_per_run <changes_per_run>` instead.
+
 
 Disabling In-Memory Caching
 ............................
@@ -370,6 +373,91 @@ cases, we can disable in-memory caching:
     @bionic.memoize(False)
     def message(subject):
         return f'Hello {subject}.'
+
+
+.. _changes_per_run :
+
+Non-Deterministic Computation
+.............................
+
+.. versionadded:: 0.6.5
+
+The basic assumption behind Bionic's caching behavior is that entity functions
+are *deterministic*: if you call them multiple times with the same input, they
+always return the same output.  However, some functions are
+*non-deterministic*: their output can change even when their input doesn't.
+For example, a function that retrieves data from an external database may
+return different results whenever the database's contents change.  In
+cases like this, it's not appropriate to reuse the function's previous cached
+values; we want Bionic to recompute the value each time.
+
+You can tell Bionic that a function is non-deterministic by applying the 
+:meth:`@changes_per_run <bionic.changes_per_run>` decorator:
+
+.. code-block:: python
+
+    @builder
+    @bn.changes_per_run
+    def current_data():
+        return download_data()
+
+This causes Bionic to recompute the value instead of loading a cached value
+from disk. (However, this recomputation will only happen once for any given
+``Flow`` instance; after that, the value will be cached in memory and reused
+[#f3]_.)
+
+.. [#f3] I.e., the value is computed once per "run".  This is a compromise:
+  although it makes logical sense to recompute the value every single time,
+  it's much simpler for each entity to have a consistent value within a single
+  flow instance.
+
+``@changes_per_run`` vs ``@persist``
+::::::::::::::::::::::::::::::::::::
+
+Note that ``@changes_per_run`` has a different effect from ``@persist(False)``.
+If an entity is decorated with ``@persist(False)``, Bionic will never cache its
+value to disk, but it will still assume that its output is deterministic. The
+difference can be see when we add a downstream entity:
+
+.. code-block:: python
+
+    @builder
+    @bn.persist(False)
+    def current_data():
+        return download_data()
+
+    @builder
+    def summary(current_data):
+        return summarize(current_data)
+
+In this case, ``builder.build().get('current_data')`` will always recompute
+``current_data``, since its value is never persisted. However,
+``builder.build().get('summary')`` will use a cached value if one is available;
+Bionic won't bother to recompute ``current_data`` because it assumes its value
+will be the same anyway. In more complex flows, this incorrect assumption may
+lead to inconsistent results.
+
+By contrast, if we use the appropriate decorator, ``@bn.changes_per_run``:
+
+.. code-block:: python
+
+    @builder
+    @bn.changes_per_run
+    def current_data():
+        return download_data()
+
+    @builder
+    def summary(current_data):
+        return summarize(current_data)
+
+
+Here ``builder.build().get('summary')`` will always recompute ``current_data``
+first. Then, if ``current_data``'s value has changed, it will recompute
+``summary`` as well; otherwise it will use a cached value.
+
+As a rule: use ``@persist(False)`` for entities whose values are *impossible to
+serialize* or *not worth serializing*. Use ``@changes_per_run`` for
+entities whose values are *non-deterministic*.
 
 
 Location of the Cache Directory
@@ -388,6 +476,9 @@ two internal entities:
 
     # Cache this flow's data in /my_cache_dir/
     builder.set('core__persistent_cache__flow_dir', 'my_cache_dir')
+
+
+.. _google_cloud_storage_anchor :
 
 Caching in Google Cloud Storage
 ...............................
--- a/tests/test_flow/test_persistence.py
+++ b/tests/test_flow/test_persistence.py
@@ -3,7 +3,7 @@ import pytest
 import math
 
 from ..helpers import count_calls, ResettingCounter, RoundingProtocol
-from bionic.exception import CodeVersioningError
+from bionic.exception import (AttributeValidationError, CodeVersioningError)
 
 import bionic as bn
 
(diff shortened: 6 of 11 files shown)


bionic/decorators.py
```diff
@@ -103,7 +103,33 @@ def memoize(enabled):
     if not isinstance(enabled, bool):
         raise ValueError(f"Argument must be a boolean; got {enabled!r}")
 
-    return provider_wrapper(AttrUpdateProvider, 'should_memoize', enabled)
+    return provider_wrapper(AttrUpdateProvider, '_should_memoize', enabled)
+
+
+def volatile(enabled):
```
jqmp: This should probably default to `True`. In fact, it's not obvious to me that we need a parameter at all, although I can see why you wanted to maintain consistency.

bionic/decorators.py
```diff
@@ -103,7 +103,33 @@ def memoize(enabled):
...
+    """
+    Indicates that the entity should always be recomputed between different
```
jqmp: There are two things we can talk about here:

1. What this decorator says about the decorated function: namely, that it is non-deterministic.
2. What behavior is caused by this decorator: namely, that this entity will be recomputed even if its inputs don't change.

I would start with 1, because it's more fundamental. That's also why I suggested the name `@volatile` (a fact about the function) instead of, say, `@always_recompute` (a desired behavior).

We should also discuss 2, but I'm not sure that "in the same flow" and "between different flows" are clear enough. It's not obvious whether "different flows" means "two different flow objects" or "flows from totally different Python files". I also think the key point for 2 is that the entity will be recomputed if its inputs don't change; the same/different flow thing is secondary (but still important and worth mentioning).

WDYT?

bionic/provider.py
```diff
@@ -31,16 +31,23 @@ def __init__(
...
         self.code_version = code_version
         self.orig_flow_name = orig_flow_name
-        self.should_persist = should_persist
-        self.should_memoize = should_memoize
+        self._should_persist = should_persist
+        self._should_memoize = should_memoize
+        self.is_volatile = is_volatile
         self.is_default_value = is_default_value
 
+    def should_persist(self):
```
jqmp: It seems confusing to have a method with the same name as an attribute, but a subtly different meaning.

Also, if we disable persistence for volatile entities, that means we won't do the save-to-disk-and-reload step after computing them, right? I'm not sure if that's what we want.

bionic/provider.py
```diff
@@ -31,16 +31,23 @@ def __init__(
...
+        return self._should_persist and not self.is_volatile
+
+    def should_memoize(self):
+        return self._should_memoize or self.is_volatile
```
jqmp: What should happen if someone sets `@volatile(True)` and `@memoize(False)`? Maybe it's safer to throw an exception than to silently ignore the `memoize` part?

tests/test_flow/test_persistence.py
```diff
@@ -821,3 +821,58 @@ def y():
...
+    assert x_plus_one.times_called() == 1
+    # x_plus_two is a child of a volatile parent and should also be
+    # recomputed between flows.
+    assert x_plus_two.times_called() == 1
+    # x_plus_three is a child of a parent that has a volatile parent
+    # and is not persisted. Hence the provenance also contains the
+    # noise from volatile ancestor, cannot use the persisted value and
+    # should also be recomputed.
+    assert x_plus_three.times_called() == 1
+    # x_plus_four uses the value hash of it's parent which didn't change.
```
jqmp: nit: s/it's/its/

tests/test_flow/test_persistence.py
```diff
@@ -821,3 +821,58 @@ def y():
...
+    def x_plus_four(x_plus_three):
+        return x_plus_three + 1
+
+    flow = builder.build()
+    assert flow.get('x_plus_four') == 9
+    assert x_plus_one.times_called() == 1
+    assert x_plus_two.times_called() == 1
+    assert x_plus_three.times_called() == 1
+    assert x_plus_four.times_called() == 1
+    # In the same flow, volatile is not recomputed.
```
jqmp: nit: s/volatile/a volatile entity/

tests/test_flow/test_persistence.py
```diff
@@ -821,3 +821,58 @@ def y():
...
+    assert flow.get('x_plus_four') == 9
```
jqmp: Maybe we should test accessing the volatile entity directly as well? (`flow.get('x_plus_one')`)

tests/test_flow/test_persistence.py
```diff
@@ -821,3 +821,58 @@ def y():
 
         flow = builder.build()
         assert flow.get('y') == 1
+
+
+def test_volatile_ancestor(builder):
```
jqmp: Looks like we don't have a test for the interaction between `@volatile` and `@memoize`?

bionic/cache.py
```diff
@@ -920,6 +920,9 @@ def from_computation(
         else:
             nonfunctional_code_dict['bytecode_hash'] = bytecode_hash
 
+        if is_functionally_volatile:
```
jqmp: Maybe this warrants a quick explanatory comment?

tests/test_flow/test_persistence.py
jqmp: Any idea how hard it would be to incorporate volatile entities into `test_persistence_fuzz.py`?

bionic/decorators.py
namanjain: I initially didn't add a parameter to this decorator, so I had to default the value of `is_volatile` to `False` in `AttributeProvider`.  But when the decorator update the attribute, it failed the check where an attribute can be set only once in `AttrUpdateProvider`. It was easier to keep the enable variable like other attribute decorators.

bionic/decorators.py
namanjain: I agree about non-deterministic and recompute regardless of input parts. 

I still think re-computation between flows is important enough to mention. I can see the confusion between flows and flow objects, I can reword it to say that the volatile entity is not recomputed within the same flow object. 

How does this sound?

bionic/provider.py
namanjain: I thought I changed the variable name to `_can_persist` and `_can_memoize`, I'll make that change. 

We don't have to disable persistence for volatile entities, it's just that we won't ever use the cached value. I thought the entire purpose of save-to-disk-and-reload was to have consistency between different runs (the first time it is computed vs the next time it is reloaded from the disk). Or are you suggesting there might be some implications down the dependencies, something like if they cache a part of the value and also use it directly from the entity?

bionic/provider.py
namanjain: Fair point. Users will have to add the `memoize` decorator to not memoize the value alongside `volatile`. Maybe I should call this out in the `volatile` documentation too. What do you think?

tests/test_flow/test_persistence.py
namanjain: Sure, I'll add it

tests/test_flow/test_persistence.py
namanjain: Based on another comment thread, I guess I can add it and test if it throws error.

bionic/decorators.py
namanjain: I'm probably missing some more context here. I put a default value on the attribute since the attribute should be false by default. Without having a default attribute value, it won't be possible to set the variable to false unless we have another method to set it to false aside from this decorator.

jqmp: Ah, yes, that's also a decent reason to keep the parameter. But we should probably allow `@volatile` to be used like this:

```
@builder
@bn.volatile # no arguments!
def entity(...):
    ....
```

`@pyplot` is an existing example of a decorator that can be used with or without arguments.

bionic/decorators.py
jqmp: Oh, I totally agree that we should explain what happens with different flow objects. I would just structure the explanation so that the most foundational parts come first.

bionic/provider.py
jqmp: My concern here was that the save-reload process will subtly change the value, which means that adding `@volatile` to an entity will change that entity's value. It's not clear to me if that's desirable/acceptable.

But also, even if an entity is volatile, we still want to be able to hash its output value, right? So we still need to write it to disk anyway. But maybe we should be writing it to a temporary location.

(Sorry if I'm missing something, I'm running late and writing this in a hurry! I'll get to the other outstanding comments tomorrow.)

bionic/provider.py
jqmp: Sorry, I didn't understand this comment. Users need to *not* add the memoize decorator, right?

bionic/decorators.py
namanjain: Oh I didn't realize there was an easy way to do this!

bionic/provider.py
namanjain: We don't need to hash the output value yet since we won't use it anywhere. If or when we introduce something `@digestible`, it would change though and maybe we should reconsider the whole hashing on persist thing.

bionic/provider.py
namanjain: Oh I meant users will have to add both the decorators for us to be in invalid state (where memoize is false and volatile is true).

bionic/provider.py
namanjain: Talked offline, here is the discussion result for posterity. 

It makes sense for volatile to persist with `@persist(True)` decorator since even if we don't use the persisted value, the children can still use the hash and not recompute if not needed. Although it would be nice to not persist the value, we can make this change at a later date, perhaps when we introduce `@digestable` where we write the output to a tmp directory instead of the caching directory. 

Also, going to change the names of the variables to `_can_memoize` and `_can_persist`.

tests/test_flow/test_persistence.py
namanjain: it wasn't as easy as I hoped but done

namanjain:
Added a section in caching docs and an entry in decorator reference. I think the only thing left now is to decide on a name and update the name of the decorator.

tests/test_flow/test_persistence_fuzz.py
```diff
@@ -96,19 +96,43 @@ def entity_names(
...
+    def entity_has_nondeterministic_ancestor(self, entity_name):
+        entity = self._entities_by_name[entity_name]
+        return len(entity.all_nondeterministic_upstream_names) > 0
+
+    def entity_is_nondeterministic(self, entity_name):
+        return self._entities_by_name[entity_name].changes_per_run
+
     def called_entity_names(self):
         names = list(self._last_called_names)
         while self._last_called_names:
```
jqmp: I have no recollection of why I didn't just use `list.clear()` here; maybe it was to support Python 2.7?

tests/test_flow/test_persistence_fuzz.py
```diff
@@ -181,25 +208,25 @@ def add_entities(self, n_entities):
...
+            changes_per_run = self._random_bool()
             for name in all_names:
                 if self._random_bool():
                     dep_names.append(name)
-            new_name = self.model.add_entity(dep_names)
+            new_name = self.model.add_entity(dep_names, changes_per_run)
 
             self.model.build_flow().get(new_name)
-            assert self.model.called_entity_names() == [new_name]
+            self.model.reset_called_entity_names()
```
jqmp: Is there any sensible assert we could use to replace what we used to have here?

tests/test_flow/test_persistence_fuzz.py
```diff
@@ -221,14 +248,23 @@ def run(self, n_iterations):
                     returned_value = self.model.build_flow().get(affected_name)
                     expected_value = self.model.expected_entity_value(
                         affected_name)
-                    if make_func_change:
+
+                    # value will be different if we made a functional change and
```
jqmp: Nit: let's use complete sentences.

tests/test_flow/test_persistence_fuzz.py
```diff
@@ -221,14 +248,23 @@ def run(self, n_iterations):
...
+                    # change, the value hash will change and everything will recompute.
+                    if make_func_change and not is_updated_entity_nondeterministic:
                         assert returned_value != expected_value
                     else:
                         assert returned_value == expected_value
 
-                    # We should have used the cached value, so no new
-                    # computation should have happened.
-                    assert len(self.model.called_entity_names()) == 0
+                    # If updated entity does not change per run, we call only nondeterministic
```
jqmp: I'm not sure what the `It's safe ...` part means: could you elaborate on why it wouldn't be safe and what the "later check" refers to?

tests/test_flow/test_persistence_fuzz.py
```diff
@@ -221,14 +248,23 @@ def run(self, n_iterations):
...
+                    # ancestors. It's safe to check this now since nodeterministic ancestors
+                    # will be recomputed instead of using the disk value in the later check.
+                    if not is_updated_entity_nondeterministic:
```
jqmp: Is there an assertion we can make when `is_updated_entity_nondeterministic` is true?

tests/test_flow/test_persistence_fuzz.py
```diff
@@ -239,6 +275,11 @@ def run(self, n_iterations):
                     )
 
                 elif self._versioning_mode == 'assist':
+                    # No point in assisting when the updated entity changes per run.
```
jqmp: Are there no sensible assertions we can make here? It's possible to use `@changes_per_run` and assisted versioning in the same flow, right?

tests/test_flow/test_persistence_fuzz.py
```diff
@@ -270,17 +311,30 @@ def run(self, n_iterations):
                     self.model.expected_entity_value(affected_name)
                 )
 
-            if make_func_change or self._versioning_mode == 'auto':
-                expected_called_names = self.model.entity_names(
+            if make_func_change:
+                entity_names_in_between = self.model.entity_names(
```
jqmp: nit: why not just do `entity_names_in_between  = set(self.model.entity_names(...))`?

tests/test_flow/test_persistence_fuzz.py
```diff
@@ -270,17 +311,30 @@ def run(self, n_iterations):
...
+
+                # TODO also check that the results with multiple calls are all nondeterminstic
+                assert set(actual_called_names) == expected_called_names
             else:
-                assert len(self.model.called_entity_names()) == 0
+                actual_called_names = self.model.called_entity_names()
+                expected_called_names =\
+                    self.model.nondeterministic_upstream_entity_names(affected_names)
+                if self._versioning_mode == 'auto':
+                    expected_called_names.add(updated_name)
```
jqmp: Shouldn't we be adding every entity in between the updated and affected nodes?

bionic/cache.py
```diff
@@ -899,7 +899,7 @@ class Provenance(object):
     @classmethod
     def from_computation(
             cls, code_fingerprint, case_key, dep_provenance_digests_by_task_key,
-            treat_bytecode_as_functional):
+            treat_bytecode_as_functional, is_functionally_volatile):
```
jqmp: Do we still want to use the term "volatile"?

docs/concepts.rst
```diff
@@ -202,8 +202,8 @@ Caching and Protocols
 Whenever Bionic computes an entity's value, it automatically caches that value
 in memory (in case you access it again in from the same ``Flow`` object) and in
 persistent storage (in case you want to access it later, perhaps after
-restarting your script or notebook).  Currently the only supported location for
-persistent storage is your computer's hard disk.
+restarting your script or notebook).  Currently the only supported locations for
```
jqmp: Nice catch!

docs/concepts.rst
```diff
@@ -202,8 +202,8 @@ Caching and Protocols
...
+persistent storage are your computer's hard disk and gcs.
```
jqmp: nit: s/gcs/Google Cloud Storage/

Also, maybe we no longer need to phrase this in terms of "only".

bionic/cache.py
```diff
@@ -920,6 +920,13 @@ def from_computation(
         else:
             nonfunctional_code_dict['bytecode_hash'] = bytecode_hash
 
+        # Make the functional hash volatile by adding some noise.
+        # This exists for making the provenance unstable when the entity
+        # changes per run and does not persist so that the children won't
+        # use their cached value.
+        if is_functionally_volatile:
```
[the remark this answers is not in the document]
namanjain: Yup, we can definitely setup the provenance multiple times as long as there is an error after setting it up in `EntityDeriver._complete_task_state()`. However, I don't think anyone would use the provenance if the method threw an exception so this should not be a problem. 

If we really want this to be stable in case a change in future breaks this, we can assign a uuid to the flow object and use it instead of random uuid. Wdyt?

bionic/provider.py
```diff
@@ -30,17 +30,28 @@ class ProviderAttributes(object):
...
+        self.changes_per_run = changes_per_run
         self.is_default_value = is_default_value
 
+    def should_persist(self):
+        return self._can_persist
+
+    def should_memoize(self):
+        if not self._can_memoize and self.changes_per_run:
+            raise AttributeValidationError(
+                "Entity with names {names!r} can't change per run without memoizing")
```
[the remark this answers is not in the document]
namanjain: The names thing is still a little confusing sometimes (like if they refer to the entity name or the output of the entity or an entity for each output or a name for each case), I'll update the comment.

I can be more explicit about what exactly did the user did wrong here.

tests/test_flow/test_persistence_fuzz.py
```diff
@@ -24,10 +24,10 @@ def __init__(self, builder):
         self._entities_by_name = {}
         self._last_called_names = []
 
-    def add_entity(self, dep_names):
+    def add_entity(self, dep_names, changes_per_run=False):
```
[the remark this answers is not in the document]
namanjain: Yeah I can call in nondeterministic everywhere. It was hard to use charges per run, partly because it was awkward sometimes and partly because the names were getting huge.

tests/test_flow/test_persistence_fuzz.py
```diff
@@ -181,25 +208,25 @@ def add_entities(self, n_entities):
             all_names = self.model.entity_names()
 
             dep_names = []
+            changes_per_run = self._random_bool()
```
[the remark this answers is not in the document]
namanjain: I like the idea of reducing the frequency of nondeterministic nodes instead of adding another test. FWIW, `test_small_fixed_flow_short_fuzz` does not use `changes_per_node`.

tests/test_flow/test_persistence_fuzz.py
namanjain: lol not sure, I didn't notice it really. I'll make it use `clear()`.

tests/test_flow/test_persistence_fuzz.py
namanjain: The assertion here would really be what is asserted between lines 331-337. I can move that into a separate method and call that assert here.

namanjain: The assertion here would really be what is asserted between lines 331-337. I can move that into a separate method and call that assert here.

namanjain: Sure

tests/test_flow/test_persistence_fuzz.py
namanjain: Fair. This is one of the trickiest part of the test now IMO.

tests/test_flow/test_persistence_fuzz.py
namanjain: I changed this to `is_updated_entity_deterministic or make_nonfunc_change` and explained why we are excluding the checks when it is false. 

Those cases are tested at the end of the method (outside the `not update_version` if conditional and inside `make_func_change` if conditional).

tests/test_flow/test_persistence_fuzz.py
namanjain: Yes, you can use `assist` "mode" in the same flow that uses `@changes_per_run`, but this if conditional checks for the case when we did not update the version in assisted mode for a `changes_per_run` entity. 

It doesn't really make sense to use `@version` on the same entity as `@changes_per_run`.  I didn't want to go this far but maybe we should not even allow that. Wdyt?

tests/test_flow/test_persistence_fuzz.py
namanjain: I don't know why I didn't do that!

tests/test_flow/test_persistence_fuzz.py
namanjain: Not when there is a non functional change since we now use value hash instead of provenance in caching. 

This is honestly on me. When I made changes to this test with the value hash change, I didn't test for this case. I realized that when I really had to go through the test for this change.

namanjain: Also, given that even you missed it, this probably warrants a comment. I'll add one.

bionic/cache.py
namanjain: hmm, I wanted to let the method know that we should update the functional part and adding changes per run to it made the variable name too long.

namanjain: Now that I think of it, `can_functionally_change_per_run` is not that long really.

namanjain:
Resolved conversations that I addressed and left the unresolved ones open.

docs/concepts.rst
```diff
@@ -372,6 +372,63 @@ cases, we can disable in-memory caching:
         return f'Hello {subject}.'
 
 
+Non-Deterministic Entities
+............................
+
+.. versionadded:: 0.6.4
```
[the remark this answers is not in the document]
namanjain: I'm not sure how to bump this version. I changed the file but I suspect that's not the only thing I need to do here. Can we pair on this?

bionic/cache.py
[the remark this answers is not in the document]
jqmp: I like the idea of attaching a UUID to the flow object. If it's not a lot of work I would lean towards doing that -- otherwise I agree that it's currently hard to see how this would break.

bionic/provider.py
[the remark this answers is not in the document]
jqmp: Yeah, the current structure of the code makes this a bit confusing. We should probably have some kind of `EntityDefinition` object that contains things belonging specifically to an entity. FWIW, names are 1-to-1 with entities.

docs/concepts.rst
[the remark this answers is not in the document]
jqmp: Yeah, there's a separate process for doing releases. We do it separately from feature PRs. I'll walk you through it when we do the next release.

It's a little awkward to refer to a not-yet-existing release in the docs, especially since we don't know for certain if the next release will be 0.6.5 or 0.7.0. However, this text won't be visible in the stable docs until we ship our next release, so it's not a huge deal.

Longer term, it would probably be a good idea to maintain a proper changelog and some kind of release plan.

tests/test_flow/test_persistence_fuzz.py
jqmp: Oh right! That makes sense.

So, does this mean the test code on master is currently wrong? Is it surprising that we didn't hit this case?

docs/concepts.rst
```diff
@@ -202,8 +202,8 @@ Caching and Protocols
 Whenever Bionic computes an entity's value, it automatically caches that value
 in memory (in case you access it again in from the same ``Flow`` object) and in
 persistent storage (in case you want to access it later, perhaps after
-restarting your script or notebook).  Currently the only supported location for
-persistent storage is your computer's hard disk.
+restarting your script or notebook).  Currently the supported locations for
```
jqmp: Reading this paragraph again, maybe we should just replace "in persistent storage" with "to disk", and then mention Google Cloud Storage in the next sentence. BTW, when we mention GCS we can include a link to the section in the doc that addresses it.

docs/concepts.rst
```diff
@@ -372,6 +372,63 @@ cases, we can disable in-memory caching:
         return f'Hello {subject}.'
 
 
+Non-Deterministic Entities
```
jqmp: Maybe "Non-Deterministic Computation"?

bionic/decorators.py
```diff
@@ -103,7 +103,47 @@ def memoize(enabled):
...
+    The non-deterministic nature is only applicable between runs. When the entity
+    is called more than once in the same flow object, it does not recompute and
+    returns the previously computed result. To achieve this, the entity will be
+    memoized. For this reason, an entity cannot change per run when @memoize(False).
+
+    Parameters
+    ----------
+
+    enabled: Boolean
+        Whether this entity's values changes per run.
```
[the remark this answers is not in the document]
namanjain: I made this function's output (I'm still not sure why we should not say outputs since a function can return multiple values)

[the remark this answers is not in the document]
jqmp: I think you could make a case for either "value"/"output" or "value(s)"/"output(s)" here. However, once we start using descriptors, I think the correct interpretation will be that a function has a single value/output, which might be interpreted by Bionic as corresponding to multiple entities. For that reason, I would lean toward "value"/"output".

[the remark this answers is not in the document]
namanjain: makes sense to me

bionic/provider.py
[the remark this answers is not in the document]
namanjain: I just realized why this is still here while writing why the combination isn't supported. I think we can support it now. This really started from when I wasn't caching value on disk for functions marked with `@changes_per_run`. Now that we are caching it on disk when `_can_persist` is true, I think we can have `_can_memoize` false. We can change this once we decide to stop persisting on disk for `@changes_per_run` (or store it in a tmp location). So I'm going to remove this and revert everything back to `should_persist` and `should_memoize` unless you think it's worth keeping it for later.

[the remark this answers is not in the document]
namanjain: Talked offline, we'll require `memoize` so that we don't have to change the behavior later.

namanjain:
Addressed all comments.

tests/test_flow/test_persistence_fuzz.py
```diff
@@ -146,9 +171,10 @@ class ModelEntity(object):
     Represents one entity in the SimpleFlowModel.
     """
 
-    def __init__(self, name, dep_names):
+    def __init__(self, name, dep_names, nondeterministic):
```
jqmp: nit: wouldn't it be more standard to call this `is_nondeterministic`?

tests/test_flow/test_persistence_fuzz.py
```diff
@@ -221,14 +255,37 @@ def run(self, n_iterations):
                     returned_value = self.model.build_flow().get(affected_name)
                     expected_value = self.model.expected_entity_value(
                         affected_name)
-                    if make_func_change:
+
+                    # Returned value will be different from expected value if we made
```
jqmp: This first sentence seems to be just restating the first part of the if statement. It seems like the key idea here is is that under certain circumstances, Bionic will miss the fact that a change occurred.

Also, the comment immediately above is no longer always correct, right?

tests/test_flow/test_persistence_fuzz.py
```diff
@@ -221,14 +255,37 @@ def run(self, n_iterations):
...
+                    # everything will recompute.
+                    if make_func_change and is_updated_entity_deterministic:
                         assert returned_value != expected_value
                     else:
                         assert returned_value == expected_value
 
-                    # We should have used the cached value, so no new
-                    # computation should have happened.
-                    assert len(self.model.called_entity_names()) == 0
+                    # If updated entity is deterministic, the entity won't recompute.
```
jqmp: This feels like a lot of text for a single assert statement. The text is also confusing: it doesn't provide any introductory context to orient the reader, and it has references to system state and to specific lines of code in other sections.

The amount of effort required to explain this code makes me wonder if we should try to modify the code to be less subtle. However, if that's not possible and we really do need this much text, we should start with a couple of sentences on what this part of the code is trying to accomplish and why it's so tricky.

tests/test_flow/test_persistence_fuzz.py
```diff
@@ -270,21 +332,41 @@ def run(self, n_iterations):
...
                     downstream_of=updated_name,
                     upstream_of=affected_names,
-                )
-                assert (
-                    list(sorted(self.model.called_entity_names())) ==
-                    list(sorted(expected_called_names))
-                )
+                ))
+                nondeterministic_ancestors = set()
+                for entity_name in entity_names_in_between:
```
jqmp: Could we just use `nondeterministic_ancestors = self.model.nondeterministic_upstream_entity_names(affected_names)` here? If some entity X is upstream of any in-between entity, doesn't that imply that X is also upstream of some affected entity?

If so, I wonder if we could factor out the common parts of these two cases (`make_func_change` and `not make_func_change`).

bionic/deriver.py
```diff
@@ -26,8 +26,9 @@ class EntityDeriver(object):
 
     # --- Public API.
 
-    def __init__(self, flow_state):
+    def __init__(self, flow_state, flow_uuid):
         self._flow_state = flow_state
+        self._flow_uuid = flow_uuid
```
jqmp: nit: if a `flow_uuid` is not part of a `flow_state`, maybe it should be called a `flow_instance_uuid`?

bionic/decorators.py
```diff
@@ -103,7 +103,89 @@ def memoize(enabled):
     if not isinstance(enabled, bool):
         raise ValueError(f"Argument must be a boolean; got {enabled!r}")
 
-    return provider_wrapper(AttrUpdateProvider, 'should_memoize', enabled)
+    return provider_wrapper(AttrUpdateProvider, '_can_memoize', enabled)
+
+
+def changes_per_run(enabled=None):
+    """
+    Indicates whether this function is non-deterministic: when called multiple
```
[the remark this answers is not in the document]
namanjain: so "i.e., if, when called ...". That doesn't sound right.

bionic/decorators.py
```diff
@@ -103,7 +103,89 @@ def memoize(enabled):
...
+    times with same inputs, it can return different outputs.
+
+    When this is used to decorate an entity function, Bionic will recompute this
+    function's value (and potentially the values of anything depending on it) each
+    time this flow is instantiated, rather than reusing a value cached on disk.
+    For example, we may want to download data from a database even when the query
+    doesn't change.
+
+    However, for practical reasons, Bionic won't recompute the value in a single run.
```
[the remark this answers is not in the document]
namanjain: ah the point about taking more than one run to recompute makes sense. I'll change it to what you suggested.

bionic/decorators.py
```diff
@@ -103,7 +103,89 @@ def memoize(enabled):
...
+
+    This fails because if we call flow.get('summary') and it finds a cached value, it
+    will return the cached value because it doesn't know that current_data ought to
+    be recomputed. We need to invalidate not just current_data, but all downstream
+    entities as well.
+
+    Parameters
+    ----------
+
+    enabled: Boolean
```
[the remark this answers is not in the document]
namanjain: Ah yes, I'll make it optional and specify default value

bionic/decorators.py
```diff
@@ -103,7 +103,89 @@ def memoize(enabled):
...
+        Whether this function's output changes per run.
+
+    Returns
+    -------
+    Function:
+        A decorator which can be applied to an entity function.
+
+    Example usage:
```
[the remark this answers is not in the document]
namanjain: I kept it after `Returns` section to keep it consistent with what `@gather` does. I can also move the example usage in `@gather`. Wdyt?

docs/concepts.rst
```diff
@@ -372,6 +372,38 @@ cases, we can disable in-memory caching:
...
 
+Non-Deterministic Computation
+............................
+
+.. versionadded:: 0.6.5
+
+There are some cases where an entity is non-determinsitic and should be computed
+in every run. For example, we may want to download new data from a database.
+
+We can decorate the entity with :meth:`decorators.changes_per_run` which tell Bionic that
```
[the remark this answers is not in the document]
namanjain: should it be `bionic.changes_per_run` or `decorators.changes_per_run`?

tests/test_flow/test_persistence_fuzz.py
namanjain: It's not. I guess the comment is pretty unclear. So the returned value from the flow won't be recomputed because the provenance will not change in the if conditional. Hence, the value from the flow won't be the same as expected value which is the actual recomputed value.

But if don't make a functional change, then the values will still be the same. In similar fashion, if we make a functional change and the entity is nondeterministic, it will be recomputed and hence return the updated value. Let's take this discussion to slack.

tests/test_flow/test_persistence_fuzz.py
```diff
@@ -270,21 +323,40 @@ def run(self, n_iterations):
...
+            assert set(actual_called_names) == expected_called_names
 
     def _random_bool(self):
         return self._random.choice([True, False])
 
+    def _random_bool_with_weighted_probability(self, prob_true):
+        assert 0.0 <= prob_true <= 1.0
+        return choice([True, False], p=[prob_true, 1 - prob_true])
+
+    def _expected_called_names(
```
jqmp: As we discussed on Slack, this function signature feels awkward to me, but I don't have any better ideas either.

tests/test_flow/test_persistence_fuzz.py
```diff
@@ -239,6 +287,11 @@ def run(self, n_iterations):
                     )
 
                 elif self._versioning_mode == 'assist':
+                    # No point in assisting when the updated entity changes per run.
+                    # Ideally, we should not have versions on such entities but that
+                    # would require more changes to this test than I already made.
+                    if is_updated_entity_nondeterministic:
+                        break
```
jqmp: As we discussed, `break` is probably not what you want here.

bionic/decorators.py
```diff
@@ -103,7 +103,89 @@ def memoize(enabled):
...
         raise ValueError(f"Argument must be a boolean; got {enabled!r}")
 
-    return provider_wrapper(AttrUpdateProvider, 'should_memoize', enabled)
+    return provider_wrapper(AttrUpdateProvider, '_can_memoize', enabled)
+
+
+def changes_per_run(enabled=None):
+    """
+    Indicates whether this function is non-deterministic: i.e., if it’s called multiple
+    times with the same inputs, can it return different outputs.
```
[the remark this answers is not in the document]
namanjain: oh yeah, question mark works too

PR #98, opened 2020-03-30
Standardize code with black and add support for pre-commit

damienrj:
Standardize code and add support for pre-commit for flake8


.pre-commit-config.yaml
```diff
@@ -0,0 +1,5 @@
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
```
namanjain: This is pretty awesome! But [pre-commit-hooks](https://github.com/pre-commit/pre-commit-hooks) says that this hook is deprecated and points to [upstream flake8](https://gitlab.com/pycqa/flake8). Can you update `repo` to the new one? Optionally, leave [using-hooks](https://flake8.pycqa.org/en/latest/user/using-hooks.html) link for anyone who wants to modify this later.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,5 @@
+repos:
```
namanjain: Can you add a hook for black as well?

bionic/extras.py
```diff
@@ -10,38 +10,40 @@
...
-        'jupyter',
-        'bumpversion',
+        "pytest",
+        "flake8",
+        "sphinx",
+        "sphinx_rtd_theme",
+        "sphinx-autobuild",
+        "nbsphinx",
+        "jupyter",
+        "bumpversion",
```
namanjain: I think we'll want `black` as part of `dev`

bionic/flow.py
```diff
@@ -1406,96 +1469,95 @@ def partial(*args, **kwargs):
...
         path = PosixPath(bucket_name) / object_path_str
-        return f'gs://{path}'
+        return f"gs://{path}"
 
     @builder
     @decorators.immediate
-    def core__persistent_cache__local_store(
-                core__persistent_cache__flow_dir,
-            ):
+    def core__persistent_cache__local_store(core__persistent_cache__flow_dir,):
```
namanjain: super nit: comma at the end is pretty awkward now. This occurs quite a few times. Would you mind quickly replacing all `,)` with `)`?

bionic/__init__.py
```diff
@@ -1,9 +1,17 @@
 from .flow import Flow, FlowBuilder  # noqa: F401
 from .decorators import (  # noqa: F401
-    version, output, outputs, docs, gather, persist, memoize, pyplot,
-    immediate, changes_per_run
+    version,
```
namanjain: Not related to your PR but this really makes me want to add `isort` to sort the imports as well

setup.cfg
```diff
@@ -1,6 +1,7 @@
 [flake8]
 max-line-length=100
 exclude = docs,.venv
+ignore = E203, E501, W503, F811
```
namanjain: How do you feel about bumping `max-line-length` to 88 instead of ignoring `E501`? Check this section for details https://github.com/psf/black#line-length

.pre-commit-config.yaml
damienrj: Sure, I can add the black hook to this as well and see about updating the flake8 one.

.pre-commit-config.yaml
damienrj: Added, also it turns out that isort and black don't agree with each other.

setup.cfg
damienrj: The max is actually already set higher than 88, so could probably get rid of that error.

jqmp: I'd like to bring the line length down to 88, in order to maximize the number of side-by-side editor windows that can fit on a screen.

(Personally I was happy with 79, but I'm willing to give 88 a shot since it's the default.)

bionic/deriver.py
```diff
@@ -188,12 +188,12 @@ def _initialize_task_state(self, task_state):
             dep_state.children.append(task_state)
 
         # All names in this task should point to the same provider.
-        (task_state.provider,) = set(
+        (task_state.provider) = set(
```
namanjain: ah looks like black added brackets on the left side as well which weren't present before. Removing the comma before brackets made a lot of these variables a set which does not work anymore. I guess you will have to revert the `,)` to `)` change 😞.

setup.cfg
damienrj: Black is using the default 88, Just need to update the max-line-length argument for flake8

damienrj:
image

setup.cfg
```diff
@@ -1,6 +1,7 @@
 [flake8]
```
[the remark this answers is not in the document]
damienrj: Just using the default black arguments

damienrj:
> Looks good! (I mean, I don't like every choice it makes, but overall I think it will be a big benefit.)
> 
> Aside from the outstanding comments, we'll probably also want to update our contributing docs to mention `black` and our CI to run `black --check`, but I'm happy to do these myself in a later PR.
> 
> Thanks Damien!

Probably good to copy Pandas with their https://pandas.pydata.org/docs/development/contributing.html#python-pep8-black

jqmp:
> Probably good to copy Pandas with their https://pandas.pydata.org/docs/development/contributing.html#python-pep8-black

Yeah, I was thinking the same thing!

setup.cfg
```diff
@@ -1,6 +1,7 @@
...
-max-line-length=100
+max-line-length=88
 exclude = docs,.venv
+ignore = E203, E501, W503, F811, F401
```
jqmp: Just curious, what do we need F401 ("module not imported") for?

bionic/extras.py
jqmp: Just bumping this to make sure we add it before merging.

setup.cfg
damienrj: It didn't like how some of the decorators were used in bionic

PR #101, opened 2020-04-01
Reformat docstrings before rendering as tooltips

jqmp:
Raw docstrings don't look great as tooltips because of their line
breaks, so I added a simple algorithm that attempts to rewrap each
docstring before displaying it as a tooltip.

### DIFF
--- a/bionic/dagviz.py
+++ b/bionic/dagviz.py
@@ -24,7 +24,7 @@ class FlowImage:
         Replicates the PIL APIs for save and show, for both PIL-supported and SVG formats.
         """
         self._pil_image = Image.open(BytesIO(pydot_graph.create_png()))
-        self._xml_text = pydot_graph.create_svg()
+        self._xml_bytes = pydot_graph.create_svg()
 
     def save(self, fp, format=None, **params):
         """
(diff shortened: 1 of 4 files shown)


jqmp:
I'm still looking into the issue with the SVG diagrams not rendering in the docs.

namanjain:
> I'm still looking into the issue with the SVG diagrams not rendering in the docs.

I think we want to use [SVG-support-for-LaTeX](https://nbsphinx.readthedocs.io/en/0.5.1/markdown-cells.html#SVG-support-for-LaTeX) for the rendering issue.

jqmp:
I pushed a change that should fix the rendering of SVG images in our tutorial docs as well.

bionic/dagviz.py
```diff
@@ -53,7 +54,7 @@ def show(self):
 
     def _repr_svg_(self):
         """Rich display image as SVG in IPython notebook or Qt console."""
-        return str(self._xml_text)
+        return self._xml_text.decode("utf8")
```
namanjain, line 57: Nice find, this fix was easy enough! Q: why don't we pass encoding to `pydot_graph.create_svg()` instead of decoding it here?

jqmp: Thanks, this is a good point. I don't think there's any argument to `create_svg()` that will make it return a string, but I do think it's weird to have an `_xml_text` field that's actually bytes. I decided to keep it as bytes (since we also send it to file objects) but rename it.

PR #104, opened 2020-04-07
Create TaskStates already initialized

jqmp:
Instead of creating all the TaskStates in one pass and then initializing
them in a second pass, we now initialize each state as we create it.
This unblocks the ability to incrementally add more TaskStates over
time, rather than creating them all at once.

However, to do this we have to remove the bi-directional links between
states, since it's impossible to maintain them without going back and
editing existing states. So, I've removed the `children` field of
`TaskState`. This field is mainly useful for keeping track of which
states are waiting for the current one to finish; that logic has been
moved into a `BlockageTracker` class which is only used during
computation. This is part of a continued project of separating out the
per-flow-instance and per-get-call parts of EntityDeriver and TaskState.

### DIFF
--- a/bionic/deriver.py
+++ b/bionic/deriver.py
@@ -2,6 +2,8 @@
 Contains the core logic for resolving Entities by executing Tasks.
 """
 
+from collections import defaultdict
+
 from .datatypes import ProvenanceDigest, Query, Result, ResultGroup
 from .cache import Provenance
 from .descriptors import DescriptorNode
@@ -113,15 +115,9 @@ class EntityDeriver(object):
                     doc=self._docs_by_entity_name.get(entity_name),
                 )
 
-                for child_state in state.children:
-                    for child_task_key in child_state.task.keys:
-                        if not should_include_entity_name(
-                            child_task_key.dnode.to_entity_name()
-                        ):
-                            continue
-                        if task_key not in child_state.task.dep_keys:
-                            continue
-                        graph.add_edge(task_key, child_task_key)
+                for dep_state in state.dep_states:
+                    for dep_task_key in dep_state.task.keys:
+                        graph.add_edge(dep_task_key, task_key)
 
         return graph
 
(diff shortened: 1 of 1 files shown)


bionic/deriver.py
```diff
@@ -601,8 +591,74 @@ def __init__(self, task):
 
         self.is_complete = False
 
+    def incomplete_dep_states(self):
+        return [dep_state for dep_state in self.dep_states if not dep_state.is_complete]
+
+    @property
```
namanjain, line 597: `@property` is pretty neat.

bionic/deriver.py
```diff
@@ -150,52 +146,51 @@ def _get_ready_for_bootstrap_resolution(self):
...
-        self._task_states_by_key = {}
+        # Create a lookup table for all tasks.
+        self._tasks_by_key = {}
         for tasks in self._task_lists_by_dnode.values():
             for task in tasks:
-                task_state = TaskState(task)
-                for key in task.keys:
-                    self._task_states_by_key[key] = task_state
+                for task_key in task.keys:
+                    self._tasks_by_key[task_key] = task
```
namanjain, line 161: We might have talked about this before but to be sure, would task list still be while task states can change with new descriptors?

bionic/deriver.py
```diff
@@ -273,32 +268,32 @@ def _compute_result_group_for_dnode(self, dnode):
...
-                    and not child_state.is_blocked()
-                ):
-                    ready_task_states.append(child_state)
-                    blocked_task_key_tuples.remove(child_state.task.keys)
+            # See if we can unblock any other states now that we've completed this one.
+            unblocked_states = blockage_tracker.get_unblocked_by(state)
+            ready_task_states.extend(unblocked_states)
+
+        blocked_states = blockage_tracker.get_all_blocked_states()
+        assert not blocked_states, blocked_states
```
namanjain, line 295: I don't expect you to change this since `if not seq` is in [PEP 8 guide](https://www.python.org/dev/peps/pep-0008/) but I find it hard to read.

bionic/deriver.py
```diff
@@ -601,8 +591,74 @@ def __init__(self, task):
...
+
+        for completed_tk in completed_state.task.keys:
+            affected_blockages = self._blockage_lists_by_blocking_tk[completed_tk]
+            for blockage in affected_blockages:
+                blockage.mark_task_key_complete(completed_tk)
+                if blockage.is_resolved():
+                    yield blockage.blocked_state
+
+    def get_all_blocked_states(self):
+        return [
```
namanjain, line 659: I think this would duplicates since a blockage can be in multiple task keys. How do you feel about returning a set instead?

bionic/deriver.py
```diff
@@ -262,7 +257,7 @@ def _compute_result_group_for_dnode(self, dnode):
 
         ready_task_states = list(requested_task_states)
 
-        blocked_task_key_tuples = set()
+        blockage_tracker = BlockageTracker()
```
namanjain, line 260: How do you feel about extending this tracker to track ready task states as well?

bionic/deriver.py
```diff
@@ -601,8 +591,74 @@ def __init__(self, task):
...
+    @property
     def is_blocked(self):
-        return not all(parent.is_complete for parent in self.parents)
+        return len(self.incomplete_dep_states()) > 0
 
     def __repr__(self):
         return f"TaskState({self.task!r})"
+
+
+class Blockage:
```
namanjain, line 605: We can maybe say what this blocks, how about `TaskBlockage` / `TaskStateBlockage`?

bionic/deriver.py
```diff
@@ -150,52 +146,51 @@ def _get_ready_for_bootstrap_resolution(self):
...
-        dep_states = [self._task_states_by_key[dep_key] for dep_key in task.dep_keys]
-
-        for dep_state in dep_states:
-            self._initialize_task_state(dep_state)
-
-            task_state.parents.append(dep_state)
-            dep_state.children.append(task_state)
+    def _get_or_create_task_state_for_key(self, task_key):
+        task_state = self._task_states_by_key.get(task_key, None)
+        if task_state is not None:
```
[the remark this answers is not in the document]
jqmp: Yeah, I could go either way. I have a slight aversion to the repetition (and the double lookup) but it does save a line of code. I'll try changing it.

bionic/deriver.py
jqmp: Yep, tasks will still be immutable.

bionic/deriver.py
```diff
@@ -150,52 +146,51 @@ def _get_ready_for_bootstrap_resolution(self):
...
+            return task_state
 
+        task = self._tasks_by_key[task_key]
+        dep_states = [
+            self._get_or_create_task_state_for_key(dep_key) for dep_key in task.dep_keys
+        ]
         # All names in this task should point to the same provider.
-        (task_state.provider,) = set(
+        (provider,) = set(
```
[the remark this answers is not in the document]
jqmp: Sure, that sounds reasonable.

bionic/deriver.py
```diff
@@ -150,52 +146,51 @@ def _get_ready_for_bootstrap_resolution(self):
...
             self._flow_state.get_provider(task_key.dnode.to_entity_name())
             for task_key in task.keys
         )
         # And all the task keys should have the same case key.
-        (task_state.case_key,) = set(task_key.case_key for task_key in task.keys)
+        (case_key,) = set(task_key.case_key for task_key in task.keys)
+
+        task_state = TaskState(
+            task=task, dep_states=dep_states, provider=provider, case_key=case_key,
```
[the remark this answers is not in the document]
jqmp: Nope, Python won't match the names for you. If you write

    task_state = TaskState(task, dep_states, provider, case_key)

then Python doesn't care what the variable names are, just their ordering -- which in this case happens to be wrong. I added the redundant names so that I can't screw it without being obvious.

If Bionic were a language it would totally do this for you, though.

bionic/deriver.py
jqmp: Yeah, this is another point I go back and forth on. I'm happy to change it to `len(blocked_states) == 0`.

bionic/deriver.py
jqmp: Sure, that's reasonable. (In practice I doubt it will matter but it makes sense to avoid it.)

bionic/deriver.py
jqmp: That's a great question. I was planning to do this later, by breaking out a separate `FlowExecution` class, whose lifetime is confined to a single `get` call, and then refactoring it and merging `BlockageTracker` into it. The reason I did it in two stages was we definitely need to get rid of `children`, but I'm not yet 100% sure the full refactor will be a net improvement. On the other hand, the full refactor will probably restructure the blockage tracker stuff, so what you see here is kind of a transitional state.

bionic/deriver.py
namanjain: I think it's fine this way so we are consistent with the style guide, for later when someone else starts contributing to Bionic.

PR #107, opened 2020-04-16
Compute task/descriptor info incrementally

jqmp:
This commit makes it possible to compute task metadata incrementally and
lazily, instead of doing it all at once for all tasks during
initialization. This will be important later when we introduce more
complex descriptors, since the number of possible descriptors will be
infinite.

We still pre-compute all the "essential" descriptors ahead of time in
order to surface errors as early as possible.

I introduced a DescriptorInfo class to hold per-descriptor metadata.
Eventually descriptors will be 1:1 with tasks, so we'll want to refactor
this and TaskState.

Finally, I added docstrings for all the private methods in
EntityDeriver.

### DIFF
--- a/bionic/deriver.py
+++ b/bionic/deriver.py
@@ -4,6 +4,8 @@ Contains the core logic for resolving Entities by executing Tasks.
 
 from collections import defaultdict
 
+import attr
+
 from .datatypes import ProvenanceDigest, Query, Result, ResultGroup
 from .cache import Provenance
 from .descriptors import DescriptorNode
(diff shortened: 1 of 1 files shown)


bionic/deriver.py
```diff
@@ -140,38 +149,97 @@ def _get_ready_for_full_resolution(self):
...
+        """
+
+        for dnode in dnodes:
+            dinfo = self._get_or_create_dinfo_for_dnode(dnode)
+            for task_key, task in dinfo.tasks_by_key.items():
+                self._get_or_create_task_state_for_key(task_key)
+
+    def _get_essential_dnodes(self):
+        """
+        Returns the list of descriptor nodes needed to compute all user-defined entities
```
namanjain, line 182: Same here, I'm not sure what an "essential" dnode is. Once we have descriptors, I'm not sure what all user-defined entities would mean either. How do you feel about base / basic dnodes?

bionic/deriver.py
```diff
@@ -140,38 +149,97 @@ def _get_ready_for_full_resolution(self):
             "core__versioning_policy"
         )
 
-        self._is_ready_for_full_resolution = True
+        self._bootstrap_is_complete = True
+
+    def _prevalidate_essential_dnodes(self):
+        """
+        Checks that all 'essential' descriptors can be computed.
```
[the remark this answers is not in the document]
jqmp: The way I was thinking of it, the purpose of this function is just to confirm that the nodes *can* be constructed; the fact that the data us cached and reused is sort of an implementation detail. You can skip calling this function and the same setup will eventually happen, just later.

That said, I can see how it's confusing. Do you think more detail in the docstring would help, or should we change the name?

bionic/deriver.py
jqmp: Yeah, "essential" is not a great name. "Base" or "basic" might be better, although I worry that they suggest the dnodes just have a simple structure. "Core" might be good if we weren't already using it elsewhere. Let's talk more about it tomorrow morning.

PR #110, opened 2020-04-21
Allow parallel processing with a single subprocess

namanjain:
Executor and manager are both gated behind `core__parallel_processing__enabled` entity


tests/conftest.py
```diff
@@ -3,17 +3,53 @@
...
 
 # We provide this at the top level because we want everyone using FlowBuilder
 # to use a temporary directory rather than the default one.
 @pytest.fixture(scope="function")
-def builder(tmp_path):
+def builder(executor, manager, tmp_path):
     builder = bn.FlowBuilder("test")
     builder.set("core__persistent_cache__flow_dir", str(tmp_path / "BNTESTDATA"))
+
+    # NOTE: This is disabled till we are ready to switch to parallel processing for tests.
```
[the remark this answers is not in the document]
namanjain: I was thinking more like we will run all the tests with and without executor / manager using [parameterized](https://docs.pytest.org/en/latest/parametrize.html#pytest-generate-tests) builders. This would be the builder that takes the manager and there will be another one without manager.

tests/conftest.py
```diff
@@ -3,17 +3,53 @@
...
+def executor():
+    return get_reusable_executor()
+
+
+@pytest.fixture(scope="session")
+def manager():
+    class MyManager(SyncManager):
+        pass
+
+    # MyManager.register('Counter', ResettingCounter)
```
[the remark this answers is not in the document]
namanjain: Ah uncommented them

bionic/deriver.py
```diff
@@ -149,6 +151,10 @@ def _set_up_bootstrap(self):
             "core__versioning_policy"
         )
 
+        self._executor = self._bootstrap_singleton_entity("core__process_executor")
```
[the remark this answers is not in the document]
namanjain: I'll add the code that uses the executor in this PR based on your overall comment.

tests/conftest.py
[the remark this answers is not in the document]
jqmp: I agree, that makes sense. (Although we might still want the parameterization to be configurable, since this will significantly increase the running time of the tests.)

In the shorter term, though -- assuming you don't already have a PR ready to do the above -- it would be nice to be able to have an interface for this other than commenting/uncommenting code. Otherwise we'll have to keep remembering to uncomment the code, make changes, then re-comment before committing.

bionic/deriver.py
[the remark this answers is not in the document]
namanjain: Done

namanjain:
Addressed all the comments and added logic to execute the tasks using loky in the PR. I also changed the name of the PR since it doesn't reflect all the changes in the PR.

bionic/deriver.py
```diff
@@ -340,7 +340,29 @@ def _compute_result_group_for_dnode(self, dnode):
...
+                # using cloudpickle.
+                complete_task_state(state, task_key_logger)
+            else:
+                # NOTE 1: Logging support for multiple processes not done yet.
+                # NOTE 2: Non-persisted entities are not computed inside the executor yet.
+                # Right now the tests pass because all non-persisted entities are serializable
+                # but that won't always be the case.
+                ex = self._executor.submit(complete_task_state, state, task_key_logger)
+                ex.result()
+                state.is_complete = True
```
jqmp, line 362: Maybe we should set this at the end of `_populate_task_state_value_hash`? Strictly speaking, this shouldn't be true until `result_value_hashes_by_name` is set, right? (For example, suppose we threw an exception in between setting `is_complete` and setting `result_value_hashes_by_name`)?

tests/test_flow/test_protocols.py
```diff
@@ -28,44 +28,50 @@
 
 
 @pytest.mark.parametrize("value", PICKLABLE_VALUES)
-def test_picklable_value(builder, value):
+def test_picklable_value(builder, manager, value):
```
jqmp, line 31: It feels a little bit awkward to me that we need to use this `manager` object to create our counter objects, since the connection between a process manager and a counter is not very obvious. We could use a `make_counter` factory fixture to make the relationship clearer:

    @pytest.fixture
    def make_counter(manager):
        def _make_counter():
            return manager.ResettingCounter()

    def def test_picklable_value(builder, make_counter, value):
        counter = make_counter()

What do you think? Is this an improvement?

tests/conftest.py
```diff
@@ -3,17 +3,57 @@
...
+from bionic.deriver import TaskKeyLogger
+
+
+@pytest.fixture(scope="session")
+def executor():
+    return get_reusable_executor()
+
+
+@pytest.fixture(scope="session")
+def manager():
```
jqmp, line 22: Maybe this should be something like `process_manager` as well?

tests/test_flow/test_persistence_fuzz.py
```diff
@@ -120,15 +120,15 @@ def entity_is_nondeterministic(self, entity_name):
...
+        self.reset_called_entity_names()
         return names
 
     def peek_called_entity_names(self):
         names = list(self._last_called_names)
         return names
 
     def reset_called_entity_names(self):
-        self._last_called_names.clear()
+        self._last_called_names[:] = []
```
jqmp, line 131: Does `clear` not work?

bionic/deriver.py
```diff
@@ -7,9 +7,11 @@
 import attr
 
 from .datatypes import ProvenanceDigest, Query, Result, ResultGroup
+from .datatypes import ProvenanceDigest, Query, ResultGroup
```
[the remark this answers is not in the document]
namanjain: Oh gross merge conflict, will remove the previous line

bionic/cache.py
```diff
@@ -271,7 +271,10 @@ def _get_nearest_entry_with_artifact(self):
         return None
 
     def _get_local_entry(self):
-        if self._stored_local_entry is None:
+        if (
```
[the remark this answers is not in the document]
namanjain: Yeah, this one is a little gross but it was the quickest way to fix it. 

Any lookup done in the subprocess isn't communicated back so this really only happens when someone does a lookup in main process before the entity is computed and the entity is computed in subprocess. This only happens during assisted versioning when we try to load provenance for checking any versioning issues. When the cache doesn't exist, this entry is set to a not matched entry. When computed in subprocess, the updated entry set isn't populated back (which we do on reregister results by clearing stored entries) and any time someone does a get in main process, it returns a not matched entry with no result.

Some other ways we can consider are:
- peek provenance instead of loading provenance during versioning check.
- force reload entries inside the main process as soon as they are computed in subprocess. 

But to answer your question directly, this would increase the number of lookup at most once for an entity computed in subprocess but called in main process, and only when we are in assisted versioning mode.

bionic/flow.py
```diff
@@ -11,6 +11,9 @@
 from textwrap import dedent
 from uuid import uuid4
 
+from loky import get_reusable_executor
```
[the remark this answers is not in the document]
namanjain: oh yeah, my bad. Fixed.

bionic/deriver.py
```diff
@@ -340,7 +340,29 @@ def _compute_result_group_for_dnode(self, dnode):
...
+                # which we should not be really sending using IPC. We should read/write
+                # a tmp file for this instead to use protocol for serialization instead of
+                # using cloudpickle.
+                complete_task_state(state, task_key_logger)
+            else:
+                # NOTE 1: Logging support for multiple processes not done yet.
+                # NOTE 2: Non-persisted entities are not computed inside the executor yet.
+                # Right now the tests pass because all non-persisted entities are serializable
+                # but that won't always be the case.
+                ex = self._executor.submit(complete_task_state, state, task_key_logger)
```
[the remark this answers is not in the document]
namanjain: Yes, that's `NOTE 2` which is right above this line. I added comments in the code and also in the commit message after our discussion in the morning.

tests/test_flow/generate_test_compatibility_cache.py
```diff
@@ -25,43 +26,47 @@
...
         builder.set("core__persistent_cache__flow_dir", cache_dir)
         builder.assign("x", 2)
         builder.assign("y", 3)
         builder.assign("z", 4)
 
+        xy_counter = manager.ResettingCounter()
+
         @builder
-        @count_calls
+        @count_calls(xy_counter)
```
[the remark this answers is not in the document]
namanjain: oh your suggestion is much cleaner! I'll try it out.

tests/test_flow/test_protocols.py
namanjain: I didn't realize you can easily define factory methods in `pytest` like this. This is pretty neat, thanks!

tests/test_flow/test_persistence_fuzz.py
namanjain: So this will be a shared proxy list object when parallel processing is turned on.  The shared proxy list object does not have a clear method. I found this way of clearing it from this [stack overflow post](https://stackoverflow.com/questions/23499507/how-to-clear-the-content-from-a-listproxy).

tests/test_flow/generate_test_compatibility_cache.py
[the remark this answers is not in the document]
namanjain: For some reason, `counter.count_calls` isn't pickable.

bionic/deriver.py
namanjain: Yeah, that's fair. I'll move it after populating value hashes.

bionic/execution.py
```diff
@@ -0,0 +1,117 @@
+"""
```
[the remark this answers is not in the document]
namanjain: I like the idea of making these function methods on `TaskState`. Let's discuss this tomorrow as well.

bionic/deriver.py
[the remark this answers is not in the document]
jqmp: Yep, I saw `NOTE 2`, but I'm more concerned about the cost of serializing every single memoized value (including any persisted ancestors that also happen to be memoized). Not saying we need to fix this before merging, but it seems like a pretty salient drawback that isn't directly addressed by these notes.

tests/test_flow/test_persistence_fuzz.py
jqmp: Oh, weird! I guess the `ListProxy` API isn't up to date with the built-in `list` API.

bionic/cache.py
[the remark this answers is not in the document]
jqmp: I'm more concerned about the number of times this will happen during normal (non-parallel) operation. _get_local_entry gets called several times for each entity, so if the artifact is not there it will now attempt to look it up each time. This is not a huge deal but it does add a lot of little filesystem operations.

I like your "force reload" idea -- maybe by adding a method that flushes the local entries on an accessor. In general, it would be nice to keep this new parallel code as isolated as possible until it's ready for prime time.

namanjain:
Addressed comments and rebased to resolve merge conflicts. Ready for another review.

bionic/task_state.py
```diff
@@ -0,0 +1,306 @@
...
+        # But we cache the hashed values eagerly since they are cheap to load.
+        if provider.attrs.should_persist():
+            self.result_value_hashes_by_name = result_value_hashes_by_name
+
+    def sync_after_subprocess_completion(self):
+        """
+        Syncs the task state by populating and reloading data in the current process
+        after completing the task state in a subprocess.
+
+        Values populated in the task state are not communicated back from the subprocess.
```
jqmp, line 169: This sentence seems like it explains the previous paragraph rather than starting a new thought, so maybe it should start with "This is necessary because ..."?

bionic/task_state.py
```diff
@@ -0,0 +1,306 @@
...
+        """
+
+        # First, reload cache accessors
```
jqmp, line 172: This doesn't actually do any reloading, right? Can we say a bit more about why this is necessary?

Also, super nit: let's end each of these sentences with a period.

bionic/task_state.py
```diff
@@ -0,0 +1,306 @@
...
+        for accessor in self.cache_accessors:
+            accessor.flush_stored_entries()
+
+        # Then, populate the value hash
+        if self.result_value_hashes_by_name is None:
+            self.result_value_hashes_by_name = {}
+            for ix, query in enumerate(self.queries):
+                accessor = self.cache_accessors[ix]
+                value_hash = accessor.load_result_value_hash()
```
jqmp, line 181: Should we validate that this doesn't return `None`?

bionic/task_state.py
```diff
@@ -0,0 +1,306 @@
...
+        """
+
+        # First, reload cache accessors
+        for accessor in self.cache_accessors:
+            accessor.flush_stored_entries()
+
+        # Then, populate the value hash
+        if self.result_value_hashes_by_name is None:
+            self.result_value_hashes_by_name = {}
+            for ix, query in enumerate(self.queries):
```
jqmp, line 179: You could avoid using an index with `for query, accessor in zip(self.queries, self.cache_accessors)`. But it would be even simpler to do this, I think:

    for accessor in self.cache_accessors:
        value_hash = accessor.load_result_value_hash()
        self.result_value_hashes_by_name[
            accessor.query.dnode.to_entity_name()
        ] = value_hash

bionic/task_state.py
```diff
@@ -0,0 +1,306 @@
...
+        self.dep_states = dep_states
+        self.case_key = case_key
+        self.provider = provider
+
+        # These are set by complete(), just
+        # before the task state becomes eligible for cache lookup / computation.
+        #
+        # They will be present if and only if is_complete is True.
+        self.is_initialized = False
+        self.provenance = None
```
jqmp, line 25: Can any of these be made private?

bionic/task_state.py
```diff
@@ -0,0 +1,306 @@
...
+        self.task = task
+        self.dep_states = dep_states
+        self.case_key = case_key
+        self.provider = provider
+
+        # These are set by complete(), just
+        # before the task state becomes eligible for cache lookup / computation.
+        #
+        # They will be present if and only if is_complete is True.
+        self.is_initialized = False
```
jqmp, line 24: `is_initialized` will be present even if `is_complete` is false, right?

bionic/cache.py
```diff
@@ -104,6 +104,12 @@ def __init__(self, parent_cache, query):
         self._stored_local_entry = None
         self._stored_cloud_entry = None
 
+    def flush_stored_entries(self):
```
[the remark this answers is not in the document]
namanjain: Oh I can totally remove the private version and collapse it into this.

bionic/deriver.py
```diff
@@ -48,10 +46,7 @@ def __init__(self, flow_state, flow_instance_uuid):
         # The "bootstrap state" needs to be complete before we can compute user-defined
         # entities.
         self._bootstrap_is_complete = False
```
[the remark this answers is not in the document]
namanjain: Ah fair. I'm checking existence of bootstrap in task state initialization to see if it's complete, I should change it here too.

bionic/task_state.py
namanjain: Yeah, makes sense

bionic/task_state.py
namanjain: Oh accessor contains query too. That makes it easy.

bionic/task_state.py
namanjain: Oh yeah, good catch! This comment is now outdated. I'll update it.

bionic/task_state.py
namanjain: Oh yeah, so many of these can be made private. I'll make everything used only by `TaskState` private.

bionic/deriver.py
```diff
@@ -355,135 +372,25 @@ def _compute_result_group_for_dnode(self, dnode):
...
-
-        accessors_needing_saving = []
-        for accessor in task_state.cache_accessors:
-            old_prov = accessor.load_provenance()
-
-            if old_prov is None:
-                continue
 
-            new_prov = accessor.query.provenance
+class Bootstrap:
```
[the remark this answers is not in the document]
namanjain: of course!

bionic/task_state.py
namanjain: I added a check and threw `AssertionError` just in case. But this should never happen so no tests for this assertion failing.

PR #112, opened 2020-04-23
Janek/use attrs more

jqmp:
This refactors our simple data classes to use the `attrs` library. I also removed an unused class and a bunch of cases where we explicitly inherit from `object`, which is not necessary in Python 3.

### DIFF
--- a/bionic/datatypes.py
+++ b/bionic/datatypes.py
@@ -4,37 +4,39 @@ Contains various data structures used by Bionic's infrastructure.
 
 from collections import namedtuple
 
+import attr
+
 from .util import ImmutableSequence, ImmutableMapping
 
 
-# TODO Consider using the attr library here?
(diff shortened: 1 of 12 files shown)


bionic/datatypes.py
```diff
@@ -49,48 +51,54 @@ def __repr__(self):
...
     """
 
-    def __init__(self, task_key, protocol, provenance):
-        self.task_key = task_key
-        self.dnode = task_key.dnode
-        self.case_key = task_key.case_key
-        self.protocol = protocol
-        self.provenance = provenance
+    task_key = attr.ib()
+    case_key = attr.ib()
```
namanjain, line 61: I don't think you need this case_key since you have it as a property

jqmp: You're right -- thanks!

PR #114, opened 2020-04-24
Adding GeoPandas Dataframe protocol serializer

dodger487:


dodger487:
I will need to add tests to this as well. 

Another issue that I just discovered is that GeoPandas are based off of shapefiles (`.shp`) which impose a ten character limit on column lengths. If you have a column of longer than 10 characters, it is simply truncated. So we'd either need to add in special behavior (probably bad) or throw a warning or error if someone has column names of longer than ten characters in their dataframe.

jqmp:
This is really cool -- thank you for adding this!

Two other things we'll want to think about:

1. How will we actually import the GeoPandas module? We probably don't want GeoPandas to be a mandatory dependency for Bionic, so we should recover gracefully if it's not available. We could also consider adding it as an optional dependency, as documented [here](https://bionic.readthedocs.io/en/stable/get-started.html#extra-packages) and implemented [here](https://github.com/square/bionic/blob/master/bionic/extras.py#L19). The only drawback I can see to this is that we might need to add an explanation on how to handle GeoPandas's non-Python dependencies. (I'm not sure which of those dependencies are needed for just reading/writing shapefiles.)

2. Should we detect and handle GeoPandas automatically? We can do this by adding
a `validate` method to the protocol, and then adding the protocol [here].

jqmp:
>  If you have a column of longer than 10 characters, it is simply truncated. So we'd either need to add in special behavior (probably bad) or throw a warning or error if someone has column names of longer than ten characters in their dataframe.

Good point. I like the idea of throwing an exception here. The `ParquetDataFrameProtocol` has a similar problem with categorical problems and handles it the same way.

dodger487:
Not the cleanest github-foo here, but I found it easier to just create a new PR #122 rather than update this one.

PR #116, opened 2020-04-24
Janek/improve protocol interface

jqmp:
See commits for details.


bionic/protocols.py
```diff
@@ -48,33 +48,69 @@ def check_is_like_protocol(obj):
...
 
     def read(self, path):
+        """Deserializes an object from the pathlib path ``path``, and returns it."""
         raise NotImplementedError()
 
+    def read_with_extension(self, path, extension):
+        """
+        Deserializes an object from the pathlib path ``path``, and returns it.
+
+        May use the file extension ``extension`` to determine what deserialization
```
namanjain, line 108: nit: is file extension ``extension`` a typo?

bionic/protocols.py
```diff
@@ -86,15 +122,22 @@ def read(self, path):
...
+        """
+        Converts a valid value into a unique hash which can be used as a filename.
+        """
         if type(value) in self.SIMPLE_TYPES:
             return tokenization.tokenize(value)
         else:
             return tokenization.tokenize(value, self._write_to_bytes)
 
     def tokenize_file(self, path):
+        """Like ``tokenize``, but operates on a serialized value."""
```
namanjain, line 134: Can we be more specific here that the serialized value is coming from the contents of the path file (or dir)?

namanjain: Related to this, even I have trouble remembering if this path is a pathlib path. I remember seeing [typeguard](https://github.com/agronholm/typeguard) before for this, how do you feel about using for at least paths in bionic?

bionic/protocols.py
jqmp: No, I was trying to use the same pattern as "pathlib path ``path``" above. If I just wrote "the file ``extension``" , that would sound to me like ``extension`` was a file.

jqmp:
NOTE: for now, let's plan to avoid to merging this until https://github.com/square/bionic/pull/114 is merged.

jqmp:
Based on offline conversations, #114 isn't going get merged for a few more days, so I'm going to go ahead and merge this PR.

PR #117, opened 2020-04-26
Support unserializable entities for parallel processing

namanjain:
Stop sending any memoized results through IPC. Instead, use the
persisted cache for persistable entities and compute non-persisted
entities inside the subprocess.

The major benefit of this change is that Bionic can now support
using unserializable entities inside subprocess instead of throwing
an error during pickling the entity. It also reduces the IPC overhead
as reading persisted entities will use the Bionic Protocols which
are often tailored to the entities rather than using inefficient
cloud pickling for IPC.

This also changes the definition of complete for a task message.
Instead of being able to always retrieve the cached values, a completed
task state can also compute the value from its deps without having to
set up any task states.

### DIFF
--- a/bionic/task_state.py
+++ b/bionic/task_state.py
@@ -1,3 +1,5 @@
+import copy
+
 from .cache import Provenance
 from .datatypes import ProvenanceDigest, Query, Result
 from .exception import CodeVersioningError
@@ -47,8 +49,8 @@ class TaskState:
     def complete(self, task_key_logger):
         """
         Ensures that a task state reaches completion -- i.e., that its results are
-        available and can be retrieved. This can happen either by computing the task's
-        values or by confirming that cached values already exist.
+        available and can be retrieved or can be readily computed. The results can
+        be fetched from `self.get_results_assuming_complete()`.
         """
 
         assert self._is_initialized
@@ -74,9 +76,19 @@ class TaskState:
         self.is_complete = True
 
     def get_results_assuming_complete(self, task_key_logger):
-        "Returns the results of an already-completed task state."
+        """
+        Returns the results of an already-completed task state. This can happen
+        by returning the cached results if they exist, or by computing the results
+        using `self._compute()` and returning them.
+        """
 
         assert self.is_complete
+
+        # If task state should persist but results aren't cached, that's probably
+        # because the results aren't communicated between processes. Compute the
+        # results to populate in memory cache.
+        if not self.provider.attrs.should_persist() and not self._results_by_name:
+            self._compute(task_key_logger)
 
         if self._results_by_name:
             for task_key in self.task.keys:
@@ -317,3 +329,41 @@ class TaskState:
 
         for accessor in accessors_needing_saving:
             accessor.update_provenance()
+
+    # NOTE: This can be optimized further.
+    def new_state_for_completion(self, new_task_states_by_key):
+        """
+        Returns a copy of task state after keeping only the necessary data
+        required for completion. Along with keeping only necessary data for
+        completion, this also removes all the memoized results since they can
+        be expensive to serialize.
+
+        Mainly used to reduce IPC overhead when sending the state over to a
+        subprocess.
+
+        Parameters
+        ----------
+
+        new_task_states_by_key: Dict from key to new task states.
+            The cache for tracking new task states. Should be an empty dict when
+            first called.
+        """
+
+        # All task keys should point to the same task state.
+        if self.task.keys[0] in new_task_states_by_key:
+            return new_task_states_by_key[self.task.keys[0]]
+
+        # Let's make a copy of the task state.
+        # Note that this is not a deep copy so don't mutate so be careful when
+        # mutating state variables.
+        task_state = copy.copy(self)
+        task_state._results_by_name = None
+
+        new_dep_states = []
+        for dep_state in task_state.dep_states:
+            new_dep_state = dep_state.new_state_for_completion(new_task_states_by_key)
+            new_dep_states.append(new_dep_state)
+        task_state.dep_states = new_dep_states
+
+        new_task_states_by_key[task_state.task.keys[0]] = task_state
+        return task_state
--- a/bionic/task_state.py
+++ b/bionic/task_state.py
@@ -331,7 +331,7 @@ class TaskState:
             accessor.update_provenance()
 
     # NOTE: This can be optimized further.
-    def new_state_for_completion(self, new_task_states_by_key):
+    def new_state_for_completion(self, new_task_states_by_key, to_be_computed=True):
         """
         Returns a copy of task state after keeping only the necessary data
         required for completion. Along with keeping only necessary data for
@@ -347,6 +347,9 @@ class TaskState:
         new_task_states_by_key: Dict from key to new task states.
             The cache for tracking new task states. Should be an empty dict when
             first called.
+
+        to_be_computed: Boolean.
+            Whether the task state will be computed in the subprocess.
         """
 
         # All task keys should point to the same task state.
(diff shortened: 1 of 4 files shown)


tests/test_flow/test_persistence.py
```diff
@@ -21,6 +21,7 @@ def read(self, path, extension):
 # It would be nice to move the builder setup into fixtures, but since we need
 # to access the bound functions as well (to check the number of times they were
 # called), it's easiest to just have one long test.
+@pytest.mark.no_parallel
```
namanjain, line 24: These tests fail now because the memoized entities can be computed multiple times. This note in deriver should really have been added with this change.

> NOTE 1: This makes non-persisted entities compute here as well
   as in executor pool. We need to keep track of what needs to be
   computed in main process vs subprocess when entity is not persisted.

bionic/task_state.py
```diff
@@ -317,3 +329,54 @@ def _check_accessors_for_version_problems(self):
 
         for accessor in accessors_needing_saving:
             accessor.update_provenance()
+
+    # NOTE: This can be optimized further.
+    def new_state_for_completion(self, new_task_states_by_key, to_be_computed=True):
```
jqmp, line 334: It seems to me that the important thing about this method is not that it's *for completion* -- that's something that happens to every state regardless of the execution mode -- but rather that it *strips down* the state object in a way that allows it to be *transmitted* to *another process*. Maybe the name should reflect that?

bionic/task_state.py
```diff
@@ -317,3 +329,54 @@ def _check_accessors_for_version_problems(self):
...
+        # mutating state variables.
+        task_state = copy.copy(self)
+
+        # Clear up memoized cache to avoid sending it through IPC.
+        task_state._results_by_name = None
+
+        if to_be_computed:
+            new_dep_states = []
+            for dep_state in task_state.dep_states:
+                # We will need to compute the dependency in subprocess if it's not persisted.
```
jqmp, line 370: phrasing: "in subprocess" is not grammatical. Maybe "in the other subprocess"? Also, might be clearer to say "recompute" instead of "compute"?

Likewise a few lines down.

bionic/task_state.py
```diff
@@ -317,3 +329,54 @@ def _check_accessors_for_version_problems(self):
...
+        to_be_computed: Boolean.
+            Whether the task state will be computed in the subprocess.
+        """
+
+        # All task keys should point to the same task state.
+        if self.task.keys[0] in new_task_states_by_key:
+            return new_task_states_by_key[self.task.keys[0]]
+
+        # Let's make a copy of the task state.
+        # Note that this is not a deep copy so don't mutate so be careful when
```
jqmp, line 360: > so don't mutate so be careful when mutating

It's awkward to have two "so" clauses, especially since both are imperative but the first is really descriptive while the second is prescriptive. I think it's sufficient to say "this is not a deep copy so we'll avoid mutating any of the member variables".

bionic/task_state.py
```diff
@@ -317,3 +329,54 @@ def _check_accessors_for_version_problems(self):
 
         for accessor in accessors_needing_saving:
             accessor.update_provenance()
+
+    # NOTE: This can be optimized further.
+    def new_state_for_completion(self, new_task_states_by_key, to_be_computed=True):
+        """
+        Returns a copy of task state after keeping only the necessary data
+        required for completion. Along with keeping only necessary data for
```
jqmp, line 337: phrasing: You don't need to say "keeping only necessary data for completion" twice; you can just start the first sentence with "In addition,".

bionic/task_state.py
```diff
@@ -317,3 +329,54 @@ def _check_accessors_for_version_problems(self):
...
+        completion, this also removes all the memoized results since they can
+        be expensive to serialize.
+
+        Mainly used to reduce IPC overhead when sending the state over to a
```
jqmp, line 341: It's not just overhead, but also some things are simply impossible to serialize, right?

bionic/task_state.py
```diff
@@ -74,10 +76,20 @@ def complete(self, task_key_logger):
...
-        "Returns the results of an already-completed task state."
+        """
+        Returns the results of an already-completed task state. This can happen
+        by returning the cached results if they exist, or by computing the results
+        using `self._compute()` and returning them.
+        """
 
         assert self.is_complete
 
+        # If task state should persist but results aren't cached, that's probably
```
jqmp, line 87: Should this say "should *not* persist"? Also, "cached" is ambiguous here -- "memoized" may be clearer.

bionic/task_state.py
```diff
@@ -74,10 +76,20 @@ def complete(self, task_key_logger):
...
+        # because the results aren't communicated between processes. Compute the
+        # results to populate in memory cache.
```
jqmp, line 89: nit: s/populate in memory/populate the in-memory/

bionic/task_state.py
```diff
@@ -317,3 +329,54 @@ def _check_accessors_for_version_problems(self):
 
         for accessor in accessors_needing_saving:
             accessor.update_provenance()
+
+    # NOTE: This can be optimized further.
```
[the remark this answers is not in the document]
namanjain: Mostly if we can set provider as `None` once we remove provider attrs into task state. This might also benefit a lot from your tuple refactor to take stuff out from `Provider` into some sort of entity defn. Once we have a structure like that in place, we can remove usage of provider in task state completion. Also a few other stuff like setting task and queries as `None` when tasks won't be recomputed in subprocess.

bionic/task_state.py
```diff
@@ -47,8 +49,8 @@ def __repr__(self):
     def complete(self, task_key_logger):
         """
         Ensures that a task state reaches completion -- i.e., that its results are
-        available and can be retrieved. This can happen either by computing the task's
-        values or by confirming that cached values already exist.
+        available and can be retrieved or can be readily computed. The results can
+        be fetched from `self.get_results_assuming_complete()`.
```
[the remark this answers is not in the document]
namanjain: Yeah, `complete` is getting muddier as we add more complexity to computation logic. Let's talk about this too today.

bionic/task_state.py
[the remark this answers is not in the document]
namanjain: Ah I was using the default arg without knowing the memoized def argument gotcha and removed it altogether once you told me about it. I'll add this.

bionic/task_state.py
namanjain: Yeah this makes no sense, `so don't mutate` seems to be a mistake.

bionic/task_state.py
namanjain: oh yeah, good point. I'll call them both out here.

bionic/task_state.py
namanjain: Yeah, this should say should _not_ persist.

bionic/task_state.py
```diff
@@ -32,27 +34,44 @@ def __init__(self, task, dep_states, case_key, provider):
...
 
     def __repr__(self):
         return f"TaskState({self.task!r})"
 
     def complete(self, task_key_logger):
         """
         Ensures that a task state reaches completion -- i.e., that its results are
-        available and can be retrieved. This can happen either by computing the task's
-        values or by confirming that cached values already exist.
+        available and can be retrieved or can be readily computed. The results can
```
jqmp, line 69: is "can be readily computed" still applicable?

bionic/task_state.py
```diff
@@ -32,27 +34,44 @@ def __init__(self, task, dep_states, case_key, provider):
...
+    # would also match the definition of is_computable.
     def incomplete_dep_states(self):
         return [dep_state for dep_state in self.dep_states if not dep_state.is_complete]
 
     @property
-    def is_blocked(self):
-        return len(self.incomplete_dep_states()) > 0
+    def is_computable(self):
+        """
+        Indicates whether the task state can be computed without any additional
```
jqmp, line 52: "any additional setup" is vague and not really true, right?

Also, maybe we should include a note explaining why this definition is so complicated.

bionic/task_state.py
```diff
@@ -32,27 +34,44 @@ def __init__(self, task, dep_states, case_key, provider):
         # This can be set by get_results_assuming_complete() or _compute().
         self._results_by_name = None
 
+        # A completed task state has it's results computed and cached somewhere for
+        # easy retrieval.
         self.is_complete = False
 
+    # FIXME: We need a coherent story around incomplete states between parallel
```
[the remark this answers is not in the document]
namanjain: Yeah, that sgtm! I'll change it to a `TODO`.

bionic/task_state.py
```diff
@@ -32,27 +34,44 @@ def __init__(self, task, dep_states, case_key, provider):
...
         Ensures that a task state reaches completion -- i.e., that its results are
-        available and can be retrieved. This can happen either by computing the task's
-        values or by confirming that cached values already exist.
+        available and can be retrieved or can be readily computed. The results can
+        be fetched from `self.get_results_assuming_complete()`.
         """
 
         assert self._is_initialized
-        assert not self.is_blocked
+        assert self.is_computable
```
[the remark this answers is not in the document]
namanjain: ah fair point, I'll remove the first assert.

bionic/task_state.py
```diff
@@ -74,7 +93,11 @@ def complete(self, task_key_logger):
         self.is_complete = True
 
     def get_results_assuming_complete(self, task_key_logger):
-        "Returns the results of an already-completed task state."
+        """
+        Returns the results of an already-completed task state. This can happen
+        by returning the cached results if they exist, or by computing the results
+        using `self._compute()` and returning them.
```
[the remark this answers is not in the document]
namanjain: Nope, not anymore!

bionic/task_state.py
namanjain: Yeah this behavior changed too, I missed updating the existing docs

bionic/task_state.py
namanjain: Done

PR #118, opened 2020-04-27
Move attribute validation to TaskState setup

jqmp:
In a future refactor I want to separate the entity-specific
attributes (like `_can_memoize`) from the function-specific ones (like
`changes_per_run`), so we need to do this validation in a different
place.


jqmp:
> lgtm

Faster than Travis -- nice! :)

PR #119, opened 2020-04-28
Refactor ValueProvider to provide multiple entities jointly

jqmp:
(This PR has three commits, but most of the action is in the middle commit.)

Previously, when we defined entities jointly using `add_case`, each
entity was provided by a single ValueProvider. Now they are provided by
a single multi-valued ValueProvider. This makes them more consistent
with derived (function) entities.

Now it's the case that jointly-defined entities are always handled by a
single provider, which will simplify future refactoring.

There should be no functional changes in this commit (except the text of
exception messages, and perhaps some untested edge cases).

### DIFF
--- a/bionic/flow.py
+++ b/bionic/flow.py
@@ -122,7 +122,7 @@ class FlowState(pyrs.PClass):
         # If there's only one provider, then the grouping is already done.
         if len(providers_set) == 1:
             provider = providers[0]
-            if not set(provider.attrs.names) == set(names):
+            if set(provider.attrs.names) != set(names):
                 if len(names) == 1:
                     message = f"""
                     Can't assign to individual entity {names[0]!r}:
@@ -192,8 +192,7 @@ class FlowState(pyrs.PClass):
         if names != provider.attrs.names:
             values = [values[provider.attrs.names.index(name)] for name in names]
 
-        provider = provider.copy()
-        provider.add_case(case_key, values)
+        provider = provider.add_case(case_key, values)
 
         return self._set_provider(provider).touch()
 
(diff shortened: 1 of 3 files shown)


jqmp:
I added one more commit which makes ValueProvider immutable. Now all providers are immutable.

bionic/provider.py
```diff
@@ -156,24 +155,31 @@ def __init__(self, names, protocols, docs):
...
         )
 
-        self.clear_cases()
+        self.key_space = CaseKeySpace()
+        self._has_any_values = False
+        self._value_tuples_by_case_key = {}
+        self._token_tuples_by_case_key = {}
 
-    def copy(self):
+    def copy_and_add_case(self, case_key, values):
```
namanjain, line 165: now that all the providers are immutable, you can drop `copy_and_` since we know it will return a new object (much like how other immutable objects work in bionic, like flow state).

bionic/flow.py
```diff
@@ -115,12 +115,84 @@ def install_provider(self, provider):
...
-    def add_case(self, name, case_key, value):
-        provider = self.get_provider(name).copy_if_mutable()
+    def group_names_together(self, names):
+        providers = [self.get_provider(name) for name in names]
+        providers_set = set(providers)
+
+        # If there's only one provider, then the grouping is already done.
+        if len(providers_set) == 1:
+            provider = providers[0]
+            if not set(provider.attrs.names) == set(names):
```
namanjain, line 125: any specific reason to not use `!=` instead?

jqmp: Nope, just an oversight -- thanks!

bionic/provider.py
jqmp: Yeah... I guess that's consistent with how FlowState works. I was more explicit here because, unlike FlowState, providers don't have any other methods that behave like this, but I guess `add_case` is still pretty clear.

PR #120, opened 2020-05-01
Execute tasks in parallel

namanjain:
Move task tracking and compute logic out from EntityDeriver into a
separate class called TaskCompletionRunner.

TaskCompletionRunner follows the current stack based logic for single
process execution but now submits multiple tasks to process executor
and sees them to completion by also tracking the futures returned.

### DIFF
--- a/bionic/deriver.py
+++ b/bionic/deriver.py
@@ -3,6 +3,8 @@ Contains the core logic for resolving Entities by executing Tasks.
 """
 
 from collections import defaultdict
+from concurrent.futures import wait, FIRST_COMPLETED
+from enum import Enum, auto
 
 import attr
 
@@ -310,63 +312,12 @@ class EntityDeriver:
             self._get_or_create_task_state_for_key(task.keys[0]) for task in dinfo.tasks
         ]
 
-        ready_task_states = list(requested_task_states)
-
-        blockage_tracker = TaskBlockageTracker()
+        task_runner = TaskCompletionRunner(self._bootstrap, self._flow_instance_uuid)
 
         log_level = logging.INFO if self._bootstrap is not None else logging.DEBUG
         task_key_logger = TaskKeyLogger(log_level)
 
-        while ready_task_states:
-            state = ready_task_states.pop()
-
-            # If this task is already complete, we don't need to do any work.
-            if state.is_complete:
-                continue
-
-            # If blocked, let's mark it and try to derive its dependencies.
-            incomplete_dep_states = state.incomplete_dep_states()
-            if incomplete_dep_states:
-                blockage_tracker.add_blockage(
-                    blocked_state=state, blocking_states=incomplete_dep_states,
-                )
-                ready_task_states.extend(incomplete_dep_states)
-                continue
-
-            # Initialize the task state before attempting to complete it.
-            state.initialize(self._bootstrap, self._flow_instance_uuid)
-
-            # If the task isn't complete or blocked, we can complete the task.
-
-            # self._bootstrap.process_executor is None when parallel processing is
-            # disabled or when entities are bootstrapped.
-            if (
-                not state.provider.attrs.should_persist()
-                or self._bootstrap.process_executor is None
-            ):
-                # NOTE 1: This makes non-persisted entities compute here as well
-                # as in executor pool. We need to keep track of what needs to be
-                # computed in main process vs subprocess when entity is not persisted.
-                # NOTE 2: Right now, non-persisted entities include simple lookup values
-                # which we should not be really sending using IPC. We should read/write
-                # a tmp file for this instead to use protocol for serialization instead of
-                # using cloudpickle.
-                state.complete(task_key_logger)
-            else:
-                # NOTE 1: Logging support for multiple processes not done yet.
-                new_state_for_subprocess = state.strip_state_for_subprocess()
-                ex = self._bootstrap.process_executor.submit(
-                    new_state_for_subprocess.complete, task_key_logger
-                )
-                ex.result()
-                state.sync_after_subprocess_completion()
-
-            # See if we can unblock any other states now that we've completed this one.
-            unblocked_states = blockage_tracker.get_unblocked_by(state)
-            ready_task_states.extend(unblocked_states)
-
-        blocked_states = blockage_tracker.get_all_blocked_states()
-        assert not blocked_states, blocked_states
+        task_runner.run(requested_task_states, task_key_logger)
 
         for state in requested_task_states:
             assert state.is_complete, state
@@ -443,14 +394,14 @@ class DescriptorInfo:
         return self.tasks_by_key.values()
 
 
-class TaskBlockage:
+class EntryBlockage:
     """
     Represents a blocking relationship between a task state and a collection of
     not-yet-completed task keys it depends on.
     """
 
-    def __init__(self, blocked_state, blocking_tks):
-        self.blocked_state = blocked_state
+    def __init__(self, blocked_entry, blocking_tks):
+        self.blocked_entry = blocked_entry
         self._blocking_tks = set(blocking_tks)
 
     def mark_task_key_complete(self, blocking_tk):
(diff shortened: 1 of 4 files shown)


bionic/deriver.py
```diff
@@ -443,63 +393,201 @@ def tasks(self):
...
-            blocking_tk
-            for blocking_state in blocking_states
-            for blocking_tk in blocking_state.task.keys
+            self._visit_entry(entry, task_key_logger)
+
+        assert len(self._ready_entries) == 0
+        assert len(self._in_progress_entries) == 0
+        assert len(self._blocked_entries_by_blocking_tk) == 0
+
+    def _visit_entry(self, entry, task_key_logger):
```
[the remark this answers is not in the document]
namanjain: Yeah, I don't think I need visited anymore. Visited is basically the same as `IN_PROGRESS` or `COMPLETED` as of now. I'll get rid of this bit.

bionic/deriver.py
```diff
@@ -443,63 +393,201 @@ def tasks(self):
...
-        that are newly unblocked.
+        Handles entry transition and updates cached entries.
+
+        Parameters
+        ----------
+
+        entry: TaskRunnerEntry
+            Entry to transition.
+
+        transition_type: String with allowed values in [
```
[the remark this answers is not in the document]
namanjain: I was thinking of it as more of an action to apply rather than the next stage. The issue with keeping the next stage was that I had to keep it in sync with the task state status. For example, we don't have a blocked stage on runner entry. However, the code is much simplified now and might be easier to attach a stage enum to keep track of an entry's progress, like how you expanded on in your [other comment](https://github.com/square/bionic/pull/120#discussion_r419537877).

bionic/deriver.py
```diff
@@ -460,46 +411,194 @@ def is_resolved(self):
...
+            del self._in_progress_entries[completed_entry.state.task.keys[0]]
+
+        completed_entry.stage = EntryStage.COMPLETED
+        self._unblock_entries(completed_entry)
+
+    def _ready_blocking_entries(self, blocked_entry):
+        incomplete_dep_states = blocked_entry.state.incomplete_dep_states()
+        for incomplete_dep_state in incomplete_dep_states:
+            incomplete_entry = self._get_or_create_entry_for_state(incomplete_dep_state)
+            if incomplete_entry.stage == EntryStage.READY:
```
jqmp, line 565: It seems weird to me to call `_mark_entry_ready` on an entry that's already `READY`. Do you think we could make `entry.stage == READY` synonymous with `entry in self._ready_entries`?

Maybe it would be helpful to introduce a new stage or stages for entries that have just been created (and should be put on the ready stack) or have just been popped off the ready stack (and are about to be processed).

bionic/deriver.py
```diff
@@ -460,46 +411,194 @@ def is_resolved(self):
...
+    `TaskState` execution and tracking.
+    """
+
+    def __init__(self, state):
+        self.state = state
+        self.stage = EntryStage.READY
+        self.future = None
+
+
+class EntryStage(Enum):
```
jqmp, line 600: Maybe we should document what each stage means?

bionic/deriver.py
```diff
@@ -460,46 +411,194 @@ def is_resolved(self):
...
+    def _next_ready_entry(self):
+        assert len(self._ready_entries) != 0
+        return self._ready_entries.pop()
+
+    def _wait_on_in_progress_entries(self):
+        "Waits on any in-progress entry to finish."
+        futures = [
+            entry.future
+            for entry in self._in_progress_entries.values()
+            if entry.future is not None
```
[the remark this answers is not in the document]
namanjain: Oh yeah, it cannot. I'll remove this if conditional.

bionic/deriver.py
namanjain: This felt a little weird to me too but it's because `READY` is the default stage. I'll add a `NEW` stage. 

Also, now that I read this again, I feel that `ready` term is a bit overloaded in the whole runner. I'll rename `_ready_entries` to `_entries_to_process`.

bionic/deriver.py
```diff
@@ -460,46 +411,227 @@ def is_resolved(self):
...
+            completed_entry.future = None
+            del self._in_progress_entries[completed_entry.state.task.keys[0]]
+
+        completed_entry.stage = EntryStage.COMPLETED
+        self._unblock_entries(completed_entry)
+
+    def _ready_blocking_entries(self, blocked_entry):
+        incomplete_dep_states = blocked_entry.state.incomplete_dep_states()
+        for incomplete_dep_state in incomplete_dep_states:
+            incomplete_entry = self._get_or_create_entry_for_state(incomplete_dep_state)
```
jqmp, line 560: Are there any assertions we can/should make about this entry?

bionic/deriver.py
```diff
@@ -460,46 +411,227 @@ def is_resolved(self):
...
+    NEW = auto()
+
+    """
+    Entry is ready to be processed.
+    Valid next stages: [ACTIVE]
+    """
+    READY = auto()
+
+    """
+    Entry is actively running.
```
jqmp, line 615: "running" doesn't sound quite right to me, since the entry could end up being blocked. Could we say something like "We are actively attempting to start running this entry."?

bionic/deriver.py
```diff
@@ -460,46 +411,227 @@ def is_resolved(self):
...
+    ACTIVE = auto()
+
+    """
+    Entry is blocked by another entry(ies).
+    Valid next stages: [READY]
+    """
+    BLOCKED = auto()
+
+    """
+    Entry is currently processing.
```
jqmp, line 629: "running in another process"?

bionic/deriver.py
```diff
@@ -460,46 +411,227 @@ def is_resolved(self):
...
+
+
+class EntryStage(Enum):
+    """
+    Represents the stage of a `TaskRunnerEntry`.
+    """
+
+    """
+    Entry is created and hasn't been processed.
+    This is the first stage in processing.
```
jqmp, line 603: I think "processing" is being used to mean a few different things here. Since we have a `_process_entry` method, I'd avoid using the term for anything else. In this case, just saying "Entity was just created." seems sufficient.

bionic/deriver.py
```diff
@@ -460,46 +411,227 @@ def is_resolved(self):
...
+    Valid next stages: [READY]
+    """
+    NEW = auto()
+
+    """
+    Entry is ready to be processed.
```
jqmp, line 609: I know the name "ready" predates this PR, but I wonder if something like "waiting" is more accurate? "Ready" makes it sounds like the opposite of blocked. (And in particular, we say `ready to be processed` but we won't necessarily be ready to call `_process_entry` on it.)

bionic/deriver.py
```diff
@@ -460,46 +411,218 @@ def is_resolved(self):
...
-        blocking_tks = [
-            blocking_tk
-            for blocking_state in blocking_states
-            for blocking_tk in blocking_state.task.keys
-        ]
-        blockage = TaskBlockage(blocked_state, blocking_tks)
+        while self._has_more_to_process():
+            entry = self._next_entry_to_process()
+
+            if entry.stage != EntryStage.READY:
```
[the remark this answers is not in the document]
namanjain: This cannot happen, got rid of it.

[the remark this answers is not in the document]
jqmp: Sorry, I think this comment was left over from a review I started before our Slack conversation.

bionic/deriver.py
namanjain: This entry can be a new entry if first seen, or an already `READY` or `IN_PROGRESS` entry (if the entry blocks multiple entries and was added before). I guess it can't be active or completed since the task state wouldn't be incomplete but not sure if that requires an assertion. What do you think?

bionic/deriver.py
namanjain: I like `WAITING`!

bionic/deriver.py
jqmp: Maybe after the `if` block we could `assert incomplete_entry in (EntryStage.READY, EntryStage.IN_PROGRESS)`? Then we can be certain that we'll end up making progress on it.

bionic/deriver.py
namanjain: ah, we also use `futures.wait` and as a result, a method called `_wait_on_in_progress_entries`. I think `WAITING` would clash with that. I was `STAGED` but then realized we call this `EntryStage`. Any other suggestion?

PR #121, opened 2020-05-01
Set default pickle protocol to 4

EthanRosenthal:
In python versions 3.0-3.7, the default [pickle protocol](https://docs.python.org/3/library/pickle.html#data-stream-format) is `3`. Using this protocol, one cannot pickle files larger than 4GB. This PR changes the default pickle protocol used by bionic to `4` which allows for pickling larger files. `4` is also the default protocol for python 3.8 FWIW.

BTW I made the protocol a kwarg named `protocol`. Not sure if that's a name collision with all the other protocol stuff in bionic.

### DIFF
--- a/bionic/protocols.py
+++ b/bionic/protocols.py
@@ -198,14 +198,22 @@ class PicklableProtocol(BaseProtocol):
     """
     Decorator indicating that an entity's values can be serialized using the
     ``pickle`` library.
+
+    Parameters:
+        pickle_protocol_version: int (default: 4)
+            The pickle serialization protocol to use.
     """
+
+    def __init__(self, pickle_protocol_version=4):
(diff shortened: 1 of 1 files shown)


bionic/protocols.py
```diff
@@ -198,14 +198,22 @@ class PicklableProtocol(BaseProtocol):
     """
     Decorator indicating that an entity's values can be serialized using the
     ``pickle`` library.
+
+    Parameters:
+        protocol: int (default: 4)
```
jqmp, line 203: I don't think this name will break anything, but it may be a little bit confusing. WDYT about something like `pickle_protocol_version`?

jqmp:
This is great -- thanks!

PR #123, opened 2020-05-01
Stop unnecessary unserializable entity computation

namanjain:
For parallel processing mode, unserializable entities don't need to be
computed inside the main process unless returned. This change stops the
current unnecessary computation.

### DIFF
--- a/bionic/deriver.py
+++ b/bionic/deriver.py
@@ -429,16 +429,18 @@ class TaskCompletionRunner:
         self._pending_entries = []
         self._in_progress_entries = {}
         self._blockage_lists_by_blocking_tk = defaultdict(list)
+        self._requested_task_keys = set()
 
     def run(self, states, task_key_logger):
         for state in states:
+            self._requested_task_keys.update(state.task.keys)
             entry = self._get_or_create_entry_for_state(state)
             self._mark_entry_pending(entry)
 
         while self._has_pending_entries():
             entry = self._activate_next_pending_entry()
 
-            if len(entry.state.incomplete_dep_states()) > 0:
+            if len(entry.state.blocking_dep_states()) > 0:
                 self._mark_entry_blocked(entry)
                 continue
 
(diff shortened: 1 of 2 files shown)


bionic/deriver.py
```diff
@@ -459,29 +461,38 @@ def _process_entry(self, entry, task_key_logger):
         # Initialize the task state before attempting to complete it.
         state.initialize(self._bootstrap, self._flow_instance_uuid)
 
-        # self._bootstrap.process_executor is None when parallel processing is
-        # disabled or when entities are bootstrapped.
         if (
-            not state.provider.attrs.should_persist()
+            # Bootstrap entity.
```
jqmp, line 465: Nit: can we make these complete sentences, like "This is a bootstrap entity."?

bionic/task_state.py
```diff
@@ -41,43 +41,42 @@ def __init__(self, task, dep_states, case_key, provider):
...
-
-    @property
-    def is_completable(self):
+    # consider making the way it computes in parallel mode the default.
+    def blocking_dep_states(self):
         """
-        Indicates whether the task state can be completed in it's current shape.
-        True when the state is initialized, all of its persisted dependencies are
-        completed, and all its other dependencies are computable.
+        Returns all of it's persisted dependencies that aren't yet completed and
```
jqmp, line 47: s/it's/its/ -- or better yet, /this task state's/.

bionic/task_state.py
```diff
@@ -41,43 +41,42 @@ def __init__(self, task, dep_states, case_key, provider):
...
+        return set(
+            blocking_dep_state_tk
+            for blocking_dep_state in blocking_dep_states
+            for blocking_dep_state_tk in blocking_dep_state.task.keys
         )
 
+    @property
+    def is_completable(self):
+        """
+        Indicates whether the task state can be completed in it's current shape.
```
jqmp, line 75: s/it's/its/

bionic/deriver.py
```diff
@@ -459,29 +461,38 @@ def _process_entry(self, entry, task_key_logger):
...
             # a tmp file for this instead to use protocol for serialization instead of
             # using cloudpickle.
             state.complete(task_key_logger)
             self._mark_entry_completed(entry)
-        else:
-            # NOTE 1: Logging support for multiple processes not done yet.
+
+        # Process serializable entity in parallel.
+        elif state.provider.attrs.should_persist():
+            # NOTE: Logging support for multiple processes not done yet.
```
jqmp, line 484: Would it make sense for these `NOTE`s to be `TODO`s?

namanjain: Yeah, makes sense to me.

namanjain:
> Makes sense to me! I hope we can come back to the definitions of "blocking", "completable", and "complete" at some point though. :)

Definitely, that's why I left a TODO :)

PR #124, opened 2020-05-05
Support unset values cleanly

### LINKED ISSUE #103: Errors from unset entities are hard to debug
If you try to compute an entity value but one of its ancestors doesn't have a value set, the error message is not very helpful. It just looks like this: `ValueError: Entity 'MY_ENTITY' has no defined values`. We should make it easier to figure out what the problem is and how to resolve it.

Probably the most straightforward solution would be to attach an explanation that describes which of the ancestors are not set.

jqmp:
(I've updated this text to match the current commit message, since
I was able to make a major simplification thanks to a great suggestion
by @namanjain!)

This commit improves the handling of unset entities. An entity is
"unset" if it meets any of the following criteria:

1. It has been declared but not set.
2. It has been set with an empty list of values.
3. Any of its ancestors is unset.

(Note that this conflates the concepts of "empty list" and "null value".
This conflation is deliberate and based on the idea that Bionic should
automatically convert between these two representations. In the future
the user will be able to distinguish between these two concepts using
descriptors, but Bionic will still automatically translate between them
when the result is unambiguous.)

I've fixed three outstanding problems with unset entities:

1. They didn't appear in the DAG visualization. Now they do.

2. When the user tried to compute their value, they got a vague error
message (see #103). Now they get
a helpful error message describing exactly which entities are unset.

3. `@gather` didn't handle them properly. Now it does: an entity with an
empty list of values can be gathered into an empty dataframe. (However,
emptiness/unset-ness can also be propagated when the also= argument is
used. As always with `@gather`, the details are subtle. I added several
gathering tests for this.)

Architecturally, this required the following changes:

1. CaseKeys can now include "missing" values, where "missing" is
represented by the sentinel constant `CaseKey.MISSING`.
2. A Task or Result is now considered "missing" if its CaseKey has any
missing values. Any Task with missing inputs must also have a missing
output. When a Task has a missing output, we don't compute it; instead
we generate a missing Result. Missing values are filtered out at the
last minute during `get` and `gather` operations.
3. I refactored some of the code in GatherProvider to be a little
clearer. (However, that provider's code is still insanely complex.)

Fixes #103

### DIFF
--- a/bionic/deriver.py
+++ b/bionic/deriver.py
@@ -217,8 +217,14 @@ class EntityDeriver:
             for dep_dinfo in dep_dinfos
         }
 
+        # TODO Maybe of having these two separate variables that we pass around, we
+        # should just have a single method:
+        #
+        #     provider.get_dinfo(dep_dinfos_by_dnode)
         key_space = provider.get_key_space(dep_key_spaces_by_dnode)
-        tasks = provider.get_tasks(dep_key_spaces_by_dnode, dep_task_key_lists_by_dnode)
+        tasks = provider.get_tasks(
+            dep_key_spaces_by_dnode, dep_task_key_lists_by_dnode,
+        )
         tasks_by_key = {
             task_key: task
             for task in tasks
@@ -299,7 +305,18 @@ class EntityDeriver:
                 value; got {len(values)} ({values!r})"""
                 )
             )
-        return result_group[0].value
+        (result,) = result_group
+        if result.value_is_missing:
+            raise ValueError(
+                oneline(
+                    f"""
+                Bootstrap entity {entity_name!r} could not be computed because
+                the following entities are declared but not set:
+                {", ".join(result.query.case_key.missing_names)}
+                """
+                )
+            )
+        return result.value
 
     def _compute_result_group_for_dnode(self, dnode):
         """
@@ -383,6 +400,15 @@ class TaskKeyLogger:
 class DescriptorInfo:
     """
     Holds useful metadata about a descriptor.
+
+    Attributes
+    ----------
+    dnode: DescriptorNode
+        The descriptor this object refers to.
+    key_space: CaseKeySpace
+        Each of this descriptor's tasks' CaseKeys will have this key space.
+    tasks_by_key: dict from TaskKey to Task
+        All the tasks for this decriptor, organized by TaskKey.
     """
 
     dnode = attr.ib()
@@ -468,7 +494,7 @@ class TaskCompletionRunner:
             or self._bootstrap.process_executor is None
             # This is a non-serializable entity that needs to be returned.
             or (
-                not state.provider.attrs.should_persist()
+                not state.should_persist()
                 and entry.state.task.keys[0] in self._requested_task_keys
             )
         ):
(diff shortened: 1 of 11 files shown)


jqmp:
Looks like some of the tests don't work in parallel mode. I'll look into this.

jqmp:
> Looks like some of the tests don't work in parallel mode. I'll look into this.

This is fixed.

bionic/deriver.py
```diff
@@ -522,6 +549,7 @@ def _wait_on_in_progress_entries(self):
         for finished_future in finished_futures:
             task_key = finished_future.result()
             entry = self._entries_by_task_key[task_key]
+            print("SYNC", entry.state.task.keys[0], entry.state.should_persist())
```
namanjain, line 552: remove print statement

bionic/task_state.py
```diff
@@ -77,6 +77,17 @@ def is_completable(self):
         """
         return self._is_initialized and len(self.blocking_dep_states()) == 0
 
+    # TODO Now that we've introduced this method, let's remove
+    # provider.attrs.should_persist(), which is just a wrapper for
+    # provider.attrs._can_persist anyway.
+    def should_persist(self):
+        return self.provider.attrs.should_persist() and not self.has_missing_output()
+
+    def has_missing_output(self):
```
namanjain, line 86: It's not very clear what missing output means here. Maybe `missing_dep_outputs`?

bionic/task_state.py
```diff
@@ -168,6 +179,26 @@ def _compute(self, task_key_logger):
 
         dep_values = [dep_result.value for dep_result in dep_results]
 
+        # If we have any missing outputs, exit early with a missing result.
+        if self.has_missing_output():
```
namanjain, line 183: I think this is fine here for now but we should consider moving this logic for entries that exit early like this one and simple lookups somewhere else.

bionic/util.py
```diff
@@ -81,6 +81,15 @@ def single_element(iterable):
     return next(iter(iterable))
```
namanjain, line 81: can you remove this line too?

tests/test_flow/test_gather.py
```diff
@@ -213,6 +417,72 @@ def summary(df1, df2):
     }
 
 
+def test_multiple_gathers__unset_x(preset_builder):
+    builder = preset_builder
+
+    builder.clear_cases("x")
+
+    @builder
+    @bn.gather("w", "wx", "df1")
```
namanjain, line 426: I didn't realize we can use gather multiple times like this. This is pretty cool!

bionic/deriver.py
jqmp: Oops! Thank you!

bionic/task_state.py
jqmp: This actually means that the *output* of the task has (or would have) a missing value. Maybe `output_would_be_missing`?

bionic/task_state.py
jqmp: Simple lookups don't exit early, right? They get handled by `task.compute` like everything else.

That said, yes, as we add more "generic" rules, like with tuples, we should think about how to handle them in a uniform way.

bionic/util.py
jqmp: Oh yeah! Good catch.

PR #125, opened 2020-05-08
Reduce IPC cost for parallel processing

namanjain:
Reduce the amount of data sent to process executor by removing provider
for all task states in task state strip down. Also removes task,
queries and provenance when not required.

### DIFF
--- a/bionic/deriver.py
+++ b/bionic/deriver.py
@@ -494,7 +494,7 @@ class TaskCompletionRunner:
             or self._bootstrap.process_executor is None
             # This is a non-serializable entity that needs to be returned.
             or (
-                not state.should_persist()
+                not state.should_persist
                 and entry.state.task.keys[0] in self._requested_task_keys
             )
         ):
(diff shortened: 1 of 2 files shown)


bionic/deriver.py
```diff
@@ -468,7 +468,7 @@ def _process_entry(self, entry, task_key_logger):
             or self._bootstrap.process_executor is None
             # This is a non-serializable entity that needs to be returned.
             or (
-                not state.provider.attrs.should_persist()
+                not state.should_persist
```
namanjain, line 471: This is going to conflict with your PR. I'll rebase my PR on latest master once you merge your PR.

bionic/task_state.py
```diff
@@ -396,28 +390,34 @@ def strip_state_for_subprocess(self, new_task_states_by_key=None):
...
 
         # Any non-persisted value will have to be recomputed again in the other
         # subprocess and isn't complete anymore.
-        if not task_state.provider.attrs.should_persist():
+        if not task_state.should_persist:
             task_state.is_complete = False
 
         if task_state.is_complete:
             task_state.dep_states = []
+            task_state.task = None
```
jqmp, line 414: It's getting even harder to reason about what states a `TaskState` can be in now....

Not a blocker for this PR, but something to keep an eye on.

namanjain:
> Any idea how much of an improvement this is?

Test time for parallel processing tests roughly went down from 11.5 mins to 6.5 mins

PR #126, opened 2020-05-11
Fix bug breaking the use of fixed dataframes

jqmp:
In commit 3b86ba0fdbc83a4f95bf14b84e3d030365313bcf, I added some code
that applied `==` to user-provided values. This was a bad idea because some
classes (like pandas.DataFrame) override `==` and return a value that
can't be interpreted as a boolean. Thankfully this caused our
documentation build to fail.

I changed the way we do this comparison and added a test to catch this.

### DIFF
--- a/bionic/datatypes.py
+++ b/bionic/datatypes.py
@@ -202,7 +202,9 @@ class CaseKey(ImmutableMapping):
         self.missing_names = [
             name
             for name, value, token in name_value_token_triples
-            if value == self.MISSING
+            # We don't want to say `value == MISSING` because `value` might override the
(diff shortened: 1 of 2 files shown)


bionic/datatypes.py
```diff
@@ -202,7 +202,9 @@ def __init__(self, name_value_token_triples):
         self.missing_names = [
             name
             for name, value, token in name_value_token_triples
-            if value == self.MISSING
+            # We don't want to say `value == MISSING` because `value` might override the
```
namanjain, line 205: TIL!

Issue #127, opened 2020-05-11
The "provider" concept is too broad


jqmp:
Currently Bionic's internals use a concept of a "provider" object which contains the logic for computing each entity value. This class (or family of classes) handles two categories of things:

1. Generate the task objects responsible for applying a particular data transformation. Each provider applies exactly one transformation. (That transformation may be composed of several others, but it's consistent for a given provider). 
2. Serve as the source of truth for entity properties (names, docstrings, protocols, etc.). A provider is usually responsible for one entity, but can sometimes have more.

The fact that a provider is not 1:1 with an entity suggests that we should separate these two concepts. In fact, as we add non-entity descriptors, we'll soon want to apply transformations that aren't associated with any entity at all. (For example, constructing or deconstructing a tuple.)

Furthermore, providers are structured in such a way that they can be incrementally transformed by decorator functions. This means that during the decoration process, a provider can have an incomplete or invalid configuration, with the expectation that it will be cleaned up or rejected by the final decorator prior to being added to a Flow. This need to support both clean and unclean states makes the provider interface and data model even more complicated.

To address this two issues, I propose splitting out two additional classes

1. An `EntityDefinition` class which acts as the source of truth for entity properties. This object is immutable and always in a clean, valid state.
2. A `DecorationAccumulator` class which holds the messy entity configuration data built up by Bionic decorators. When a decorated function is incorporated into a flow, this accumulator will be validated, normalized, and converted into a provider and one or more `EntityDefinition`s.

The provider concept will remain, but it will only be responsible for computing values by applying transformations.

PR #128, opened 2020-05-11
Refactor Bionic decorators

jqmp:
This is the first step towards addressing #127.

This changes the way Bionic decorators are implemented. Previously each
decorator worked by accepting either a function or a provider, and
wrapping it in a new provider. Now they accept a function and return it
after modifying a special `bionic_decorator_accumulator` attribute.
This is conceptually cleaner (no more `func_or_provider`) and introduces
the `DecorationAccumulator` class described in the issue linked above.

### DIFF
--- a/bionic/decoration.py
+++ b/bionic/decoration.py
@@ -1,0 +1,63 @@
+"""
+Code for creating and applying Bionic decorators.
+
+Bionic decorators are expected to be applied like this:
+
+    @builder
+    @decorator1
+    @decorator2
+    def func(arg1, arg2, ...):
+        ...
+
+Each decorator attaches information to the decorated function by by creating or updating
+a DecorationAccumulator object, set as an attribute on the function. The assumption is
+that the decorator at the top, (``@builder``) will be a FlowBuilder object which removes
+this accumulator object and uses it to define a new entity.
+"""
+
+import attr
+
+from .provider import FunctionProvider
+
+
+@attr.s
+class DecorationAccumulator:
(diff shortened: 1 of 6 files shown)


bionic/decoration.py
```diff
@@ -0,0 +1,61 @@
...
+
+        return func
+
+    return decorator
+
+
+ACC_ATTR_NAME = "bionic_decorator_accumulator"
+
+
+def init_accumulator_if_not_set_on_func(func):
```
namanjain, line 47: nit: you can make this a private method as well

bionic/decoration.py
```diff
@@ -0,0 +1,61 @@
...
+def decorator_wrapping_provider(wrapper_fn, *args, **kwargs):
+    """
+    Returns a decorator which transforms the accumulated provider on a decorated
+    function by applying the given wrapping function and arguments.
+    """
+
+    def decorator(func):
+        init_accumulator_if_not_set_on_func(func)
+        acc = get_accumulator_from_func(func)
+        acc.provider = wrapper_fn(acc.provider, *args, **kwargs)
```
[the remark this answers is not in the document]
jqmp: I updated this. I can't decide if I like it more or not, so I'll leave it the way you suggested it. :)

bionic/decoration.py
jqmp: This is also called directly in `FlowBuilder.__call__`.

Also, meta-nit: it's a function, not a method. 😄

PR #132, opened 2020-05-12
Mark fixed entities as _can_persist

jqmp:
Previously fixed entities had `_can_persist=False`, which sort of makes
sense since we don't actually want to persist them. However, since we do
expect fixed entities to be serializable, it's not really true that we
*can't* persist them. With this change, we now mark them as
`_can_persist=True`, but we also update `TaskState.should_persist` to
be `False` for fixed entities, so all behavior stays the same.

The upshot is that now all entities default to being persistable. This
makes it easier to break out a separate immutable entity definition
class, since the persistence properties don't change when you convert a
derived entity to a fixed one.

### DIFF
--- a/bionic/provider.py
+++ b/bionic/provider.py
@@ -152,7 +152,7 @@ class ValueProvider(BaseProvider):
                 names=names,
                 protocols=protocols,
                 docs=docs,
-                can_persist=False,
+                can_persist=True,
                 can_memoize=True,
                 changes_per_run=False,
             ),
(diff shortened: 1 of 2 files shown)


jqmp:
> This is fine for now but I think we should consider adding a `is_serializable` entity attribute and stop using persistable as a proxy for it.

I agree, but I feel like it's not worth it until we also have a way to say "this may be serializable but I don't want you to persist it". If we're just renaming the decorator, I don't think the increase in clarity outweighs the annoyance of the API change. But if we can replace `@persist` with two more specific versions, it makes sense to ask users to change their code and clarify which one they meant.

WDYT?

PR #134, opened 2020-05-13
Convert namedtuple to attr classes

namanjain:
Namedtuples were used before we introduced attr classes in bionic. Now
that we have attr classes, there is no real reason to keep using
namedtuples since we don't want those classes as a tuple.


namanjain:
> Sorry, lost track of this PR. Thanks for doing this!
> 
> BTW, WDYT about using `attrs` classes for these?

Changed it to attr classes as we discussed.

PR #136, opened 2020-05-14
Janek/entity definitions

### LINKED ISSUE #127: The "provider" concept is too broad

jqmp:
Fixes #127

This introduces a new `EntityDefinition` class with the following fields:
- `name`
- `protocol`
- `doc`
- `can_persist`
- `can_memoize`

These fields replace the corresponding fields on `ProviderAttributes`
(except for `names`, which still exists). Corresponding fields have also
been added to `DecorationAccumulator`, which is now responsible for
propagating these settings during the decoration process.

There should be no functional changes, aside from the wording in some
error messages and the use of `AttributeValidationError` in some cases
where we used to use `ValueError`.

### DIFF
--- a/bionic/decorators.py
+++ b/bionic/decorators.py
@@ -11,7 +11,8 @@ These are the decorators we expose to Bionic users.  They are used as follows:
 """
 
 from .datatypes import CodeVersion
-from .decoration import decorator_wrapping_provider
+from .decoration import decorator_updating_accumulator
+from .exception import AttributeValidationError
 from .provider import (
     GatherProvider,
     AttrUpdateProvider,
@@ -19,6 +20,7 @@ from .provider import (
     RenamingProvider,
     NameSplittingProvider,
 )
+from .util import oneline
 from . import interpret
 
 
@@ -59,8 +61,10 @@ def version(major=None, minor=None):
         A decorator which can be applied to an entity function.
     """
 
-    return decorator_wrapping_provider(
-        AttrUpdateProvider, "code_version", CodeVersion(major, minor)
+    return decorator_updating_accumulator(
+        lambda acc: acc.wrap_provider(
+            AttrUpdateProvider, "code_version", CodeVersion(major, minor)
+        )
     )
 
 
@@ -87,7 +91,9 @@ def persist(enabled):
     if not isinstance(enabled, bool):
         raise ValueError(f"Argument must be a boolean; got {enabled!r}")
 
-    return decorator_wrapping_provider(AttrUpdateProvider, "_can_persist", enabled)
+    return decorator_updating_accumulator(
+        lambda acc: acc.update_attr("can_persist", enabled, "@persist")
+    )
 
 
 def memoize(enabled):
@@ -109,7 +115,9 @@ def memoize(enabled):
     if not isinstance(enabled, bool):
         raise ValueError(f"Argument must be a boolean; got {enabled!r}")
 
-    return decorator_wrapping_provider(AttrUpdateProvider, "_can_memoize", enabled)
+    return decorator_updating_accumulator(
+        lambda acc: acc.update_attr("can_memoize", enabled, "@memoize")
+    )
 
 
 def changes_per_run(enabled=None):
@@ -189,7 +197,9 @@ def changes_per_run(enabled=None):
     if not isinstance(enabled, bool):
         raise ValueError(f"Argument must be a boolean; got {enabled!r}")
 
-    return decorator_wrapping_provider(AttrUpdateProvider, "changes_per_run", enabled)
+    return decorator_updating_accumulator(
+        lambda acc: acc.wrap_provider(AttrUpdateProvider, "changes_per_run", enabled)
+    )
 
 
 def output(name):
@@ -211,7 +221,9 @@ def output(name):
         A decorator which can be applied to an entity function.
     """
 
-    return decorator_wrapping_provider(RenamingProvider, name)
+    return decorator_updating_accumulator(
+        lambda acc: acc.wrap_provider(RenamingProvider, name)
+    )
 
 
 def outputs(*names):
@@ -239,7 +251,9 @@ def outputs(*names):
         A decorator which can be applied to an entity function.
     """
 
-    return decorator_wrapping_provider(NameSplittingProvider, names)
+    return decorator_updating_accumulator(
+        lambda acc: acc.wrap_provider(NameSplittingProvider, names)
+    )
 
 
 def docs(*docs):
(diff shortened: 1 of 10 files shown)


bionic/flow.py
```diff
@@ -254,22 +302,40 @@ def mark_all_providers_default(self):
...
 
-    def _erase_provider(self, name):
+    def _erase_entity_def(self, name):
+        assert name not in self.providers_by_name
+
         state = self
-        if name in state.providers_by_name:
-            state = state.set(providers_by_name=state.providers_by_name.remove(name))
+        state = state.set(entity_defs_by_name=state.entity_defs_by_name.remove(name),)
         if name in state.default_provider_names:
```
namanjain, line 310: it feels weird that you are deleting `default_provider_names` inside `_erase_entity_def`. Let's talk about this on monday.

bionic/flow.py
```diff
@@ -820,69 +896,75 @@ def __call__(self, func):
...
-        if provider.attrs.protocols is None:
-            provider = ProtocolUpdateProvider(provider, DEFAULT_PROTOCOL)
-        if provider.attrs.docs is None:
-            docs = [None] * len(provider.attrs.names)
-            provider = AttrUpdateProvider(provider, "docs", docs)
-        if provider.attrs._can_persist is None:
-            provider = AttrUpdateProvider(provider, "_can_persist", True)
-        if provider.attrs._can_memoize is None:
-            provider = AttrUpdateProvider(provider, "_can_memoize", True)
-        if provider.attrs.changes_per_run is None:
```
namanjain, line 834: You are missing this case. I think it works because if `changes_per_run` is not set (and is `None` by default), it's the same as `False` everywhere wherever Bionic is using it. 

I think this is another indicator that `changes_per_run` should be a part of entity defn.

bionic/flow.py
```diff
@@ -820,69 +896,75 @@ def __call__(self, func):
...
+        }
 
         state = self._state
 
+        # Delete the original definitions.
         state = state.delete_providers(provider.attrs.names)
+        state = state.delete_entity_defs(provider.attrs.names)
+
+        # Create the new definitions.
+        for name in provider.attrs.names:
```
namanjain, line 966: nit: I think this would be easier to read if you iterate over `entity_defs_by_name.values()` and define them rather than using `provider.attrs.names` for iteration.

bionic/protocols.py
```diff
@@ -186,8 +185,16 @@ def __call__(self, func=None, **kwargs):
                     )
                 )
 
-            wrapper = decorator_wrapping_provider(ProtocolUpdateProvider, self)
-            return wrapper(func)
+            def update_accumulator(acc):
```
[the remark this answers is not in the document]
jqmp: You're right! I lost track of the fact that we weren't checking this before. I agree that it's unlikely anyone is doing this but it's good to avoid breakage on principle. I'll change this to a warning.

bionic/decorators.py
```diff
@@ -189,7 +210,10 @@ def summary(current_data):
     if not isinstance(enabled, bool):
         raise ValueError(f"Argument must be a boolean; got {enabled!r}")
 
-    return decorator_wrapping_provider(AttrUpdateProvider, "changes_per_run", enabled)
+    def update_accumulator(acc):
+        acc.wrap_provider(AttrUpdateProvider, "changes_per_run", enabled)
```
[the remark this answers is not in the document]
jqmp: I see `changes_per_run` as a property of the *function*, not of the *entity*. It's describing the way the value is computed, but not the value itself.

bionic/flow.py
jqmp: You're right, this should now probably be called `default_entity_names`.

bionic/flow.py
jqmp: Good catch! Actually I think we should have some kind of `FunctionAttributes` object that holds `version`, `changes_per_run`, etc.

bionic/flow.py
jqmp: You're right!

bionic/decorators.py
```diff
@@ -87,7 +90,16 @@ def persist(enabled):
     if not isinstance(enabled, bool):
         raise ValueError(f"Argument must be a boolean; got {enabled!r}")
 
-    return decorator_wrapping_provider(AttrUpdateProvider, "_can_persist", enabled)
+    def update_accumulator(acc):
+        if acc.can_persist is not None:
```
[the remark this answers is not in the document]
jqmp: I originally did this to have nicer error messages than what we had with `AttrUpdateProvider`; however, I'll see if I can do this with a helper without making the function signature too ugly.

bionic/flow.py
```diff
@@ -102,18 +123,24 @@ def has_provider(self, name):
         return name in self.providers_by_name
 
     def create_provider(self, name, protocol, doc):
+        state = self
```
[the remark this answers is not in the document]
jqmp: Yeah, I'm on the fence here. My thinking here was to keep it consistent and make it easy to add changes -- similar to using `{...}` even when you don't need to, like

    if (enabled) {
        doThing();
    }

But I think it's harder to mess this up than the `{...}` case, so I'll take another look at what the shorter way looks like.

PR #138, opened 2020-05-19
Change `attrs.names` to `attrs.dnodes`

jqmp:
This is the first step to allowing a provider to output a tuple
descriptor rather than an entity descriptor. 

Later, once we fully support tuple descriptors, each provider will
have just one output descriptor, so it will be `attrs.dnode`, not
`attrs.dnodes`.

### DIFF
--- a/bionic/flow.py
+++ b/bionic/flow.py
@@ -130,7 +130,7 @@ class FlowState(pyrs.PClass):
         return self._set_provider(provider).touch()
 
     def install_provider(self, provider):
-        for name in provider.attrs.names:
+        for name in provider.entity_names:
             if name in self.providers_by_name:
                 raise AlreadyDefinedEntityError.for_name(name)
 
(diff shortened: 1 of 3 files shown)


bionic/flow.py
```diff
@@ -577,6 +579,7 @@ def full_name(first_name, last_name):
         state = self._state
 
         names = [name for name, value in name_value_pairs]
+        print(names)
```
namanjain, line 582: remove print statement. Maybe we should add a flake8 check for print statements too.

bionic/provider.py
```diff
@@ -90,27 +90,38 @@ def get_source_func(self):
...
         return ValueProvider(names)
 
     def __init__(self, names):
+        dnodes = [entity_dnode_from_descriptor(name) for name in names]
         super(ValueProvider, self).__init__(
-            attrs=ProviderAttributes(names=names, changes_per_run=False,),
+            attrs=ProviderAttributes(dnodes=dnodes, changes_per_run=False,),
         )
 
+        self.names = names
```
namanjain, line 124: Wdyt about using ``self.attrs.entity_names` instead of storing the names here?

bionic/flow.py
jqmp: Great idea. It looks like [flake8-print](https://pypi.org/project/flake8-print/) does this -- I'll see if I can set it up. (I also want to check out [flake8-fixme](https://pypi.org/project/flake8-fixme/), which addresses another common annoyance of mine.)

bionic/provider.py
jqmp: I don't feel strongly either way -- happy to change it.

PR #139, opened 2020-05-20
Janek/extra flake8 plugins

jqmp:
I added the `flake8-print` and `flake8-fixme` plugins to catch stray `print`s and `FIXME`s in our code. I also re-enabled checking for unused imports, which had gotten turned off for some reason.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,9 +7,6 @@ ignore =
     W503  # "line break occurred before a binary operator"
     # Black handles line lengths for us (slightly less strictly than flake8).
     E501  # "line too long"
-    # TODO I think we can get rid of these two.
-    F811  # "redefinition of unused name"
-    F401  # "module imported but unused"
     # We allow TODO and XXX comments in code (but not FIXME).
     T101  # "fixme found (TODO)"
     T102  # "fixme found (XXX)"
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,7 +1,18 @@
 [flake8]
 max-line-length=88
 exclude = docs,.venv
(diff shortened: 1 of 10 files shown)


setup.cfg
```diff
@@ -1,7 +1,19 @@
 [flake8]
 max-line-length=88
 exclude = docs,.venv
-ignore = E203, E501, W503, F811, F401
+# ignore = E203, E501, W503, F811, F401, T101, T102
```
namanjain, line 4: remove this line

tests/test_flow/test_api.py
```diff
@@ -726,7 +726,7 @@ def uncomputable_value():
         builder.build().get("uncomputable_value")
     except EntityComputationError as e:
         assert isinstance(e.__cause__, ZeroDivisionError)
-        # FIXME: Parallel processing should assert the error differently.
+        # TODO: Parallel processing should assert the error differently.
```
namanjain, line 729: oh this one is done. I probably forgot to remove the `FIXME`. Can you remove this too?

setup.cfg
```diff
@@ -1,7 +1,19 @@
 [flake8]
 max-line-length=88
 exclude = docs,.venv
-ignore = E203, E501, W503, F811, F401
```
namanjain, line 4: I think `F811` used to produce false positives and why we ignored it. Not sure if the problem is fixed.

jqmp: F811 does complain when we re-define a function with the same name (which can happen when you override an entity in Bionic) but we're already using `noqa: F811` in those places, which seems reasonable to me.

PR #140, opened 2020-05-21
Refactor `results_by_name` to `results_by_dnode`

jqmp:
The point of this field is to hold multiple results for a single
task (that differ only by name). Once we have tuple descriptors, we'll
be able to remove this concept and every task will have a single result.
However, in order to implement tuple descriptors we need to support
descriptors that aren't entity names, so for now it's convenient to
organize results by descriptor.

I also made corresponding changes to `value_hashes_by_name` and
`key_for_entity_name`.

There should be no functional change.

### DIFF
--- a/bionic/task_state.py
+++ b/bionic/task_state.py
@@ -48,10 +48,10 @@ class TaskState:
         #
         # This will be present if and only if both is_complete and
         # should_persist are True.
-        self._result_value_hashes_by_name = None
+        self._result_value_hashes_by_dnode = None
 
         # This can be set by get_results_assuming_complete() or _compute().
-        self._results_by_name = None
+        self._results_by_dnode = None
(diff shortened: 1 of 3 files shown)


bionic/task_state.py
```diff
@@ -344,6 +341,7 @@ def initialize(self, bootstrap, flow_instance_uuid):
         # Lastly, set up cache accessors.
         if self.should_persist:
             if bootstrap is None:
+                # TODO TaskKey.entity_name does not exist, so this will crash.
```
namanjain, line 344: Why don't you fix this too? Looks like we use this to identify the entity in error log.

namanjain: Oh I see that it's addressed in a follow up PR #141

PR #142, opened 2020-05-21
Move manager into Executor class

namanjain:
This change moves multiprocessing Manager inside the Executor class
which now contains all the components required for parallel processing
to work.

### DIFF
--- a/bionic/executor.py
+++ b/bionic/executor.py
@@ -11,34 +11,42 @@ import sys
 import threading
 import traceback
 
+from multiprocessing.managers import SyncManager
+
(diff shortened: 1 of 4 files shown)


bionic/executor.py
```diff
@@ -61,6 +69,9 @@ def flush_logs(self):
     def submit(self, fn, *args, **kwargs):
         return self._process_pool_exec.submit(fn, *args, **kwargs)
 
+    def synchronized_set(self):
```
jqmp, line 72: nit: especially since "set" is a verb, maybe `create_synchronized_set`?

PR #143, opened 2020-05-21
Make process worker count configurable

namanjain:
Process pool worker count can now be configured using
core__parallel_processing__worker_count. By default, the value is None
which uses the maximum possible cpu available. This is also the default
behavior in loky.

### DIFF
--- a/bionic/executor.py
+++ b/bionic/executor.py
@@ -10,6 +10,7 @@ import queue
 import sys
 import threading
 import traceback
+import warnings
 
 from multiprocessing.managers import SyncManager
 
@@ -19,16 +20,30 @@ from .util import oneline, SynchronizedSet
 _executor = None
 
 
-def get_reusable_executor():
+def get_reusable_executor(worker_count):
     global _executor
 
-    # TODO: For now, executor is always the same. But once the worker
-    # count is configurable, we should try and update the loky executor
-    # and log a warn if the executor changed.
     if _executor is None:
-        _executor = Executor()
+        _executor = Executor(worker_count)
+    else:
+        prev_worker_count = _executor.worker_count
+        process_pool_resized = _executor.init_or_resize_process_pool(worker_count)
+        if process_pool_resized:
+            warning = f"""
+            The number of workers in the global Loky process pool has
+            changed from {_format_worker_count(prev_worker_count)} to
+            {_format_worker_count(worker_count)}. Because this pool is
+            global, all parallel jobs will now run with the new number
+            of workers, even in Bionic flows that were configured with
+            the old number.
+            """
+            warnings.warn(oneline(warning))
 
     return _executor
+
+
+def _format_worker_count(worker_count):
+    return worker_count if worker_count is not None else "the default value"
 
 
 class Executor:
@@ -38,7 +53,9 @@ class Executor:
     logging to work seamlessly.
     """
 
-    def __init__(self):
+    def __init__(self, worker_count):
+        self.worker_count = worker_count
+
         class MyManager(SyncManager):
             pass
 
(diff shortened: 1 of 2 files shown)


bionic/executor.py
```diff
@@ -50,18 +58,27 @@ class MyManager(SyncManager):
         self._logging_receiver = LoggingReceiver(self._logging_queue)
         self._logging_receiver.start()
 
+        self._process_pool_exec = None
+        self.update_process_pool(worker_count)
+
+    def update_process_pool(self, worker_count):
+        """
+        Updates process pool and returns whether the process pool was
```
jqmp, line 66: nit: "Updates *the* process pool"

bionic/executor.py
```diff
@@ -19,14 +20,21 @@
...
-    # count is configurable, we should try and update the loky executor
-    # and log a warn if the executor changed.
     if _executor is None:
-        _executor = Executor()
+        _executor = Executor(worker_count)
+    else:
+        process_pool_changed = _executor.update_process_pool(worker_count)
+        if process_pool_changed:
+            workers = worker_count if worker_count is not None else "the default value"
+            warning = f"""
```
jqmp, line 32: WDYT about something like
```
The number of workers in the global Loky process pool has changed from {old} to {new}.
Because this pool is global, all parallel jobs will now run with the new number of workers,
even in Bionic flows that were configured with the old number.
```
I know this requires tracking a little more state, but I think it will be easier to understand, especially considering that this message will pop up unexpectedly and the user probably won't know what a ProcessPoolExecutor is.

namanjain: Done. I am not tracking the actual worker count, but the value of the `worker_count` sent to this method. The difference between the two is, for the default value of `None`, the warning will print `the default value` instead of the worker count in the `loky` executor. I'm doing this because the `worker_count` in `loky` executor is private.

namanjain: Actually, loky isn't changing the executor when worker count changes. It just resizes the executor. I'm going to change this logic.

namanjain: Fixed

bionic/executor.py
```diff
@@ -50,18 +67,38 @@ class MyManager(SyncManager):
         self._logging_receiver = LoggingReceiver(self._logging_queue)
         self._logging_receiver.start()
 
+        self._process_pool_exec = None
+        self.init_or_resize_process_pool(worker_count)
+
+    def init_or_resize_process_pool(self, worker_count):
+        """
+        Initializes the process pool if not initialized yet and
```
jqmp, line 75: nit: this structure is a little weird -- it's not obvious that "if not initialized yet" applies to "returns True", at least until you read the next section. Probably better (and more symmetrical) to start this sentence with "If the process pool is not initialized..."

bionic/executor.py
```diff
@@ -50,18 +67,38 @@ class MyManager(SyncManager):
...
+        returns True.
+
+        If already initialized, resizes the process pool if the
+        worker_count is different than the worker count used
+        previously. When resizing, it returns whether the process
```
jqmp, line 80: "When resizing, it returns whether the process pool was resized." This makes it sound like it always return True! Maybe we can just say "... and returns a boolean indicating whether the pool was resized"?

bionic/flow.py
```diff
@@ -1716,13 +1716,16 @@ def core__persistent_cache(
         )
 
     builder.assign("core__parallel_processing__enabled", False)
+    # Uses max available CPUs when set to None.
```
jqmp, line 1719: nit: To make this a complete sentence, we could say "The executor uses ...".

PR #144, opened 2020-05-21
Fix persistence tests for parallel processing

namanjain:
For some reason, changing the tests started emitting `F811` warnings
again. I moved the `noqa` comment next to the functions to fix all
those warnings.

### DIFF
--- a/tests/test_flow/test_persistence.py
+++ b/tests/test_flow/test_persistence.py
@@ -21,8 +21,7 @@ class ReadCountingProtocol(bn.protocols.PicklableProtocol):
 # It would be nice to move the builder setup into fixtures, but since we need
 # to access the bound functions as well (to check the number of times they were
 # called), it's easiest to just have one long test.
-@pytest.mark.no_parallel
-def test_caching_and_invalidation(builder, make_counter):
+def test_caching_and_invalidation(builder, make_counter, parallel_processing_enabled):
     # Set up the builder with singleton values.
 
     builder.assign("x", 2)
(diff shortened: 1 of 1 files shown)


tests/test_flow/test_persistence.py
```diff
@@ -60,10 +59,19 @@ def xy_plus_yz(xy, yz):
...
     assert flow.get("yz") == 12
+    assert xy_counter.times_called() == 0
     assert yz_counter.times_called() == 1
 
     assert flow.get("xy_plus_yz") == 18
     assert flow.get("xy_plus_yz") == 18
+    assert xy_counter.times_called() == 0
+    if parallel_processing_enabled:
+        # This is different from serial processing because we don't pass
+        # in-memory cache to the subprocesses. The subprocess computes
```
jqmp, line 70: nit: "the in-memory cache"

tests/test_flow/test_persistence.py
```diff
@@ -175,8 +200,8 @@ def f(x, y):
 
     builder.delete("f")
 
-    @builder  # noqa: F811
-    def f(x, y):
+    @builder
+    def f(x, y):  # noqa: F811
```
jqmp, line 204: Hmm, weird. Flake8 seemed happy with the original way for me.

namanjain: When I run `flake8` from terminal (v `3.8.1`), it wants the original style too. But `pre-commit` `flake8` prefers the new style (v `3.7.9`). 

Looks like the behavior changed with `flake8` versions. I'll try updating the pre-commit flake8 version.

namanjain: Looks like the behavior changed for python version instead of flake8 version. I'm just going to add it on both lines since we support python 3.6+. 

https://gitlab.com/pycqa/flake8/-/issues/583

jqmp: Wow, that sucks! Oh well.

PR #150, opened 2020-05-22
Use tuple descriptors internally

jqmp:
This changes implementation of `@outputs` and `builder.adding_cases` to
use tuple descriptors instead of using providers that return multiple
values. This has no functional changes, with the exception that tuples
nodes now appear in the DAG visualization as their own nodes.  (The
logging for these nodes is probably also slightly different.)

One big benefit of this change is that now every task has exactly one
key and returns exactly one value, which means we can substantially
simplify a lot of code in a future refactor.

### DIFF
--- a/bionic/decorators.py
+++ b/bionic/decorators.py
@@ -18,7 +18,6 @@ from .provider import (
     AttrUpdateProvider,
     PyplotProvider,
     RenamingProvider,
-    NameSplittingProvider,
     ArgDescriptorSubstitutionProvider,
     NewOutputDescriptorProvider,
 )
@@ -254,9 +253,7 @@ def outputs(*names):
         A decorator which can be applied to an entity function.
     """
 
-    return decorator_updating_accumulator(
-        lambda acc: acc.wrap_provider(NameSplittingProvider, names)
-    )
+    return returns(",".join(names) + ",")
 
 
 def docs(*docs):
--- a/bionic/provider.py
+++ b/bionic/provider.py
@@ -28,7 +28,7 @@ from .code_hasher import CodeHasher
 from .deps.optdep import import_optional_dependency
 from .descriptors.parsing import entity_dnode_from_descriptor
 from .descriptors import ast
-from .exception import EntityComputationError, EntityValueError, IncompatibleEntityError
+from .exception import EntityComputationError, IncompatibleEntityError
 from .utils.misc import groups_dict, oneline
 
 import logging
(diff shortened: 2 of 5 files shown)


jqmp:
Actually I think there's one other behavioral change: multi-output values will be memoized even when `@memoize(False)` is set. For example:

    @builder
    @bn.memoize(False)
    @bn.outputs('train_data', 'test_data')
    def _(...):
        ...

In this case, the individual `train_data` and `test_data` values will not be memoized, but the `train_data, test_data` tuple will be -- which defeats the point of using `@memoize(False)`.

This not good, but we could fix it by adding code to discard memoized values as soon as they're no longer needed for the current computation. I think this is a good idea in general and would also make the serial processing mode more consistent with the parallel mode. Maybe it would be best to make that change before merging this PR.

namanjain:
Re `memoize` comment:
I also touched on this part in your previous PR ([comment](https://github.com/square/bionic/pull/149#discussion_r431206292)) which I think introduced this behavior (although this pronounces it by changing `@outputs` to use tuples). I agree we need to revisit our caching policy. Let's talk about this off github.

jqmp:
> Let's talk about this off github.

We discussed this offline and another good point came up: in the `train_data, test_data` example above, the code here will not persist both entities unless they're both required for the current computation. E.g., if the user does `flow.get('train_data')`, then `train_data` will be persisted but but `test_data` won't; and if the user later does `flow.get('test_data')`, both of them will be recomputed. This probably doesn't match what the user expects, so I need to think more about how to handle this.

jqmp:
I've cleaned up this PR a little, and retargeted it on top of `janek/improve-error-reporting-for-multiple-outputs`, since there would otherwise be a small collision. Now that the follow-up tasks are merged, all the tests pass.

bionic/provider.py
```diff
@@ -136,18 +136,25 @@ def from_single_value_providers(cls, value_providers):
         return ValueProvider(names)
 
     def __init__(self, names):
-        dnodes = [entity_dnode_from_descriptor(name) for name in names]
+        # TODO Move this special-casing out of this class -- perhaps into the Flow code,
```
namanjain, line 139: 👍

bionic/decorators.py
```diff
@@ -254,9 +253,7 @@ def outputs(*names):
         A decorator which can be applied to an entity function.
     """
 
-    return decorator_updating_accumulator(
-        lambda acc: acc.wrap_provider(NameSplittingProvider, names)
-    )
+    return returns(",".join(names) + ",")
```
namanjain, line 256: Now that we are done with this change, what do you think about marking `returns` decorator stable and deprecating `outputs` in favor of `returns`? Or do you want to wait till we have another descriptor that uses `returns`?

namanjain (approved):
It was a long road, but all tests pass and everything looks so clean ✨ ✨ ✨

bionic/decorators.py
jqmp: Since it doesn't really enable any new functionality compared to `@output`/`@outputs`, I'd like to leave `@returns` undocumented for now.

In fact, I'm leaning towards leaving it undocumented/experimental for a while, and instead continuing to add special-purpose decorators like `@output_as_path`, `@arg_as_path`, `@output_as_future`, which would just be wrappers around the descriptor functionality. That way we can keep iterating on the descriptor syntax, and it'll be easier to document them when they're a coherent architectural feature instead of a small collection of tricks.

PR #151, opened 2020-05-27
Remove values from CaseKey

namanjain:
CaseKeys are sent to the subprocesses as part of a lot of objects. We
don't want values to be sent across directly to subprocesses as they
can be unpicklable or user might have defined a custom protocol for
them that we should use instead.

This change stops using values inside CaseKey. It is only used to get
the values for the multi-index when fetching a series for the user.
Instead of storing values in CaseKey, we now store them inside
providers which already takes care of CaseKeys for us, are immutable,
and aren't sent to the subprocesses.

### DIFF
--- a/bionic/datatypes.py
+++ b/bionic/datatypes.py
@@ -179,23 +179,10 @@ class CaseKeySpace(ImmutableSequence):
         return f'CaseKeySpace({", ".join(repr(name) for name in self)})'
 
 
-@attr.s(frozen=True)
-class MissingCaseKeyToken:
-    def __str__(self):
-        return "<MISSING>"
-
-
 class CaseKey(ImmutableMapping):
     """
     A collection of name-token pairs that uniquely identifies a case.
     """
-
-    # This is a sentinel value used to indicate that no value is available. We can't use
-    # None because None is itself a valid token for None Value.
-    # Normally I would prefer to represent missing-ness out-of-band by making the
-    # `missing_names` field the source of truth here, but the relational methods like
-    # `project` are cleaner when we use a sentinel value.
-    MISSING = MissingCaseKeyToken()
 
     def __init__(self, name_token_pairs):
         tokens_by_name = {name: token for name, token in name_token_pairs}
@@ -205,7 +192,13 @@ class CaseKey(ImmutableMapping):
         self.tokens = tokens_by_name
         self.space = CaseKeySpace(list(tokens_by_name.keys()))
         self.missing_names = [
-            name for name, token in name_token_pairs if token == self.MISSING
+            # None is a sentinel value used to indicate that no value is available.
+            # Normally I would prefer to represent missing-ness out-of-band by making the
+            # `missing_names` field the source of truth here, but the relational methods like
+            # `project` are cleaner when we use a sentinel value.
+            name
+            for name, token in name_token_pairs
+            if token is None
         ]
         self.has_missing_values = len(self.missing_names) > 0
 
--- a/bionic/datatypes.py
+++ b/bionic/datatypes.py
@@ -180,49 +180,40 @@ class CaseKeySpace(ImmutableSequence):
 
 
 @attr.s(frozen=True)
-class MissingCaseKeyValue:
-    pass
+class MissingCaseKeyToken:
+    def __str__(self):
+        return "<MISSING>"
 
 
 class CaseKey(ImmutableMapping):
     """
-    A collection of name-value pairs that uniquely identifies a case.
+    A collection of name-token pairs that uniquely identifies a case.
     """
 
     # This is a sentinel value used to indicate that no value is available. We can't use
-    # None because None is itself a valid value.
+    # None because None is itself a valid token for None Value.
     # Normally I would prefer to represent missing-ness out-of-band by making the
     # `missing_names` field the source of truth here, but the relational methods like
     # `project` are cleaner when we use a sentinel value.
-    MISSING = MissingCaseKeyValue()
+    MISSING = MissingCaseKeyToken()
 
-    def __init__(self, name_value_token_triples):
-        values_by_name = {
-            name: value for name, value, token in name_value_token_triples
-        }
-        tokens_by_name = {
-            name: token for name, value, token in name_value_token_triples
-        }
+    def __init__(self, name_token_pairs):
+        tokens_by_name = {name: token for name, token in name_token_pairs}
 
         super(CaseKey, self).__init__(tokens_by_name)
-        self._nvt_triples = name_value_token_triples
-        self.values = values_by_name
+        self._name_token_pairs = name_token_pairs
         self.tokens = tokens_by_name
-        self.space = CaseKeySpace(list(values_by_name.keys()))
+        self.space = CaseKeySpace(list(tokens_by_name.keys()))
         self.missing_names = [
-            name
-            for name, value, token in name_value_token_triples
-            # We don't want to say `value == MISSING` because `value` might override the
-            # `==` operator to mean something different (like a Pandas DataFrame does).
-            if type(value) is MissingCaseKeyValue
+            name for name, token in name_token_pairs if token == self.MISSING
         ]
         self.has_missing_values = len(self.missing_names) > 0
 
     def project(self, key_space):
         return CaseKey(
             [
-                (name, value, token)
-                for name, value, token in self._nvt_triples
+                (name, token)
+                for name, token in self._name_token_pairs
                 if name in key_space
             ]
         )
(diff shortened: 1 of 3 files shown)


bionic/flow.py
```diff
@@ -1231,9 +1231,21 @@ def get(self, name, collection=None, fmt=None, mode=object):
...
+                    if not isinstance(provider, ValueProvider):
+                        message = f"""
+                        Provider for the entity {case_name} was expected to be
+                        a ValueProvider but wasn't.
+                        """
+                        raise AssertionError(oneline(message))
                 index = multi_index_from_case_keys(
                     case_keys=[result.query.case_key for result in result_group],
                     ordered_key_names=list(result_group.key_space),
+                    providers_by_name=providers_by_name,
```
jqmp, line 1248: The way we're pulling the values directly out of the ValueProviders is a little awkward, and it will be even more awkward when we have tuple descriptors. Instead, what if we used `self._deriver.derive()` here to get the values for each of entities in the key space, and then use those values to construct the index? Then we wouldn't need as many changes to `ValueProvider`.

namanjain: This definitely sounds neater to me but I don't really understand how to implement this. How are we going to match the case key of `name` in `Flow.get(name)` with the case keys of the result group derived for the entity names in key space? The case keys for them can be different.

bionic/flow.py
```diff
@@ -1231,9 +1231,21 @@ def get(self, name, collection=None, fmt=None, mode=object):
             return set(values)
         elif collection is pd.Series or collection == "series":
             if len(result_group.key_space) > 0:
+                providers_by_name = {
+                    case_name: self._state.get_provider(case_name)
+                    for case_name in list(result_group.key_space)
```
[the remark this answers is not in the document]
namanjain: Ah you're right, we don't. I just copied what we were sending to `multi_index_from_case_keys` since I haven't used `ImmutableSequence` before. But going to remove this.

bionic/flow.py
jqmp: Here is my thinking (but it's possible I've missed something):

Each name in the original entity's key space should be the name of an ancestor entity, and that ancestor entity's key space should be a subset of the original's. So to find the ancestor value that corresponds to the original value, we should just need to find the ancestor case key whose values match the original within that subset.

If that's still confusing, want to chat by phone?

namanjain: I'll see how hard it is to implement a is_subset / is_superset methods on `CaseKey`. Otherwise, this makes sense! Thank you!

(Rebased the branch to pickup github actions updates)

bionic/datatypes.py
```diff
@@ -194,62 +194,52 @@ class CaseKey(ImmutableMapping):
...
-                vt_pairs_by_name[name] = value, token
+                tokens_by_name[name] = token
 
-        return CaseKey(
-            [(name, value, token) for name, (value, token) in vt_pairs_by_name.items()]
-        )
+        return CaseKey([(name, token) for name, token in tokens_by_name.items()])
 
     def __repr__(self):
         args_str = ", ".join(f"{name}={token}" for name, token in self.items())
```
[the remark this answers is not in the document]
namanjain: We use tokens for returning a token in `multi_index_from_case_keys` and for finding out `MISSING` values.

bionic/flow.py
namanjain: Doh, `CaseKey` and `CaseKeySpace` already support set operations that I need (in the form of `select` and `project`). Thanks Janek for pointing out that set operations exist!

namanjain: Done!

bionic/datatypes.py
```diff
@@ -180,76 +180,63 @@ def __repr__(self):
...
 
 class CaseKey(ImmutableMapping):
     """
-    A collection of name-value pairs that uniquely identifies a case.
+    A collection of name-token pairs that uniquely identifies a case.
     """
 
     # This is a sentinel value used to indicate that no value is available. We can't use
-    # None because None is itself a valid value.
+    # None because None is itself a valid token for None Value.
```
jqmp, line 194: Wait, is it true that `None` is a valid token for the `None` value? Doesn't `tokenize(None)` return a string? If so, can we just use `None` as our sentinel value?

bionic/flow.py
```diff
@@ -1231,9 +1232,34 @@ def get(self, name, collection=None, fmt=None, mode=object):
             return set(values)
         elif collection is pd.Series or collection == "series":
             if len(result_group.key_space) > 0:
+                case_keys = [result.query.case_key for result in result_group]
+
+                resultgroups_by_entity_name = {
```
jqmp, line 1237: nit: `result_groups_by_entity_name`

Also, I would recommend prefixing all the ancestor-related values with something `ancestor_` -- that would make it clearer what's going on in a line like `case_key.project(result_case_key.space)`, where it's not obvious that we're projecting an *original* case key onto the key space of an *ancestor*.

bionic/flow.py
```diff
@@ -1231,9 +1232,34 @@ def get(self, name, collection=None, fmt=None, mode=object):
...
+                    entity_name: self._deriver.derive(
+                        entity_dnode_from_descriptor(entity_name)
+                    )
+                    for entity_name in result_group.key_space
+                }
+                values_by_entity_name_by_case_key = defaultdict(lambda: defaultdict())
```
jqmp, line 1243: Why is the inner dict a `defaultdict` instead of a regular `dict`?

But also, rather than looping through every combination of original and ancestor case key, it might be cleaner to organize the ancestor values according to the *ancestor* case key. Then you can look them up like this:

```
ancestor_values_by_case_key_by_name[ancestor_name][orig_case_key.project(ancestor_key_space)]
```

bionic/flow.py
```diff
@@ -1231,9 +1232,34 @@ def get(self, name, collection=None, fmt=None, mode=object):
...
+                            result_case_key = result.query.case_key
+                            if (
+                                case_key.project(result_case_key.space)
+                                == result_case_key
+                            ):
+                                values_by_entity_name_by_case_key[case_key][
+                                    entity_name
+                                ] = result.value
+
                 index = multi_index_from_case_keys(
```
jqmp, line 1259: At this point, the signature of this function is so complicated that I wonder if it even makes sense to keep it around rather than just inlining it.

bionic/datatypes.py
namanjain: I really thought `tokenize` returns `None` token for `None` string but you are right, I tried it out and it returns a non-empty token. Turns out `str(None)` returns `"None"` and not `None`. I'll remove `MissingCaseKeyToken`. 

This will change the behavior a little bit since now the token will be `None` instead of `"<MISSING>"`.

bionic/flow.py
namanjain: lol makes sense to inline it

bionic/flow.py
namanjain: Thanks, this was very helpful!

bionic/flow.py
```diff
@@ -1231,9 +1232,40 @@ def get(self, name, collection=None, fmt=None, mode=object):
...
+                                        ancestor_key_space_by_name[ancestor_name]
+                                    )
+                                ],
+                                token=orig_case_key.tokens[ancestor_name],
+                            )
+                            for ancestor_name in result_group.key_space
+                        )
+                        for orig_case_key in orig_case_keys
+                    ],
+                    names=list(result_group.key_space),
```
jqmp, line 1268: nit: This code assumes that when you iterate over the elements of `result_group.key_space`, you get them in the same order each time. This happens to be true because `CaseKeySpace` happens to be implemented as a `tuple`, but if it were implemented as a `set` it would not be true. Since a key space is conceptually similar to a set, we might want to pick one ordering and use it each time.

bionic/provider.py
```diff
@@ -211,18 +211,15 @@ def get_tasks(self, dep_key_spaces_by_dnode, dep_task_key_lists_by_dnode):
                 for case_key in self._value_tuples_by_case_key.keys()
             ]
 
-        # If we have no cases, we instead return a single "missing value" task.
+        # If we have no cases, we instead return a single task with no token.
```
jqmp, line 214: To me, the original comment was clearer. The important thing is that the task's case key indicates that the value is missing -- the fact that we represent this with a missing token seems more like an implementation detail.

bionic/provider.py
```diff
@@ -915,12 +912,18 @@ def __repr__(self):
         return f"HashableWrapper({self._value!r})"
 
 
-def multi_index_from_case_keys(case_keys, ordered_key_names):
+def multi_index_from_case_keys(
```
jqmp, line 915: Does anything still use this?

bionic/datatypes.py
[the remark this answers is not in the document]
jqmp: As you noted, we will now be printing `None` instead of `<MISSING>` when the token is missing. Since this is the same thing as what we print when the actual value is `None`, maybe we should continue to print `<MISSING>` here?

bionic/provider.py
namanjain: This is fair. I'll add both to be accurate and clear.

bionic/provider.py
namanjain: Nope, removing it.

PR #159, opened 2020-06-03
Add changelog file

namanjain:
Added the rst file that will contain the changelog and added it to the
docs. I'll add notes for 0.8.0 soon and we can consider adding notes
for the older releases later.

I picked the template from [keep-a-changelog](https://github.com/olivierlacan/keep-a-changelog) but happy to try a
different template if you have preference.

This is what the changelog looks like in the documentation.
Screen Shot 2020-06-03 at 2 43 06 PM

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,5 +1,6 @@
 include README.md
 include LICENSE.txt
+include CHANGELOG.md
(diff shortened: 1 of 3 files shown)


namanjain:
Declining in favor of #164

PR #160, opened 2020-06-08
Update parallel processing executor to use correct worker count

namanjain:
Bionic maintains a single instance of process pool executor that it
shares between all flow instances in the current process. If the worker
count between flows is different, that creates a problem since the
executor is shared.

Till now, Bionic used the worker count as defined by the last flow and
emitted a user warning if the worker count changed. With this change,
Bionic resizes the executor using each flows worker count before any
entity is derived. This removes the user overhead of making sure that
the worker count is correct and allows them to experiment with
different worker count as well.

### DIFF
--- a/bionic/deriver.py
+++ b/bionic/deriver.py
@@ -145,6 +145,9 @@ class EntityDeriver:
                 "core__versioning_policy"
             ),
             executor=self._bootstrap_singleton_entity("core__executor"),
+            executor_worker_count=self._bootstrap_singleton_entity(
+                "core__parallel_processing__worker_count"
+            ),
         )
 
     def _prevalidate_base_dnodes(self):
@@ -337,6 +340,9 @@ class EntityDeriver:
             self._get_or_create_task_state_for_key(task.keys[0]) for task in dinfo.tasks
         ]
 
+        if self._bootstrap is not None:
+            self._bootstrap.resize_executor()
+
         task_runner = TaskCompletionRunner(self._bootstrap, self._flow_instance_uuid)
         task_runner.run(requested_task_states)
 
@@ -364,6 +370,11 @@ class Bootstrap:
     persistent_cache = attr.ib()
     versioning_policy = attr.ib()
     executor = attr.ib()
+    executor_worker_count = attr.ib()
+
+    def resize_executor(self):
+        if self.executor is not None:
+            self.executor.init_or_resize_process_pool(self.executor_worker_count)
 
 
 class TaskKeyLogger:
--- a/bionic/deriver.py
+++ b/bionic/deriver.py
@@ -145,9 +145,6 @@ class EntityDeriver:
                 "core__versioning_policy"
             ),
             executor=self._bootstrap_singleton_entity("core__executor"),
-            executor_worker_count=self._bootstrap_singleton_entity(
-                "core__parallel_processing__worker_count"
-            ),
         )
 
     def _prevalidate_base_dnodes(self):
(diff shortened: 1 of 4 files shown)


bionic/deriver.py
```diff
@@ -337,6 +340,9 @@ def _compute_result_group_for_dnode(self, dnode):
             self._get_or_create_task_state_for_key(task.keys[0]) for task in dinfo.tasks
         ]
 
+        if self._bootstrap is not None:
+            self._bootstrap.resize_executor()
```
jqmp, line 344: WDYT about doing this inside `Executor.submit`? The need for resizing seems to me to be a detail of the way we use Loky, and this would also simplify the deriver and remove some `is not None` checks.

namanjain: That seems a lot cleaner. Thanks!

bionic/executor.py
```diff
@@ -55,58 +26,74 @@ class Executor:
...
-            return False
-
-        self.worker_count = worker_count
-
         # Loky uses cloudpickle by default. We should investigate further what would
         # it take to make is use pickle and wrap the functions that are non-picklable
         # using `loky.wrap_non_picklable_objects`.
         loky = import_optional_dependency("loky", purpose="parallel processing")
 
         self._process_pool_exec = loky.get_reusable_executor(
```
jqmp, line 45: Just to confirm: this is very cheap if the pool is not being resized, right? It might be worth stating this in a comment, because we're going to be calling this function a lot.

bionic/executor.py
```diff
@@ -55,58 +26,74 @@ class Executor:
...
-            max_workers=worker_count,
+            max_workers=self.worker_count,
             initializer=logging_initializer,
-            initargs=(self._logging_queue,),
+            initargs=(self._manager.logging_queue,),
         )
         return True
```
jqmp, line 50: Do we still need to return a value?

bionic/executor.py
```diff
@@ -55,58 +26,74 @@ class Executor:
...
-        self._logging_receiver = LoggingReceiver(self._logging_queue)
-        self._logging_receiver.start()
-
+        self._manager = get_singleton_manager()
         self._process_pool_exec = None
-        self.init_or_resize_process_pool(worker_count)
+        self.init_or_resize_process_pool()
 
-    def init_or_resize_process_pool(self, worker_count):
+    def init_or_resize_process_pool(self):
```
jqmp, line 33: Should this be private now?

tests/test_flow/test_executor.py
```diff
@@ -0,0 +1,53 @@
...
+
+from ..conftest import ExecutionMode
+
+from bionic.executor import get_singleton_manager, logging_initializer
+from bionic.optdep import import_optional_dependency
+
+
+# TODO use a marker to run parallel execution mode tests.
+@pytest.fixture(params=[ExecutionMode.PARALLEL])
+def parallel_processing_enabled(request):
```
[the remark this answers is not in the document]
namanjain: Yes, this is overriding `parallel_processing_enabled` in conftest. Added a comment.

tests/test_flow/test_executor.py
```diff
@@ -0,0 +1,53 @@
...
+        return x
+
+    builder.set("core__parallel_processing__worker_count", 2)
+    flow1 = builder.build()
+
+    builder.set("core__parallel_processing__worker_count", 3)
+    flow2 = builder.build()
+
+    assert flow1.get("y") == 1
+    # It's gross to check a private variable of the executor but this is
```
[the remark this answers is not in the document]
namanjain: Right, but we will have to synchronize the tasks to not finish till all the tasks are complete. I can use a lock (or some other locking interface) from manager to do it. But I'll leave a TODO for now.

bionic/executor.py
namanjain: Right, this is very cheap. Loky compares the original and new arguments to `get_reusable_executor` and simply returns the executor.

PR #163, opened 2020-06-11
Add geopandas protocol to default protocols

namanjain:
Added geopandas to the list of default protocols. Users won't need to
decorate geopandas entity from now on.

Here's how the documentation changes look like:

Screen Shot 2020-06-12 at 11 10 29 AM

Protocols:
Screen Shot 2020-06-11 at 11 05 58 AM

### DIFF
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -393,7 +393,7 @@ return different results whenever the database's contents change.  In
 cases like this, it's not appropriate to reuse the function's previous cached
 values; we want Bionic to recompute the value each time.
 
-You can tell Bionic that a function is non-deterministic by applying the 
+You can tell Bionic that a function is non-deterministic by applying the
 :meth:`@changes_per_run <bionic.changes_per_run>` decorator:
 
 .. code-block:: python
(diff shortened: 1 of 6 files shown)


docs/concepts.rst
```diff
@@ -541,10 +541,12 @@ object types it's more efficient or more idiomatic to use a different format.
 There are also some types of objects that can't be pickled at all.  Bionic uses
 ``pickle`` by default, but handles some types specially: `Pandas
 <https://pandas.pydata.org/>`_ DataFrames are serialized in the `Parquet
-<https://parquet.apache.org/>`_ format, while `Pillow
-<https://pillow.readthedocs.io/en/stable/>`_ Images are serialized as `PNG
-<https://en.wikipedia.org/wiki/Portable_Network_Graphics>`_\ s.  You can
-explictly specify a serialization strategy for an entity by attaching a
+<https://parquet.apache.org/>`_ format, `Geopandas <https://geopandas.org/>`_
```
jqmp, line 544: I guess we should add the other formats we automatically handle as well: NumPy and Dask.

Also, it looks like it's "GeoPandas", not "Geopandas".

setup.py
```diff
@@ -22,6 +22,7 @@
     "pandas",
     "pyarrow",
     "pyrsistent",
+    "geopandas",
```
jqmp, line 25: I don't think this needs to be a required dependency. Let's handle this the way we handle Dask, etc. (That means we'll want to add an entry to the "Extra Packages" section of the "Get Started" docs.)

jqmp:
Oh, also: we should probably add a new test to capture the new behavior! (I'm not sure we've been doing this with the other default protocols, but we probably should have.)

setup.py
namanjain: Ah I see how protocols really work. The key is `BaseProtocol.value_is_valid()` method where it catches the exception if the optional dep is not satisfied. 

What you suggested makes sense, I'll change it accordingly.

namanjain:
> Oh, also: we should probably add a new test to capture the new behavior! (I'm not sure we've been doing this with the other default protocols, but we probably should have.)

Hah, this caught a potential bug with geopandas protocol in default protocol list. Turns out, geopandas df inherits pandas df. So it should be before other df protocols.

docs/concepts.rst
```diff
@@ -539,13 +539,22 @@ Most Python objects can be serialized with Python's built-in `pickle
...
-``pickle`` by default, but handles some types specially: `Pandas
-<https://pandas.pydata.org/>`_ DataFrames are serialized in the `Parquet
-<https://parquet.apache.org/>`_ format, while `Pillow
-<https://pillow.readthedocs.io/en/stable/>`_ Images are serialized as `PNG
-<https://en.wikipedia.org/wiki/Portable_Network_Graphics>`_\ s.  You can
-explictly specify a serialization strategy for an entity by attaching a
-`Protocol`_ to its definition.
+``pickle`` by default, but handles some types specially:
+
+- `Dask <https://docs.dask.org/>`_ Dataframes are serialized in the `Parquet
```
jqmp, line 544: nit: It seems a little off that more common formats like Pandas come towards the end here. Maybe we could order them by roughly descending popularity?

My guess:
* Pandas
* NumPy
* Pillow
* Dask
* GeoPandas

tests/test_flow/test_protocols.py
```diff
@@ -31,12 +31,20 @@
 ]
 
 
+# A passthrough decorator that can be used in plae of protocol
```
jqmp, line 34: s/plae/place/
s/of protocol decorators/of a protocol decorator/

docs/concepts.rst
```diff
@@ -539,13 +539,22 @@ Most Python objects can be serialized with Python's built-in `pickle
...
+  <https://parquet.apache.org/>`_ format,
+- `GeoPandas <https://geopandas.org/>`_ Dataframes are serialized as `SHP
+  <https://en.wikipedia.org/wiki/Shapefile#Shapefile_shape_format_(.shp)>`_\ s,
+- `NumPy <https://numpy.org/>`_ Arrays are serialized in the `NPY
+  <https://numpy.org/devdocs/reference/generated/numpy.lib.format.html#npy-format>`_
+  format,
+- `Pandas <https://pandas.pydata.org/>`_ DataFrames are serialized in the
+  `Parquet <https://parquet.apache.org/>`_ format,
+- `Pillow <https://pillow.readthedocs.io/en/stable/>`_ Images are serialized as
```
jqmp, line 553: We might as well use the same sentence structure for each of these bullet points. Maybe "$LIBRARY $OBJECTs are serialized using $FORMAT."?

(Also, let's put a period at the end of each bullet point.)

PR #164, opened 2020-06-11
Add Release Notes to documentation

jqmp:
This new page tracks user-facing changes in each of our releases. It also describes our versioning scheme.

This builds on https://github.com/square/bionic/pull/159, but with a different format and all the historical changes.


jqmp:
The new page looks like this:

image

namanjain:
New changes look good to me

PR #166, opened 2020-06-15
Make ProcessManager global

namanjain:
ProcessManager was defined inside ExternalProcessLoggingManager.__init__
which makes ProcessManager class unpickable. We noticed this problem
when running parallel computing tests on MacOS with python version 3.8.
In python 3.8, multiprocessing uses spawn instead of fork as the default
mode which requires everything passed to be pickable.

This change makes ProcessManager a global class so it can be pickled and
parallel computing works for MacOS on python 3.8.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -35,13 +35,22 @@ def builder(parallel_processing_enabled, tmp_path):
     return builder
 
 
+# This is a different multiprocessing manager than the one we use in
+# ExternalProcessLoggingManager. This one is responsible for sharing test
+# objects between processes. Its' only used in make_counter and make_list
+# fixtures. The purpose here is to not pollute the manager that bionic uses.
+# I also don't want to replace the manager that bionic creates with a test
+# one.
+class PytestManager(SyncManager):
+    pass
+
+
+PytestManager.register("ResettingCounter", ResettingCounter)
+
+
 @pytest.fixture(scope="session")
 def multiprocessing_manager(request):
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -35,13 +35,16 @@ def builder(parallel_processing_enabled, tmp_path):
     return builder
 
 
+class ProcessManager(SyncManager):
+    pass
+
+
+ProcessManager.register("ResettingCounter", ResettingCounter)
```
jqmp, line 42: Obviously I'm missing something, but why don't we need to register `SynchronizedSet` here?

tests/conftest.py
```diff
@@ -35,13 +35,16 @@ def builder(parallel_processing_enabled, tmp_path):
     return builder
 
 
+class ProcessManager(SyncManager):
```
jqmp, line 38: nit: maybe this should have a different name from the other `ProcessManager`?

tests/conftest.py
namanjain: This is a different multiprocessing manager than the one we use in `ExternalProcessLoggingManager`. This one is responsible for sharing test objects between processes. Its' only used in `make_counter` and `make_list` fixtures. 

The purpose here is to not pollute the manager that bionic uses. I also don't want to replace the manager that bionic creates with a test one.  

I guess your other comment in valid that we should rename this to avoid confusion.

PR #168, opened 2020-06-15
Write user defined values to a temporary cache

namanjain:
In parallel computing mode, Bionic currently passes all the user
defined values to the sub processes using IPC. This can break if the
values are not pickable which user can get around by writing a custom
protocol.

This change writes all the user provided values to a temporary local
cache that can be read by the subprocesses. This allows us to take
advantage of protocol serialization and not worry about breaking our
user contract.

The temporary values have a life span of a get call. As soon as the
entities are derived in `Deriver._compute_result_group_for_dnode`, we
reset the tmp cache.

As a fail-safe, we also attempt to reset the tmp cache before deriving
entities. This protects against cases where the execution failed while
deriving values and leaves the cache and more importantly, task states
in a bad state.

Now that we are storing user provided values in a temporary cache, the
provenance for the values in other provenances with a dependency on it
changes. Instead of using the whole provenance of values, it now uses
the value hash instead (unless user specifically turns off persistence
for them).


namanjain:
Closing this in favor of #170

PR #169, opened 2020-06-16
Add documentation for parallel execution

jqmp:
DO NOT MERGE until parallel execution is ready for release.

### DIFF
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -545,6 +545,8 @@ you must have GCP credentials configured.  You should also use ``pip install
 
 .. _gsutil: https://cloud.google.com/storage/docs/gsutil
 
+.. _protocols :
+
 Serialization Protocols
 .......................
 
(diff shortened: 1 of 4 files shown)


docs/concepts.rst
```diff
@@ -735,6 +737,55 @@ values by "case" instead of by entity.
 Other Features
 --------------
 
+Parallel Execution
```
namanjain, line 740: How do you feel about adding `.. versionadded:: 0.8.0` for parallel execution?

namanjain, line 740: Can you also add an entry for this in release notes?

namanjain: Oh, now that we release notes, I wonder if we even need any of these `versionadded`. Users can check the release notes for when the feature or concept was released.

jqmp: This still sounds like a good idea to me, although I could go either way.

Issue #115, opened 2020-04-24
Redefining an entity with a function can produce surprising results


jqmp:
When an already-existing entity is defined with a function, its previous declaration (including any protocol, docstring, etc.) is erased. This is intended behavior, but it can be surprising -- especially since it's only mentioned [in passing](https://bionic.readthedocs.io/en/stable/concepts.html#declaring-setting-assigning-and-deriving-entities) in the docs. (It's also mentioned in the docstring of `FlowBuilder.__call__`, but that doesn't show up anywhere in our documentation.)

Here's a minimal test that a new user might expect to pass, but doesn't:

```
def test_declared_protocol_is_kept(builder):
    protocol = bn.protocol.dillable()
    builder.declare("n", protocol=protocol)

    @builder
    def n():
        return 0

    assert builder.build().entity_protocol("n") == protocol
```

There are a few ways we could improve this:

1. Document this behavior more clearly.
2. Have different decorator APIs for defining new entities vs redefining existing ones, like we do with `assign` and `set`. This would hopefully make it more obvious that the new definition completely erases the previous one.
3. Change the semantics from "overwrite previous entity and create new one" to "change the value definition of an existing entity, but keep the previous protocol, docstring, etc."

PR #170, opened 2020-06-17
Write user defined values to persistent cache

namanjain:
In parallel computing mode, Bionic currently passes all the user
defined values to the sub processes using IPC. This can break if the
values are not pickable which user can get around by writing a custom
protocol.

This change writes all the user provided values to persistence
cache that can be read by the subprocesses. This allows us to take
advantage of protocol serialization and not worry about breaking our
user contract.

Now that we are storing user provided values in a temporary cache, the
provenance for the values in other provenances with a dependency on it
changes. Instead of using the whole provenance of values, it now uses
the value hash instead (unless user specifically turns off persistence
for them).

Instead of throwing an exception that users can't use persistable
values as a dependent for internal entities, we now allow this by
not persisting those user provided values and raising a warning.

This keeps Bionic backwards compatible. User provided values are
persisted by default but they didn't before. If we disallow this
behavior, this will be a breaking change as users might be using their
own entities to generate internal entities.


bionic/cache.py
```diff
@@ -888,7 +888,7 @@ def valid_filename_from_query(query):
     return query.dnode.to_descriptor().replace(" ", "-")
 
 
-CACHE_SCHEMA_VERSION = 6
+CACHE_SCHEMA_VERSION = 7
```
jqmp, line 891: Now that we have release notes, we can include this compatibility change in them.

bionic/datatypes.py
```diff
@@ -19,8 +24,15 @@ class EntityDefinition:
...
     doc = attr.ib()
-    can_persist = attr.ib(default=True)
     can_memoize = attr.ib(default=True)
+    can_persist = attr.ib()
+
+    # NB: Order of the attributes matter when using `self` in default values.
+    # Please read the warning in documentation for more detail.
+    # https://www.attrs.org/en/stable/init.html#defaults
+    @can_persist.default
+    def _internal_entities_dont_persist(self):
```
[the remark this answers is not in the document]
namanjain: I can move this logic to `builder.declare` / `builder.assign` but we still want the default persistence to be a bit magical. If we set default True or False, then users will have to keep specifying `persist=False` for internal values or `persist=True` for user provided values.

bionic/task_state.py
```diff
@@ -22,30 +26,23 @@ def __init__(self, task, dep_states, case_key, provider, entity_defs_by_dnode):
...
-        )
-
-        # Cached values.
-        self.task_keys = task.keys
-        self.should_memoize = can_memoize
-        self.should_persist = (
-            can_persist
-            and not self.task.is_simple_lookup
-            and not self.output_would_be_missing()
+        self.should_memoize = single_unique_element(
```
jqmp, line 36: Since this is now set at a different point from `should_persist`, I wonder if it should have a different name? It has basically the same semantics as `EntityDefinition.can_persist`, so maybe it could just be called `can_persist`?

bionic/task_state.py
```diff
@@ -339,21 +336,33 @@ def initialize(self, bootstrap, flow_instance_uuid):
...
+            )
+            and not self.output_would_be_missing()
+        )
+        if can_persist and bootstrap is None:
+            descriptors = [
+                task_key.dnode.to_descriptor() for task_key in self.task_keys
+            ]
+            message = f"""
+            Descriptors {descriptors!r} are set to be persisted but that's
+            impossible; the cache is not available yet because core
```
jqmp, line 352: English nit: this semicolon probably makes more sense as a colon, because the second part of the sentence explains the first.

bionic/task_state.py
```diff
@@ -339,21 +336,33 @@ def initialize(self, bootstrap, flow_instance_uuid):
...
+            bootstrap entities depend on it.
+            We will not be saving or reloading these decorators which may
```
jqmp, line 354: "The corresponding values will not be serialized and deserialized, which may cause the values to be subtly different."

Note:
- We don't normally refer to Bionic as "We"; we can just use a passive description ("will not be serialized").
- It's not the decorators being saved, it's the values!
- I used "serialized/deserialized" instead of "saved/reloaded" to make it slightly clearer where the difference is coming from.

bionic/flow.py
```diff
@@ -927,9 +948,11 @@ def __call__(self, func):
             {len(docs)} docs {tuple(docs)!r}"""
             raise ValueError(oneline(message))
 
+        internal_names = [name for name in names if entity_is_internal(name)]
         can_persist = acc.can_persist
         if can_persist is None:
-            can_persist = True
+            can_persist = len(internal_names) == 0
```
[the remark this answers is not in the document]
namanjain: Someone can technically redefine internal entities using a function. Something like:
```
    @builder
    def core__persistent_cache__global_dir():
        return "mycachedir"
```

I don't see this happening in practice but it's better to take care of it now. 

Maybe the right answer here is that we stop redefining entities (or wiping current state) when setting entities using a function. That's one of the proposals in issue #115.

bionic/task_state.py
namanjain: do you mean `can_memoize`?

namanjain: Assuming that's what you meant, I'm fine either way. I like that `should_memoize` is similar to `should_persist` and we can later change the logic for it without worrying about naming.

But I'm fine keeping it `can_memoize` for now and worry about that later. Basically following our philosophy of **err on the side of adopting feedback** (from go/trustme).

jqmp: Sorry, yes, I did mean `can_memoize`, not `can_persist`!

I also like that we have a separate layer for both `memoize` and `persist`, but since this PR makes them less analogous, it seems less attractive for them to have analogous names. (Unless we want to move `should_memoize` to be set during `initialize()` as well?)

bionic/flow.py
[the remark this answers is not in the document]
jqmp: They can, but it will throw an exception and tell them to use `@immediate`, right? Isn't that the current behavior as well (but with the exception coming from a different place)? Since this is a fairly rare case and the ramifications may not be obvious to the user, it feels better to me to give them a detailed warning and force them to be explicit.

BTW, I'm in favor of not redefining entities in this case -- we just need to figure out what our backwards compatibility story is.

bionic/task_state.py
```diff
@@ -339,21 +336,33 @@ def initialize(self, bootstrap, flow_instance_uuid):
...
+            cause subtle changes in its value.
+            To avoid this message, disable persistence for the decorators
+            by either using `@persist(False)` or `@immediate` to your
+            corresponding function or in case of a value, pass `persist=False`
```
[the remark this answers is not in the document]
jqmp: Oops, this TODO was for myself, to remind me to leave a more detailed comment. I already did that, so you can disregard this one.

bionic/flow.py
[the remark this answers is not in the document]
namanjain: Yes but we will now emit a warning instead of throwing an exception. I can get behind the reason that the warning is better than this special logic though, as you said, it forces them to be explicit.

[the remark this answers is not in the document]
namanjain: Oh wait, I don't know how I missed this but you are right, we are validating in the entity defn and this will throw an error. I'm going to change this like you suggested.

bionic/flow.py
```diff
@@ -99,6 +105,13 @@ def define_entity(self, entity_def):
         if entity_def.name in self.entity_defs_by_name:
             raise AlreadyDefinedEntityError(entity_def.name)
 
+        if entity_is_internal(entity_def.name) and entity_def.can_persist:
+            message = f"""
+            Attempted to set @persist to True for Bionic internal
```
jqmp, line 110: Can we provide more clarity on what the user needs to do to fix this, similar to the warning we have below?

bionic/task_state.py
```diff
@@ -339,21 +336,39 @@ def initialize(self, bootstrap, flow_instance_uuid):
...
+            )
+            and not self.output_would_be_missing()
+        )
+        if should_persist and bootstrap is None:
+            descriptors = [
+                task_key.dnode.to_descriptor() for task_key in self.task_keys
+            ]
+            if self.task.is_simple_lookup:
+                disable_message = """
+                applying `@persist(False)` or `@immediate` to your corresponding
```
jqmp, line 352: nit: s/your/the/

bionic/task_state.py
```diff
@@ -339,21 +336,39 @@ def initialize(self, bootstrap, flow_instance_uuid):
...
+                function"""
+            else:
+                disable_message = """
+                passing `persist=False` argument when you `declare` / `assign` your
```
jqmp, line 356: nit: "passing `persist=False` argument" is not quite grammatically right, since it's missing a determiner. I think we can just omit "argument".

Also, I would use "the entity value" instead of "your values".

bionic/task_state.py
```diff
@@ -339,21 +336,39 @@ def initialize(self, bootstrap, flow_instance_uuid):
...
+                values"""
+            message = f"""
+            Descriptors {descriptors!r} are set to be persisted but that's
+            impossible: the cache is not available yet because core
```
jqmp, line 360: Missed a nit last time: When we say "depend on it", it sounds like "it" refers to the cache, but it really refers to the descriptors. Maybe just say "but they can't be because core bootstrap entities depend on them"?

PR #176, opened 2020-06-25
Rename cache.py to persistence.py

jqmp:
This will make it less confusing when we introduce a new user-facing
Cache class.


bionic/flow.py
```diff
@@ -17,7 +17,7 @@
 
 # A bit annoying that we have to rename this when we import it.
 from . import protocols as protos
-from .cache import LocalStore, GcsCloudStore, PersistentCache
+from .persistence import LocalStore, GcsCloudStore, PersistentCache
```
namanjain, line 20: super nit: reorder imports. I think we talked about adding a hook for this, not sure where we landed with it.

jqmp: Thanks, fixed!

PR #177, opened 2020-06-25
Reuse flows in fuzz tests

namanjain:
Fuzz tests were creating a new flow for every get operation which was
costly. This change keeps a flow object and updates it any time an
entity is added or updated. This way, we reduce the number of entities
to be completed, which is specially useful for parallel execution as
each completion takes place in a subprocess. We end up spending a lot
of time in IPC for basically just loading cached entry in cache
accessor from disk cache otherwise.

This change reduced the time taken by parallel fuzz tests by 33% on my
machine on python 3.8.


namanjain:
Added #178 to track the performance behavior issue Janek mentioned.

PR #179, opened 2020-06-25
Compute tasks states in parallel

namanjain:
Bionic completes task states in parallel when parallel execution is
turned on. But task states that are already computed and cached on
disk might not be complete. For such cases, we end up completing those
task states in the subprocesses which barely does anything and adds IPC
cost to the execution.

Bionic would now compute the task states in the subprocesses. Computing
is what we essentially want to parallelize anyway. This helps us save
IPC cost for the cases described above.

As a bonus, our total time for parallel fuzz tests also went down
by around 15-20% again.

### DIFF
--- a/bionic/task_state.py
+++ b/bionic/task_state.py
@@ -44,13 +44,13 @@ class TaskState:
         self._cache_accessors = None
         self.should_persist = None
 
-        # This can be set by complete() or _compute().
+        # This can be set by compute() or attempt_to_complete_from_cache().
         #
         # This will be present if and only if both is_complete and
         # should_persist are True.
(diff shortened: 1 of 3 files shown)


bionic/task_state.py
```diff
@@ -158,13 +130,37 @@ def get_results_assuming_complete(self, task_key_logger):
 
         return results_by_dnode
 
-    def _compute(self, task_key_logger):
+    def attempt_to_complete_from_cache(self):
+        """
+        If the results are available in persistent cache, populates value hashes
```
jqmp, line 135: Maybe we should add "Otherwise does nothing." to the end to be explicit?

PR #182, opened 2020-07-01
Address default entity names TODO in clear providers method

namanjain:


bionic/flow.py
```diff
@@ -255,24 +255,14 @@ def add_case(self, case_key, names, values):
...
-
         # Delete the providers (or fail if not possible).
         state = state.delete_providers(names)
-        state = state.delete_entity_defs(names)
 
         # Recreate an empty version of each provider.
-        for entity_def in original_entity_defs:
-            state = state.define_entity(entity_def)
-            state = state.create_provider(entity_def.name)
+        entity_names = [name for name in names if name in state.entity_defs_by_name]
```
jqmp, line 262: nit: why not just put an if statement in the for loop?

namanjain: I thought of this too but I couldn't find a nice way to represent for and if conditional together in python. I don't think something like this is allowed:
```
for name in names if name in state.entity_defs_by_name:
  pass
```

namanjain:
> This change seems fine. But now I wonder: why don't we remove entities from `default_entity_names` in `_erase_provider` instead of in `_erase_entity_def`? It seems like default-ness should be associated with the provider, not with the entity definition -- if we change the provider, then the entity is not default anymore. WDYT?

This makes sense. We always recreate the provider when an entity is changed or remove the provider when it is removed. Either way, we always erase the provider.

bionic/flow.py
jqmp: What I meant was:
```
for name in names:
    if name in state.entity_defs_by_name:
        ...
```

PR #184, opened 2020-07-02
Janek/basic cache interface

jqmp:
This has three commits: two minor refactorings, and then one large one adding a basic user-facing cache API. This API only supports listing cache entries: I'll handle deletion in a separate PR.


jqmp:
BTW, I only implemented the most minimal version of what we discussed, in order to get it out as quickly as possible. We can implement improvements as they're requested -- or we can encourage customers to submit PRs themselves!

bionic/cache_api.py
```diff
@@ -0,0 +1,108 @@
...
+        objects, one for each artifact in Bionic's persistent cache.
+        """
+
+        # These private accesses are a bit gross, but I'm not sure it's worth adding
+        # more layers of APIs to avoid them.
+        deriver = self._flow._deriver
+        deriver.get_ready()
+        persistent_cache = deriver._bootstrap.persistent_cache
+
+        stores = [
```
namanjain, line 35: nit: you don't need this tmp variable

bionic/cache_api.py
```diff
@@ -0,0 +1,108 @@
...
+        ]
+
+        return (
+            CacheEntry(cache=self, inv_item=inv_item,)
+            for store in stores
+            for inv_item in store.inventory.list_items()
+        )
+
+
+def path_from_url_if_file(url):
```
namanjain, line 48: Not in this PR, but we should move all file related helpers together in a separate util.

bionic/cache_api.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def __hash__(self):
+        return hash(self._comparison_key)
+
+    def __eq__(self, other):
+        if not isinstance(other, CacheEntry):
+            return False
+        return self._comparison_key == other._comparison_key
+
+    def __lt__(self, other):
```
namanjain, line 107: Do you not need to check that `other` is an instance of `CacheEntry` in `lt`?

bionic/persistence.py
```diff
@@ -412,6 +412,20 @@ class MetadataMatch:
     level = attr.ib()
 
 
+# TODO Should we merge this with InventoryEntry?
```
namanjain, line 415: This makes sense to me. Maybe phrase it as an action item instead of a question.

bionic/persistence.py
```diff
@@ -501,6 +515,21 @@ def find_entry(self, query):
             value_hash=metadata_record.value_hash,
         )
 
+    def list_items(self):
+        metadata_urls = [
+            url for url in self._fs.search(self.root_url) if url.endswith(".yaml")
```
namanjain, line 520: Not in this PR but we should make filesystem search relative to the `root_url`. Then we can pass it `.` instead.

tests/test_cache_api.py
```diff
@@ -0,0 +1,175 @@
...
+            artifact_path_bytes = entry.artifact_path.read_bytes()
+            assert artifact_path_bytes == artifact_bytes
+        else:
+            assert entry.artifact_path is None
+
+        # We won't make too many assumptions about the format of the metadata, but we
+        # can check that it contains the entity name. (Unfortunately it won't
+        # necessarily contain the absolute artifact URL; it may be a relative URL
+        # instead.)
+        metadata_str = read_bytes_from_url(entry.metadata_url).decode("utf-8")
```
namanjain, line 67: It does not have to be in this PR, but `CacheEntry` is a good candidate for unit tests. We can test properties easily using unit tests that would be complicated to test here.

bionic/cache_api.py
```diff
@@ -0,0 +1,108 @@
...
+    Accessible as an attribute named ``cache`` on a :class:`Flow <bionic.Flow>`
+    object. Use ``get_entries`` to iterate through the set of cache entries.
+    """
+
+    def __init__(self, flow):
+        self._flow = flow
+
+    def get_entries(self):
+        """
+        Returns a sequence of :class:`CacheEntry <bionic.cache_api.CacheEntry>`
```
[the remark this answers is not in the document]
jqmp: Oh, good idea.

bionic/cache_api.py
```diff
@@ -0,0 +1,108 @@
...
+
+    Has the following fields:
+
+    - ``tier``: "local" or "cloud", depending on which tier of the cache the artifact
+      is in.
+    - ``entity``: the name of the cached entity, or ``None`` if the artifact is does
+      not correspond to an entity
+    - ``artifact_url``: a URL to the cached artifact file or blob
+    - ``metadata_url``: a URL to the metadata file or blob describing the artifact
+    - ``artifact_path``: a Path object locating the artifact file, if it's a file
```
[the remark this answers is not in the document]
jqmp: Yes, you're right, this could be clearer.

bionic/cache_api.py
```diff
@@ -0,0 +1,108 @@
...
+class Cache:
+    """
+    A programmatic interface to Bionic's persistent cache.
+
+    Accessible as an attribute named ``cache`` on a :class:`Flow <bionic.Flow>`
+    object. Use ``get_entries`` to iterate through the set of cache entries.
+    """
+
+    def __init__(self, flow):
+        self._flow = flow
```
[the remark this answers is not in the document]
jqmp: Good point. I was originally planning to depend on the flow for other stuff, like checking whether a cache entry is used by the current flow. But for the current implementation we don't need this.

bionic/cache_api.py
jqmp: Agreed, I was thinking the same thing.

bionic/cache_api.py
jqmp: It's true we don't strictly need it, but if we inline it, the final comprehension has three nested levels and actually takes more total vertical space, so I like having it here.

That said, maybe it makes sense to add a `get_stores()` method to `PersistentCache`?

bionic/cache_api.py
jqmp: I think we have to throw an exception either way, but yeah, we can make that exception a `TypeError`.

bionic/persistence.py
jqmp: I could go either way, since we may want to add a more expensive field to one but not the other, and it's easier to merge them than un-merge them.

bionic/persistence.py
jqmp: Yep, that would work too. (I guess we'd just have to do the URL joining ourselves in the GCS filesystem.)

tests/test_cache_api.py
jqmp: What properties did you have in mind here? The things like the comparison operators?

(I'm a little skeptical of adding unit tests for this class because its interface isn't completely stable or user-visible: its properties and methods are stable, but its constructor is coupled to Bionic's internals and will probably keep changing. For example, if we do decide to merge `ExternalCacheItem` with `InventoryEntry`, we'll have to modify these unit tests.)

bionic/cache_api.py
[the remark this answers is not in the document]
jqmp: Hmm, fixing this will require rearranging the way this is initialized in Flow, because we want to call `deriver.get_ready()` at the last minute rather than in the constructor of `Flow` (because we don't want to do all the deriver setup multiple times when we do `flow.setting(...).setting(...)...`). Since I expect us to use the flow itself in the future, I'll just leave a comment here explaining that.

PR #185, opened 2020-07-02
Janek/cache delete interface

jqmp:
This has four commits:
1. Non-functional refactoring of the cache accessors.
2. Support for gracefully handling deleted entries.
3. Adding deletion to the external cache API.
4. Documentation for the external cache API.

### DIFF
--- a/bionic/deriver.py
+++ b/bionic/deriver.py
@@ -681,7 +681,7 @@ class TaskCompletionRunner:
         if task_key not in self._entries_by_task_key:
             # Before doing anything with this task state, we should make sure its
             # cache state is up to date.
-            state.refresh_all_cache_state(self._bootstrap)
+            state.refresh_all_persistent_cache_state(self._bootstrap)
             self._entries_by_task_key[task_key] = TaskRunnerEntry(
                 state=state, is_requested=is_requested
             )
--- a/bionic/deriver.py
+++ b/bionic/deriver.py
@@ -594,6 +594,14 @@ class TaskCompletionRunner:
             while self._has_pending_entries():
                 entry = self._activate_next_pending_entry()
 
+                # Before we decide if the current entry is blocked, we need to create
+                # an entry for each of its dependencies. This has the side effect of
+                # refreshing each of their cache states, so if any of them have had
+                # their cache entries deleted, their status will be correctly updated
+                # to "not complete".
+                for dep_state in entry.state.dep_states:
+                    self._get_or_create_entry_for_state(dep_state)
+
                 if len(entry.state.blocking_dep_states()) > 0:
                     self._mark_entry_blocked(entry)
                     continue
@@ -671,6 +679,9 @@ class TaskCompletionRunner:
     def _get_or_create_entry_for_state(self, state, is_requested=False):
         task_key = state.task.keys[0]
         if task_key not in self._entries_by_task_key:
+            # Before doing anything with this task state, we should make sure its
+            # cache state is up to date.
+            state.refresh_all_cache_state(self._bootstrap)
             self._entries_by_task_key[task_key] = TaskRunnerEntry(
                 state=state, is_requested=is_requested
             )
--- a/bionic/persistence.py
+++ b/bionic/persistence.py
@@ -530,9 +530,8 @@ class Inventory:
                 descriptor=metadata_record.descriptor,
             )
 
-    def delete_entry(self, metadata_url, artifact_url):
-        self._fs.delete(metadata_url)
-        self._fs.delete(artifact_url)
+    def delete_url(self, url):
+        return self._fs.delete(url)
 
     def _find_best_match(self, query):
         equivalent_url_prefix = self._equivalent_metadata_url_prefix_for_query(query)
@@ -771,7 +770,10 @@ class LocalFilesystem:
 
     def delete(self, url):
         path = path_from_url(url)
+        if not path.exists():
+            return False
         path.unlink()
+        return True
 
     def write_bytes(self, content_bytes, url):
         path = path_from_url(url)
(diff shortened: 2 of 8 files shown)


bionic/task_state.py
```diff
@@ -142,13 +145,33 @@ def attempt_to_complete_from_cache(self):
...
+        # If we haven't loaded anything from the cache, we can stop here.
+        if self._result_value_hashes_by_dnode is None:
+            return
+
+        # Otherwise, let's update our value hashes from the cache.
+        if all(axr.can_load() for axr in self._cache_accessors):
+            self._load_value_hashes()
+        else:
+            self._result_value_hashes_by_dnode = None
+            self.is_complete = False
```
namanjain, line 174: You don't need to do this, we don't have `TaskState.is_complete` anymore.

bionic/task_state.py
```diff
@@ -142,13 +145,33 @@ def attempt_to_complete_from_cache(self):
...
-        # We only load the hashed result while completing task state
-        # and lazily load the entire result when needed later.
-        value_hashes_by_dnode = {}
-        for accessor in self._cache_accessors:
-            value_hash = accessor.load_result_value_hash()
-            value_hashes_by_dnode[accessor.query.dnode] = value_hash
-        self._result_value_hashes_by_dnode = value_hashes_by_dnode
+        self._load_value_hashes()
+
+    def refresh_all_cache_state(self, bootstrap):
```
namanjain, line 150: This makes it sound like you will refresh in-memory cache too. Maybe call this `refresh_persistent_cache`?

bionic/persistence.py
```diff
@@ -765,6 +769,10 @@ def search(self, url_prefix):
             for sub_path in path_prefix.glob("**/*")
         ]
 
+    def delete(self, url):
+        path = path_from_url(url)
+        path.unlink()
```
namanjain, line 774: This throws an error if the path doesn't exist. It would be nice if this is more robust (or maybe the caller `delete_entry`) and either catches that exception and warns or passes `missing_ok=True` so it doesn't throw an exception.

bionic/persistence.py
```diff
@@ -802,6 +810,10 @@ def search(self, url_prefix):
             for blob in self._tool.blobs_matching_url_prefix(url_prefix)
         ]
 
+    def delete(self, url):
+        blob = self._tool.blob_from_url(url)
```
namanjain, line 814: Same for gcs blobs.

tests/test_cache_api.py
```diff
@@ -131,6 +143,52 @@ def xy(x, y):  # noqa: F811
...
+        return b + 1
+
+    flow = builder.build()
+
+    flow.get("b")
+
+    (b_entry,) = [entry for entry in flow.cache.get_entries() if entry.entity == "b"]
+    b_entry.delete()
+
+    flow.get("c")
```
namanjain, line 189: Can you add a comment here explaining why this is a useful test? I understand that it's because you want to refresh cache accessors so that Bionic doesn't throw invalid cache state exception but I'm sure it would be hard to remember later. I initially got confused thinking the above test is already handling this until I saw that the entities for this test aren't memoized.

bionic/persistence.py
```diff
@@ -530,6 +530,10 @@ def list_items(self):
                 descriptor=metadata_record.descriptor,
             )
 
+    def delete_entry(self, metadata_url, artifact_url):
+        self._fs.delete(metadata_url)
```
namanjain, line 534: Let's delete the artifact before metadata in case there is some error while deleting the artifact url (like connection issues when deleting from gcs).

bionic/deriver.py
```diff
@@ -594,6 +594,14 @@ def run(self, states):
             while self._has_pending_entries():
                 entry = self._activate_next_pending_entry()
 
+                # Before we decide if the current entry is blocked, we need to create
+                # an entry for each of its dependencies. This has the side effect of
+                # refreshing each of their cache states, so if any of them have had
+                # their cache entries deleted, their status will be correctly updated
+                # to "not complete".
+                for dep_state in entry.state.dep_states:
```
namanjain, line 602: Nice!

bionic/task_state.py
```diff
@@ -252,33 +275,18 @@ def sync_after_subprocess_completion(self):
...
         assert self.should_persist
 
-        # First, let's flush the stored entries in cache accessors. This is to prevent cases
-        # like current process contains an outdated stored entries.
-        #
-        # For example: In assisted mode, main process checks for versioning error which can
-        # populate the stored entries as entries without an artifact. Now subprocess can
-        # update the stored entries after computing them but it won't be communicated back.
-        # This allows the main process to fetch fresh entries which might have been updated.
+        # First, let's flush the stored entries in cache accessors. Since we just
```
namanjain, line 278: I think we can replace this method with `refresh_all_cache_state`.

bionic/task_state.py
jqmp: Nice, thanks!

bionic/task_state.py
jqmp: Yes, good point. We're a bit sloppy with the use of "cache" in a few places.

tests/test_cache_api.py
jqmp: Yeah, great idea.

bionic/persistence.py
jqmp: The reason I delete the metadata first is because the metadata refers to the artifact; if you delete the artifact and then fail to delete the metadata, then the metadata will be invalid.

bionic/task_state.py
jqmp: Discussed this offline; we'll leave it for now.

PR #186, opened 2020-07-06
Add an extended ML example

namanjain:
Extend the current toy ML example to use a bigger dataset and compute
ROC AUC.

### DIFF
--- a/docs/tutorials/ml_workflow.ipynb
+++ b/docs/tutorials/ml_workflow.ipynb
@@ -78,7 +78,7 @@
     "* Clean it (``features_frame``)\n",
     "* Split into training and test sets (``train_frame`` and ``test_frame``)\n",
     "* Fit a model on the training data (``model``)\n",
-    "* Evaluate the model on the test data (``precision_recall_frame``)\n",
+    "* Evaluate the model on the test data (``predictions_frame``, ``precision_recall_frame``)\n",
     "* Visualize the model's performance (``all_hyperparams_pr_plot``)\n",
     "\n",
     "As in the previous example, we can access any of the entity values.\n",
--- a/example/ml_workflow.py
+++ b/example/ml_workflow.py
@@ -81,18 +81,30 @@ def model(train_frame, random_seed, hyperparams_dict):
     return m
 
 
+# Predict probabilities for the test data.
+@builder
+def prediction_frame(model, test_frame):
+    """
+    A dataframe with one column, `proba`, containing predicted probabilities for the
+    test data.
+    """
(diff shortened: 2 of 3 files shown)


example/extended_ml_workflow.py
```diff
@@ -0,0 +1,130 @@
...
+
+import bionic as bn
+
+# Convert the example ml workflow to a builder we'll use for extending.
+builder = basic_flow.to_builder()
+
+
+# Predict probabilities for the test data.
+@builder
+def prediction_frame(model, test_frame):
```
jqmp, line 21: Maybe at least some of these entities can be included in the original ML workflow? I think at least `prediction_frame` and `precision_recall_frame` should, since they're basically just a refactoring of the original `precision_recall_frame`. But maybe we should just put all of these in the tutorial, and have this file just be an expanded CLI interface for the same workflow. (In fact, we could remove the CLI from the original file, since it's not helpful in the tutorial anyway.)

example/extended_ml_workflow.py
```diff
@@ -0,0 +1,130 @@
...
+    """
+    A dataframe with probability predictions for the test data.
```
jqmp, line 23: We should mention the name of the predictions column -- that's important for anyone who wants to consume this.

Maybe: "A dataframe with one column, ``proba``, containing predicted probabilities for the test data."

example/extended_ml_workflow.py
```diff
@@ -0,0 +1,130 @@
...
+    df["threshold"] = [0] + list(thresholds) + [1]
+    df["precision"] = list(precisions) + [1]
+    df["recall"] = list(recalls) + [0]
+
+    return df
+
+
+# Summarize the model's performance in one number.
+@builder
+def auc_score(test_frame, prediction_frame):
```
jqmp, line 54: This capitalization is confusing, since it doesn't add up to "AUC". Maybe put parentheses around "Receiver Operating Characteristic"?

example/extended_ml_workflow.py
```diff
@@ -0,0 +1,130 @@
...
+    if args.big_dataset:
+        builder = flow.to_builder()
+
+        @builder
+        @bn.version("covtype dataset")
+        def raw_frame():
+            dataset = datasets.fetch_covtype()
+            feature_names = [f"feature_{ix}" for ix in range(dataset.data.shape[1])]
+            df = pd.DataFrame(data=dataset.data, columns=feature_names)
+            # Pick two as the target since it is the most common shape.
```
jqmp, line 113: Why "shape"? Are cover types shapes?

jqmp:
I force-pushed the changes we discussed to this branch.

PR #189, opened 2020-07-06
Bump version: 0.7.1 → 0.8.0

namanjain:


namanjain:
> We should update the release notes too right?

Good point! Updated the release notes too.

PR #191, opened 2020-07-06
Fix duplicate footnote name in docs

jqmp:


docs/concepts.rst
```diff
@@ -627,7 +627,7 @@ up the cache:
         if entry.tier == 'local' and entry.entity == 'model':
             entry.delete()
 
-.. [#f4] Manually deleting the artifact file itself is *not* safe, because Bionic
+.. [#f5] Manually deleting the artifact file itself is *not* safe, because Bionic
```
namanjain, line 630: Shouldn't this be `f4` and the next one be `f5`?

jqmp: Yes, you're right. I need to look into whether it's possible to use meaningful names here.

PR #195, opened 2020-07-08
Format logs before sending them to the main process

namanjain:
Bionic sends all the log records to the main process and lets the main
process log them. In this case, the log formatting happens in the main
process, which requires us to send the entire argument list of the
logs. The communication between processes requires every argument to be
pickleable using cloudpickle, which seems unreasonable to enforce.

To avoid breaking logging, we eagerly format the log message the way
logging does (check LogRecord.getMessage). Although the logging works,
this has a side effect of changing the template and arguments for the
log in case users intercepts the LogRecord in their custom handlers.

### DIFF
--- a/bionic/executor.py
+++ b/bionic/executor.py
@@ -5,6 +5,7 @@ contains Logging classes used by the Executor to make logging
 work in a parallel setting.
 """
 
+import copy
 import logging
 import queue
 import sys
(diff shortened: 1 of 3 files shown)


bionic/executor.py
```diff
@@ -145,7 +145,28 @@ def __init__(self, queue):
         self._queue = queue
 
     def emit(self, record):
+        # TODO We send all log records to the main process for logging,
+        # even when logging is disabled for their log levels. This is
+        # expensive as we pay IPC cost for the log records and pay the
+        # cost of formating the log record, which can be costly for
```
jqmp, line 151: nit: s/formating/formatting/

bionic/executor.py
```diff
@@ -145,7 +145,28 @@ def __init__(self, queue):
...
+        # cases where users have a ton of disabled debug logs.
+        # We should investigate if we can either caching the log levels
```
jqmp, line 153: s/caching/cache/

bionic/executor.py
```diff
@@ -145,7 +145,28 @@ def __init__(self, queue):
...
+            # communication between processes requires every argument
+            # to be pickleable using cloudpickle. Users can log
+            # unpickleable objects like clients or handlers and expect
+            # the logger to work. To avoid breaking logging, we eagerly
+            # format the log message the way logging does
+            # (check LogRecord.getMessage). This has a side effect of
+            # changing the template and arguments for the log in case
+            # users intercepts the LogRecord in their custom handlers.
+            if len(record.args) > 0:
+                record.msg = record.msg % record.args
```
jqmp, line 168: Should we make a copy of the record before mutating it? I'm wondering if there could be other handlers in this process that will also receive this record.

namanjain: Oh right, I guess users can add handlers to a logger inside the entity functions or we can use this outside Loky subprocesses too.

namanjain:
> Looks good except for my question about copying, plus the extra edge case you mentioned on Slack.

Copying the record just in case, handled the edge case I mentioned in Slack with tests, and added an entry in release notes. PTAL.

docs/release-notes.rst
```diff
@@ -72,6 +72,12 @@ For each release, we list the following types of change (in this order):
    in the "stable" docs (corresponding to the last release) but will be visible in the
    "latest" docs (corresponding to the master branch).
 
+Bug Fixes
+---------
+
+- Parallel execution introduced in 0.8.0 had a bug in logging, where logging breaks
```
jqmp, line 78: "introduced in 0.8.0" should probably be in parentheses (or bracketed by commas). The way it is now, it looks restrictive, as if there were multiple kinds of parallel execution.

docs/release-notes.rst
```diff
@@ -72,6 +72,12 @@ For each release, we list the following types of change (in this order):
...
+  when at least one of the arguments to the log message is unpickleable.
```
jqmp, line 79: We might want to expand on "logging breaks": this causes lost log messages and warnings, but not a crash, right?

namanjain: That's correct. I'll take another stab at this.

PR #196, opened 2020-07-08
Fix my username in codeowners file

namanjain:
My GitHub username is different than my Square LDAP username, updating
my username to the GitHub one in the codeowners file.


namanjain:
> So have we been notifying the user `naman` this whole time?

It doesn't seem like it. None of the PRs added `naman` as a reviewer. I wonder if GitHub is does something smart like verify that the code owners are either an admin or a contributor for the repo.

PR #198, opened 2020-07-09
Janek/fix add case arg order bug

jqmp:
The `FlowBuilder.add_case` method accepts multiple name-value pairs;
these can be given in any order, and are automatically permuted to get
them into a canonical order. Unfortunately, we had a bug where we
instead applied the inverse of the correct permutation.

As it happened, our test for this logic only required permutations with
a single swap -- which meant there was no difference between a correct
permutation and its inverse.

Oh, and that test case was also missing the word `assert` in all its
checks.

### DIFF
--- a/tests/test_flow/test_api.py
+++ b/tests/test_flow/test_api.py
@@ -202,11 +202,12 @@ def test_add_case_out_of_order(builder):
     builder.add_case("p", 1, "q", 10, "r", 100)
     builder.add_case("p", 2, "r", 200, "q", 20)
     builder.add_case("r", 300, "q", 30, "p", 3)
+    builder.add_case("r", 400, "p", 4, "q", 40)
 
     flow = builder.build()
-    flow.get("p", set) == {1, 2, 3}
-    flow.get("q", set) == {10, 20, 30}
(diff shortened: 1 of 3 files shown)


tests/test_flow/test_api.py
```diff
@@ -202,11 +202,12 @@ def test_add_case_out_of_order(builder):
     builder.add_case("p", 1, "q", 10, "r", 100)
     builder.add_case("p", 2, "r", 200, "q", 20)
     builder.add_case("r", 300, "q", 30, "p", 3)
+    builder.add_case("r", 400, "p", 4, "q", 40)
 
     flow = builder.build()
-    flow.get("p", set) == {1, 2, 3}
-    flow.get("q", set) == {10, 20, 30}
-    flow.get("r", set) == {100, 200, 300}
+    assert flow.get("p", set) == {1, 2, 3, 4}
```
namanjain, line 208: hah, classic test issue

jqmp: Yeah... it would be nice if there were a linter for this sort of thing, but a cursory search didn't turn anything up.

PR #200, opened 2020-07-09
Add release steps to the documentation

namanjain:
Verified that it renders correctly and all links work. Here is the snapshot of my local sphinx build.

Screen Shot 2020-07-09 at 4 03 33 PM


docs/index.rst
```diff
@@ -59,5 +59,6 @@ Documentation Contents
     api/index.rst
     get-help
     contributing
+    maintaining
```
jqmp, line 62: WDYT about linking to this from the Contributing page instead of having it on the sidebar?

PR #203, opened 2020-07-10
Submit task completion runner to the subprocesses

namanjain:
Instead of submitting the task state to the subprocesses, we now send
the task completion runner and the task state. This is a step towards
making the runner resolve dependencies stack-based instead of mixing
the stack and recursion-based execution.

This change was a part of #199


bionic/deriver.py
```diff
@@ -463,6 +468,9 @@ class Bootstrap:
     executor = attr.ib()
     should_memoize_default = attr.ib()
 
+    def evolve(self, executor):
```
jqmp, line 471: Rather than having a method specifically for `executor`, should we just accept `**kwargs` here?

PR #204, opened 2020-07-13
Compute memoize-only entries using stack-based execution

namanjain:
Bionic computes entries using a stack-based execution logic. But for
some memoize-only entries, it uses a recursion-based approach. We
change the logic to compute such entities using the unified stack-based
approach.

It does so by adding a `SKIPPED` EntryStage to differentiate between
skipped and completed entries. We also track whether entries would be
needed in memory and computing memoized-only entries when they are.

### DIFF
--- a/bionic/core/flow_execution.py
+++ b/bionic/core/flow_execution.py
@@ -66,9 +66,26 @@ class TaskCompletionRunner:
                 # their cache entries deleted, their status will be correctly updated
                 # to "not complete".
                 for dep_state in entry.state.dep_states:
-                    self._get_or_create_entry_for_state(dep_state)
+                    dep_entry = self._get_or_create_entry_for_state(dep_state)
+                    # If the entry is needed in memory, any non-serializable entry that
+                    # can be memoized is also needed in memory.
+                    if entry.is_needed_in_memory:
+                        if (
+                            dep_entry.state.should_memoize
+                            and not dep_entry.state.should_persist
+                        ):
+                            dep_entry.is_needed_in_memory = True
+                            if dep_entry.stage == EntryStage.DEFERRED:
+                                self._mark_entry_pending(dep_entry)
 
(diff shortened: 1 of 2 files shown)


bionic/core/flow_execution.py
```diff
@@ -66,9 +66,26 @@ def run(self, states):
                 # their cache entries deleted, their status will be correctly updated
                 # to "not complete".
                 for dep_state in entry.state.dep_states:
-                    self._get_or_create_entry_for_state(dep_state)
+                    dep_entry = self._get_or_create_entry_for_state(dep_state)
```
jqmp, line 69: As we discussed on Slack, WDYT about replacing `is_requested` with `is_needed_in_memory` and having that be an argument to `_get_or_create_entry_for_state`?

bionic/core/task_execution.py
```diff
@@ -414,6 +450,20 @@ def initialize(self, bootstrap, flow_instance_uuid):
             for task_key in self.task_keys
         ]
 
+        self.set_up_caching_flags(bootstrap)
```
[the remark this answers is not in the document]
namanjain: We need `should_persist` caching flag before we can set up cache accessors. I can move setting up flags to the top though.

bionic/core/flow_execution.py
namanjain: While making this change, I realized that I'm setting up caching flags for state inside `_get_or_create_entry_for_state`. We need the should persist cache flag to determine whether an entry is needed in memory.

Wdyt about passing `is_requested` to  `_get_or_create_entry_for_state` but replacing `is_requested` with `is_needed_in_memory` inside the entry?

namanjain: This is going to be easier to address in #205, so I'm going to address it in that PR.

PR #206, opened 2020-07-13
Make default persistence configurable

namanjain:
Like #181, this change introduces a new core entity that lets users
control the default persistence value.

Users can continue to override the default persistence by using
@persist and @immediate decorators.

Here are screenshots for documentation updates:

Screen Shot 2020-07-13 at 2 22 28 PM
Screen Shot 2020-07-13 at 2 22 40 PM

### DIFF
--- a/bionic/core/task_execution.py
+++ b/bionic/core/task_execution.py
@@ -446,17 +446,17 @@ class TaskState:
         # grouped entities are created). At any rate, once we have tuple
         # descriptors, each task state will only be responsible for a single entity
         # and this won't be an issue.
-        optional_should_memoize, can_persist = single_unique_element(
-            (entity_def.optional_should_memoize, entity_def.can_persist)
+        optional_should_memoize, optional_should_persist = single_unique_element(
+            (entity_def.optional_should_memoize, entity_def.optional_should_persist)
             for entity_def in self.entity_defs_by_dnode.values()
         )
 
-        should_memoize = optional_should_memoize
-        if should_memoize is None:
-            if bootstrap is None:
-                should_memoize = True
-            else:
-                should_memoize = bootstrap.should_memoize_default
+        if optional_should_memoize is not None:
+            should_memoize = optional_should_memoize
+        elif bootstrap is not None:
+            should_memoize = bootstrap.should_memoize_default
+        else:
+            should_memoize = True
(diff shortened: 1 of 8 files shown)


bionic/core/task_execution.py
```diff
@@ -478,7 +478,13 @@ def set_up_caching_flags(self, bootstrap):
...
         self.should_memoize = should_memoize
 
-        should_persist = can_persist and not self.output_would_be_missing()
+        should_persist = optional_should_persist
+        if should_persist is None:
+            if bootstrap is None:
+                should_persist = False
+            else:
+                should_persist = bootstrap.should_persist_default
+        should_persist = should_persist and not self.output_would_be_missing()
```
jqmp, line 487: WDYT about writing it like this?

    if self.output_would_be_missing():
        should_persist = False
    elif optional_should_persist is not None:
        should_persist = optional_should_persist
    elif bootstrap is not None:
        should_persist = bootstrap.should_persist_default
    else:
        should_persist = False

This avoids mutation and feels easier for me to follow than the current code (which sets `should_persist`, then reads it, then sets it, then reads it, then sets it).

docs/concepts.rst
```diff
@@ -359,6 +359,24 @@ we can disable persistent caching altogether:
 If your goal is just to force an entity to be recomputed more frequently, you may
 want :ref:`@changes_per_run <changes_per_run>` instead.
 
+Persistent caching can also be globally disabled:
+
+.. code-block:: python
+
+    builder.set('core__persist_by_default', False)
+
+This only changes the default behavior, so it can be explicitly re-enabled for
```
jqmp, line 368: Do we have any documentation that explains the consequences of turning off both persistence and memoization?

docs/concepts.rst
```diff
@@ -375,14 +393,13 @@ cases, we can disable in-memory caching:
     def message(subject):
         return f'Hello {subject}.'
 
-In-memory caching can also be globally disabled:
+Like persistent caching, in-memory caching can also be globally disabled:
```
jqmp, line 396: Nice touch-up here.

docs/concepts.rst
namanjain: I don't think so. Let's talk on slack what we can add for it.

PR #207, opened 2020-07-14
Add JsonProtocol

namanjain:
This change adds a new protocol that supports JSON serialization for
simple types, namely, int, float, string, bool, list, and dict.

We might add further improvements to this protocol where users can
add a custom encoder for class types.

Release notes screenshot:

Screen Shot 2020-07-14 at 4 41 25 PM

### DIFF
--- a/bionic/flow.py
+++ b/bionic/flow.py
@@ -51,6 +51,7 @@ from .util import (
 )
 
 DEFAULT_PROTOCOL = protos.CombinedProtocol(
+    protos.JsonProtocol(),
     # Every GeoPandas DataFrame is also a Pandas DataFrame, so we need to do the
     # GeoPandas check first.
     protos.GeoPandasProtocol(),
--- a/bionic/protocols.py
+++ b/bionic/protocols.py
@@ -10,6 +10,7 @@ This module contains a BaseProtocol class and various subclasses.
 """
 
 from collections import Counter
+import json
 import pickle
 import sys
 import tempfile
@@ -195,6 +196,43 @@ class BaseProtocol:
 
     def __repr__(self):
         return f"{self.__class__.__name__}(...)"
+
+
+class JsonProtocol(BaseProtocol):
+    """
+    Decorator indicating that an entity's values are built-in types that are
+    JSON-serializable: the supported types are int, float, str, bool, list,
+    and dict. Note that dict keys must be strings, and each element of a list
+    or dict must itself be a supported built-in type.
+
+    These values will be serialized to JSON files.
+    """
+
+    def get_fixed_file_extension(self):
+        return "json"
+
+    def validate(self, value):
+        if value is None:
+            return
+        if isinstance(value, list):
+            for elem in value:
+                self.validate(elem)
+            return
+        if isinstance(value, dict):
+            for key, elem in value.items():
+                assert isinstance(key, str)
+                self.validate(elem)
+            return
+
+        assert isinstance(value, (int, float, str, bool))
+
+    def write(self, value, path):
+        with path.open("w", encoding="utf-8") as file_:
+            json.dump(value, file_, ensure_ascii=False)
+
+    def read(self, path):
+        with path.open("r", encoding="utf-8") as file_:
+            return json.load(file_)
 
 
 class PicklableProtocol(BaseProtocol):
@@ -472,11 +510,11 @@ class YamlProtocol(BaseProtocol):
         return "yaml"
 
     def write(self, value, path):
-        with path.open("w") as file_:
+        with path.open("w", encoding="utf-8") as file_:
             yaml.dump(value, file_, **self._kwargs)
 
     def read(self, path):
-        with path.open("r") as file_:
+        with path.open("r", encoding="utf-8") as file_:
             return yaml.safe_load(file_)
 
 
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -584,6 +584,8 @@ object types it's more efficient or more idiomatic to use a different format.
 There are also some types of objects that can't be pickled at all.  Bionic uses
 ``pickle`` by default, but handles some types specially:
 
+- `JSON <https://www.json.org/json-en.html>`_-serializable built-in types (int, float,
+  str, bool, list, and dict) are serialized as JSON files.
 - `Pandas <https://pandas.pydata.org/>`_ DataFrames are serialized as
   `Parquet <https://parquet.apache.org/>`_ files.
 - `NumPy <https://numpy.org/>`_ Arrays are serialized as `NPY
--- a/tests/test_flow/test_protocols.py
+++ b/tests/test_flow/test_protocols.py
@@ -22,6 +22,18 @@ from bionic.protocols import CombinedProtocol, PicklableProtocol
 from bionic.util import recursively_delete_path
 
 
+JSONABLE_VALUES = [
+    None,
+    1,
+    1.1,
+    "string",
+    True,
+    [1, 2, [3]],
+    {"a": 1, "b": {"c": [2]}, "d": None},
+    {"ń": "ņ"},
+]
+
+
 PICKLABLE_VALUES = [
     1,
     "string",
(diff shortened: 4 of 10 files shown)


bionic/protocols.py
```diff
@@ -197,6 +198,29 @@ def __repr__(self):
...
+    """
+    Decorator indicating that an entity's values can be serialized using the
+    ``json`` library.
+    """
+
+    def get_fixed_file_extension(self):
+        return "json"
+
+    def validate(self, value):
+        # We don't count sequence types and dict as valid values. Otherwise, we
```
jqmp, line 211: I feel like we might as well support lists and dicts here too -- as a user, I would expect to be able to serialize a dict as JSON.

bionic/protocols.py
```diff
@@ -197,6 +198,29 @@ def __repr__(self):
         return f"{self.__class__.__name__}(...)"
 
 
+class JsonProtocol(BaseProtocol):
+    """
+    Decorator indicating that an entity's values can be serialized using the
+    ``json`` library.
```
jqmp, line 204: I think the important thing here is that we are serializing to the JSON format -- not the particular library we're using.

bionic/protocols.py
```diff
@@ -197,6 +198,29 @@ def __repr__(self):
...
+    """
+
+    def get_fixed_file_extension(self):
+        return "json"
+
+    def validate(self, value):
+        # We don't count sequence types and dict as valid values. Otherwise, we
+        # will have to validate each entries in the sequences and dicts.
+        assert isinstance(value, (int, float, str, bool))
```
jqmp, line 213: Should we support `None`?

namanjain:
> Don't forget to add the protocol to the API reference and to the release notes!
> 
> Also, in [this section](https://bionic.readthedocs.io/en/stable/concepts.html#serialization-protocols) of the docs we describe the built-in protocols we use -- we'll want to mention JSON here now.

I think users would expect JSON procotol to work for a lot of classes that our current protocol doesn't support. They might also want to register a custom encoder with their protocol. I didn't add it to public documentation on purpose. I think we should wait till we sort all these questions before telling users about this protocol. Let's discuss this in our meeting on hangout.

bionic/protocols.py
namanjain: Good point, I think we should support None.

namanjain:
> I think users would expect JSON procotol to work for a lot of classes that our current protocol doesn't support. They might also want to register a custom encoder with their protocol. I didn't add it to public documentation on purpose. I think we should wait till we sort all these questions before telling users about this protocol. Let's discuss this in our meeting on hangout.

After talking about this, we decided with Janek's proposal to add it to the public documentation, specially because we document the default protocol behavior. We can add support for custom encoders later when or if there is a need.

bionic/protocols.py
```diff
@@ -197,6 +198,39 @@ def __repr__(self):
...
+    def validate(self, value):
+        if value is None:
+            return
+        if isinstance(value, list):
+            for elem in value:
+                self.validate(elem)
+            return
+        if isinstance(value, dict):
+            for key, elem in value.items():
+                self.validate(key)
```
jqmp, line 219: I think we may want to require that the key is a string, not just a valid value; it seems like JSON will accept other values but convert them to strings, which is probably not what we want.

namanjain: Ah got it. I thought keys could be ints and bools and even None but it looks like keys have to be strings. Fixed.

bionic/protocols.py
```diff
@@ -197,6 +198,39 @@ def __repr__(self):
...
+class JsonProtocol(BaseProtocol):
+    """
+    Decorator indicating that an entity's values can be serialized to the
+    ``JSON`` format.
+    """
+
+    def get_fixed_file_extension(self):
+        return "json"
+
+    def validate(self, value):
```
jqmp, line 210: So, while reading the docs for `json.dump`, I realized there's another edge case: a list or dict that contains itself. I've never seen a use for this, but it's actually well-behaved in Python: you can print it, pickle it, etc.

I think your code will still work here (by throwing a recursion exception, which will cause validation to fail), but we might want a test case for this.

tests/test_flow/test_protocols.py
```diff
@@ -22,6 +22,17 @@
...
 
 
+JSONABLE_VALUES = [
+    None,
+    1,
+    1.1,
+    "string",
+    True,
+    [1, 2, [3]],
+    {"a": 1, "b": {"c": [2]}, "d": None},
```
jqmp, line 32: Should we have a unicode test case as well?

bionic/protocols.py
```diff
@@ -197,6 +198,39 @@ def __repr__(self):
         return f"{self.__class__.__name__}(...)"
 
 
+class JsonProtocol(BaseProtocol):
+    """
+    Decorator indicating that an entity's values can be serialized to the
+    ``JSON`` format.
```
jqmp, line 204: I think it makes sense to omit the backquotes here, or even just say "can be serialized to JSON". The backquotes make this render like code, which it isn't.

bionic/flow.py
```diff
@@ -53,6 +53,7 @@
 DEFAULT_PROTOCOL = protos.CombinedProtocol(
     # Every GeoPandas DataFrame is also a Pandas DataFrame, so we need to do the
```
jqmp, line 54: This comment should be next to the GeoPandas line.

bionic/protocols.py
```diff
@@ -197,6 +198,39 @@ def __repr__(self):
...
+            for key, elem in value.items():
+                assert isinstance(key, str)
+                self.validate(elem)
+            return
+
+        assert isinstance(value, (int, float, str, bool))
+
+    def write(self, value, path):
+        with path.open("w", encoding="utf-8") as file_:
+            json.dump(value, file_, ensure_ascii=False)
```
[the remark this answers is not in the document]
namanjain: It doesn't escape non-ascii characters in the output. Now that users can read cached artifacts directly, it would be nice if they can see the characters as they are instead of escaped values.

bionic/protocols.py
```diff
@@ -197,6 +198,39 @@ def __repr__(self):
...
+
+    def read(self, path):
+        with path.open("r") as file_:
```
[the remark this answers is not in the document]
namanjain: Oh I probably copied the read write methods from pickle protocol and rewrote the write method. Given that we don't give it an encoding for other protocols (only yaml to be fair since it's the only one opening it in text mode), I'll keep it consistent and remove the encoding for write.

bionic/protocols.py
namanjain: TIL you can pickle circular references. Added a test for it and it fails with `RecursionError`. I don't really like relying on this error but it still feels better than tracking all the seen objects during validation. Wdyt?

jqmp: Yeah, I could go either way. Relying on this particular error is not great, but it's also a super rare case.

bionic/protocols.py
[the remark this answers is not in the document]
jqmp: Make sense to me!

bionic/protocols.py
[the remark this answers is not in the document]
jqmp: Hmm... if you don't specify an encoding, what encoding does it use?

bionic/protocols.py
```diff
@@ -197,6 +198,39 @@ def __repr__(self):
         return f"{self.__class__.__name__}(...)"
 
 
+class JsonProtocol(BaseProtocol):
+    """
+    Decorator indicating that an entity's values can be serialized to the
```
jqmp, line 203: I just realized that it might be good to specify the exact set of types we support here, since "can be serialized to JSON" is a bit vague.

bionic/protocols.py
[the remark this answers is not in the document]
namanjain: We talked more about this off-GitHub and decided to add encoding to all text files explicitly.

docs/concepts.rst
```diff
@@ -584,6 +584,8 @@ object types it's more efficient or more idiomatic to use a different format.
 There are also some types of objects that can't be pickled at all.  Bionic uses
 ``pickle`` by default, but handles some types specially:
 
+- `JSON <https://www.json.org/json-en.html>`_-serializable built-in types are
```
jqmp, line 587: Sorry I missed this in our Slack conversation, but we should probably be explicit about what "JSON-serializable" means here.

PR #208, opened 2020-07-16
Refactor tests

jqmp:
This has four commits:

- Move some flow tests into test_flow (simple file moves)
- Move flow fixtures into test_flow/conftest.py (break one file into two)
- Refactor test skipping (code refactor)
- Don't skip parallel tests, just ignore them (the only functional change in this PR)


tests/conftest.py
```diff
@@ -101,93 +17,45 @@ def pytest_addoption(parser):
...
-
-    return builder
+def pytest_collection_modifyitems(config, items):
+    run_slow = config.getoption("--slow")
+    skip_slow = pytest.mark.skip(reason="only runs when --slow is set")
+
+    has_gcs = config.getoption("--bucket")
+    skip_needs_gcs = pytest.mark.skip(reason="only runs when --bucket is set")
+
+    run_parallel = config.getoption("--all-execution-modes")
```
namanjain, line 44: This reads like it will only run parallel execution mode. How about `can_also_run_parallel` or `run_all_modes`?

tests/conftest.py
```diff
@@ -101,93 +17,45 @@ def pytest_addoption(parser):
...
+
+    items_to_keep = []
+    for item in items:
+        if "slow" in item.keywords and not run_slow:
+            item.add_marker(skip_slow)
+        if "needs_gcs" in item.keywords and not has_gcs:
+            item.add_marker(skip_needs_gcs)
+        if "parallel" in item.keywords:
```
namanjain, line 52: nit: It reads a little weird since it's all `if` conditionals in the same block and only the last one is an `elif`. Either:
- make all of them `elif`s and move the continue conditionals to top, or
- add a newline before this `if`, or
- replace `elif` with `if "parallel" not in item.keywords`.

tests/conftest.py
```diff
@@ -101,93 +17,45 @@ def pytest_addoption(parser):
...
+            if not (run_parallel or "allows_parallel" in item.keywords):
+                continue
+        elif "needs_parallel" in item.keywords:
+            continue
+
+        items_to_keep.append(item)
+
+    items.clear()
```
namanjain, line 60: This is really nice!

tests/conftest.py
jqmp: Oh, good point. I think I'll go with the newline here.

tests/conftest.py
jqmp: Hmm... I can see what you mean. That didn't strike me as the most likely interpretation (same as with `run_slow`) but I can make it more specific.

PR #210, opened 2020-07-17
Store inventory entries immediately on creation

jqmp:
I noticed that when we write to the persistent cache, we flush its
stored entries and don't reconstruct them, so we end up looking up the
same artifact again immediately. I'm not sure if this makes a real
performance difference, but it does increase the number of round-trips
to GCS, so it seemed worthwhile to fix it.

### DIFF
--- a/bionic/persistence.py
+++ b/bionic/persistence.py
@@ -216,11 +216,6 @@ class CacheAccessor:
     def _save_or_reregister_result(self, result):
         local_entry = self._get_local_entry()
         cloud_entry = self._get_cloud_entry()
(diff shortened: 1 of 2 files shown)


bionic/persistence.py
```diff
@@ -259,7 +254,9 @@ def _save_or_reregister_result(self, result):
 
         if not local_entry.exactly_matches_query:
```
namanjain, line 255: If `local_entry` exactly matches the query, we should set it to `self._stored_local_entry`. Otherwise the stored local entry would still be flushed.

Same for `cloud_entry`.

PR #211, opened 2020-07-17
Refactor utility function module

jqmp:
This converts `util.py` into a `util` package, with separate modules for different types of utilities. I also moved some generic helper functions from `persistence` into this new module, and created separate top-level modules for `oneline` and `FileCopier`.


tests/test_util.py
```diff
@@ -2,11 +2,9 @@
 
 from .helpers import equal_when_sorted
```
namanjain, line 3: It doesn't have to be in this PR, but how you feel about restructuring non-flow (unit) tests to match the dir structure of src code? It would mean that we will have a `tests/util` dir that have multiple test files, corresponding to each files in the src `util` package.

bionic/util/__init__.py
```diff
@@ -0,0 +1,8 @@
+"""
```
namanjain, line 1: How do you feel about calling it util**s**?

bionic/oneline.py
```diff
@@ -0,0 +1,19 @@
+"""
+Contains the oneline method, which is very useful for formatting error messages.
+
+This is in its own module, rather than somewhere in ``util``, because it's used in
```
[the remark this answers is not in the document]
jqmp: I thought about this and I agree that it would be nice to simplify the dependency story. However, it's not obvious to me that "`util` should not depend on `deps`". It seems possible to have other utilities that happen to depend on external packages. If we do more with Dask we'll probably have Dask-specific utils, etc. We could put those in their own top-level groups, but I wonder if we'll just end up having `util` and then `util_with_deps`.

On the other hand, we can always deal with that problem when it actually arises. WDYT?

tests/test_util.py
jqmp: Yeah, I think that makes sense!

bionic/util/__init__.py
jqmp: I feel good about it!

bionic/oneline.py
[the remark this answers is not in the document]
namanjain: I'm fine with dealing with it later.

jqmp:
@namanjain I think I implemented all your suggestions; let me know what you think if you have a chance.

namanjain:
> @namanjain I think I implemented all your suggestions; let me know what you think if you have a chance.

LGTM!

PR #216, opened 2020-07-28
Add a test for entries needed in memory

namanjain:
This change adds a test for entries needed in memory. The test also
verifies the behavior when a deferred entry moves back to pending
stage.

### DIFF
--- a/tests/test_flow/test_persistence.py
+++ b/tests/test_flow/test_persistence.py
@@ -1208,3 +1208,53 @@ def test_multiple_outputs_all_persisted_at_once(builder, make_counter):
     assert builder.build().get("x") == 1
     assert builder.build().get("y") == 2
     assert counter.times_called() == 1
+
+
+@pytest.mark.needs_parallel
+def test_entries_needed_in_memory(builder, make_counter):
+    """
+      ----> B ---->
+    E               A
+      ----> C ---->
+
+    A depends on B and C
+    B and C depends on D
+    Only C is persisted
+
+    This structure is complex because it also tests for the case where
+    a deferred entity is moved back to pending stage. In this case,
+    D is deferred by C but then is required in memory by B.
+    """
+    builder.set("core__persist_by_default", False)
+
+    a_counter = make_counter()
+    b_counter = make_counter()
+    c_counter = make_counter()
+    d_counter = make_counter()
+
+    @builder
+    @d_counter
+    def d():
+        return 1
+
+    @builder
+    @bn.persist(True)
+    @c_counter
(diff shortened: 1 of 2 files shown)


tests/test_flow/test_persistence.py
```diff
@@ -1208,3 +1208,53 @@ def x_y():
     assert builder.build().get("x") == 1
     assert builder.build().get("y") == 2
     assert counter.times_called() == 1
+
+
+@pytest.mark.needs_parallel
+def test_entries_needed_in_memory(builder, make_counter):
+    """
+      ----> B ---->
```
jqmp, line 1216: Wouldn't it make more sense to go from `A` to `D` instead of from `D` to `A`?

tests/test_flow/test_persistence.py
```diff
@@ -1208,3 +1208,53 @@ def x_y():
...
+    @b_counter
+    def b(d):
+        return 3 * d  # 3
+
+    @builder
+    @a_counter
+    def a(b, c):
+        return b + c  # 10
+
+    builder.build().get("a") == 13
```
jqmp, line 1256: Doesn't the comment above indicate that the value should be `10`? And maybe we should have an `assert` here? :)

tests/test_flow/test_persistence.py
```diff
@@ -1208,3 +1208,53 @@ def x_y():
...
+
+
+@pytest.mark.needs_parallel
+def test_entries_needed_in_memory(builder, make_counter):
+    """
+      ----> B ---->
+    E               A
+      ----> C ---->
+
+    A depends on B and C
```
jqmp, line 1220: Our goal is to test an entry being moved from deferred to pending, but doesn't that depend on the order in which we process the entries? That is, we're assuming that D gets requested (but not in memory) by C before it gets requested (in memory) by B? Is it possible that D can be completed before it gets requested by C?

tests/test_flow/test_persistence.py
```diff
@@ -1208,3 +1208,53 @@ def x_y():
...
+    B and C depends on D
+    Only C is persisted
+
+    This structure is complex because it also tests for the case where
+    a deferred entity is moved back to pending stage. In this case,
+    D is deferred by C but then is required in memory by B.
+    """
+    builder.set("core__persist_by_default", False)
```
jqmp, line 1228: Personally I think I would find this clearer if we just used `@persist(False)` for the entities we don't want to persist.

namanjain:
Declining this PR since the flag tested in this change is removed in #218

Issue #111, opened 2020-04-22
Non-persistable entities can sometimes be spuriously recomputed


jqmp:
Here's a failing test:

    def test_non_persisted_intermediate(builder):
        builder.assign('x', 2)

        @builder
        @bn.persist(False)
        @count_calls
        def x_plus_one(x):
            return x + 1

        @builder
        @count_calls
        def x_plus_two(x_plus_one):
            return x_plus_one + 1

        flow = builder.build()
        assert flow.get('x_plus_two') == 4
        assert x_plus_one.times_called() == 1
        assert x_plus_two.times_called() == 1

        flow = builder.build()
        assert flow.get('x_plus_two') == 4
        # This should return 0, because we don't need to compute x_plus_one;
        # we should just load x_plus_two from disk.
        # However, it returns 1!
        assert x_plus_one.times_called() == 0

This happens because when we request `x_plus_two`, Bionic tries to "complete" all the upstream entities, and for a non-persistable entity that means computing it. However, in this case all we need is the hashed value of `x_plus_one`, which we can figure out without actually computing it. The solution is probably to have two different levels of "completeness": one where just the hash is available, and one where the full result is available.

PR #218, opened 2020-07-30
Avoid recomputing deps unnecessarily

jqmp:
This is a substantial refactor of TaskCompletionRunner to avoid
recomputing non-persistable entities before loading their dependents
from the cache. The problem is this: in order to load a value from the
persistent cache, we need to compute the provenance digest of each of
its dependencies, which currently requires computing them. This means
that if a task has any non-persistable dependencies, they need to be
recomputed every time we want to load it from the cache. However, it's
also the case that when tasks have non-persistable output, their
provenance digest doesn't depend on their value, so this extra
computation is not even required.

To avoid this computation, we allow these non-persistable tasks to be in
a special "primed" state, where their digest has been computed but their
actual value has not. As it happens, this state is also useful for
tracking whether these tasks are ready to run in a subprocess.

I also made the following adjustments:

- I removed the DEFERRED stage for TaskEntries. I think the right way to
think about entry stages is that an entry is COMPLETED as soon as it
has met all known requirements; however, an entry can become
un-COMPLETED when new requirements become known.

- Previously task entries had a concept of "needed in memory" vs "needed
but not in memory". I changed this to "requested to have cached values"
vs "requested to be primed", which I think is a more useful distinction.

- The `_process_active_entry` method now contains the core of our
asynchronous logic, which has become more complex. The goal here is to
do as little work as possible to meet the requirements on an entry.

Fixes https://github.com/square/bionic/issues/111.

### DIFF
--- a/bionic/core/flow_execution.py
+++ b/bionic/core/flow_execution.py
@@ -26,10 +26,24 @@ def run_in_subprocess(task_completion_runner, state):
 
 class TaskCompletionRunner:
     """
-    Runs `TaskState` to completion.
+    Computes a DAG of `TaskState` objects.
 
-    Using a `Bootstrap` object, this class completes given task states
-    using a stack-based state tracking approach.
+    Given a collection of TaskStates, this class identifies and performs the necessary
+    work to compute their final values. With a simple directed acyclic graph of tasks,
+    this would be a simple as doing a topological sort and then computing the tasks
+    in order. However, there are several complicating factors:
+
+    - Some tasks should be run asynchronously in a separate process to allow
+      parallelism.
+    - Some task outputs can't be serialized, which means those tasks need to be
+      re-computed in each process that needs them.
+    - If a task's value is cached, we want to load it as quickly as possible, without
+      computing any unnecessary dependencies. (However, sometimes we do need to compute
+      dependencies in order to determine if the cached value is up-to-date or not.)
+
+    To deal with these, for each `TaskState` we maintain a `TaskRunnerEntry` which
+    tracks what is required for that particular task. Each entry has an `EntryStage`
+    enum, indicating whether we need to and/or can do more work on it.
     """
 
     def __init__(self, bootstrap, flow_instance_uuid, task_key_logger):
@@ -55,43 +69,13 @@ class TaskCompletionRunner:
 
             for state in states:
                 state.set_up_caching_flags(self._bootstrap)
-                entry = self._get_or_create_entry_for_state(
-                    state, is_needed_in_memory=not state.should_persist
-                )
-                self._mark_entry_pending(entry)
+                entry = self._get_or_create_entry_for_state(state)
+                self._request_entry_be_cached(entry)
 
             while self._has_pending_entries():
                 entry = self._activate_next_pending_entry()
-
-                # Before we decide if the current entry is blocked, we need to create
-                # an entry for each of its dependencies. This has the side effect of
-                # refreshing each of their cache states, so if any of them have had
-                # their cache entries deleted, their status will be correctly updated
-                # to "not complete".
-                dep_entries = []
-                for dep_state in entry.state.dep_states:
-                    dep_state.set_up_caching_flags(self._bootstrap)
-                    # If the entry is needed in memory, any non-serializable entries are
-                    # also needed in memory.
-                    dep_entry_needed_in_memory = (
-                        entry.is_needed_in_memory and not dep_state.should_persist
-                    )
-                    dep_entry = self._get_or_create_entry_for_state(
-                        dep_state, dep_entry_needed_in_memory
-                    )
-                    dep_entries.append(dep_entry)
-                if not entry.are_dep_entries_set():
-                    entry.set_dep_entries(dep_entries)
-
-                if entry.state.is_cached:
-                    self._mark_entry_completed(entry)
-                    continue
-
-                if len(entry.blocking_dep_entries()) > 0:
-                    self._mark_entry_blocked(entry)
-                    continue
-
-                self._process_entry(entry)
+                self._process_active_entry(entry)
+                assert entry.stage != EntryStage.ACTIVE
 
             assert len(self._pending_entries) == 0
             assert len(self._in_progress_entries) == 0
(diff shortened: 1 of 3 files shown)


bionic/core/task_execution.py
```diff
@@ -16,46 +32,78 @@
...
-        # are never deferred.
-        # NOTE An entry that is not needed in memory by one entry can
-        # be needed in memory by another entry. Depending on our
-        # execution order of such entities, this flag can change from
-        # `False` to `True` in between execution.
-        self.is_needed_in_memory = is_needed_in_memory
+        # These are initially set to False, but can be updated to True when we discover
+        # a requirement from another entry. However, they should never go from True to
+        # False!
+        self.is_requested_to_be_primed = False
```
jqmp, line 85: As discussed in a call, we should add a TODO to do the following:
1. Replace these two booleans with an enum.
2. Replace the `EntryBlockage` concept with some kind of `Requirement` object that represents a specific requirement level (using the enum above) on one entry from another.

bionic/core/task_execution.py
```diff
@@ -16,46 +32,78 @@
...
+    @property
+    def is_primed(self):
+        if self.state.should_persist is None:
+            return False
+        if self.state.should_persist:
+            return self.state.is_cached
+        else:
+            return self.state.is_initialized
 
     def blocking_dep_entries(self):
```
namanjain, line 101: Thanks for adding the other TODO. As we discussed on the same call, can you also leave a TODO here to merge the logic for `blocking_dep_entries` and `all_dependencies_satisfy_requests`?

bionic/core/flow_execution.py
```diff
@@ -108,32 +92,59 @@ def run(self, states):
...
-            new_state_for_subprocess = state.strip_state_for_subprocess()
+            return
+
+        # Otherwise, we'll need all of our dependencies to be primed so that their
+        # digests are available.
+        self._set_up_entry_dependencies(entry)
+        for dep_entry in entry.dep_entries:
+            self._request_entry_be_primed(dep_entry)
+        if not entry.all_dependencies_satisfy_requests:
+            self._mark_entry_blocked(entry)
```
namanjain, line 122: `_mark_entry_blocked` also tries to mark blocking entries pending, which `_request_entry_be_primed` and `_request_entry_be_cached` already do it for you. Can we combine those logic as well, maybe assert that blocking deps are not completed instead of marking them as pending?

bionic/core/flow_execution.py
```diff
@@ -108,32 +92,59 @@ def run(self, states):
...
+        # In theory we could have an entry that doesn't require priming, but in practice
+        # we always require it when we create one.
+        assert entry.is_requested_to_be_primed
 
-        # Attempt to complete the task state from persistence cache.
-        state.attempt_to_complete_from_cache()
-        if state.is_cached:
+        # First, if we already have a cached value, we can stop immediately without
+        # needing to process any dependencies.
+        entry.state.set_up_caching_flags(self._bootstrap)
```
namanjain, line 111: I don't think you need to set up the caching flags before initialize anymore. I had to set them up earlier because we used the caching flags to determine whether dependency entries were needed in memory.

bionic/core/flow_execution.py
```diff
@@ -145,28 +156,55 @@ def _process_entry(self, entry):
...
+        entry.compute(self.task_key_logger)
+        self._mark_entry_completed(entry)
+
+    def _set_up_entry_dependencies(self, entry):
+        if entry.dep_entries is not None:
+            return
+
+        dep_entries = []
+        for dep_state in entry.state.dep_states:
+            dep_state.set_up_caching_flags(self._bootstrap)
```
namanjain, line 177: Same here. I don't think we need this anymore.

bionic/core/flow_execution.py
jqmp: Yes, good point -- thanks!

jqmp:
I addressed the feedback and added two more changes:

1. I added a debug-level log statement for every entry stage-change. This makes debugging a lot easier!
2. I merged the `NEW` and `COMPLETED` stages. At this point I think they basically mean the same thing.

bionic/core/task_execution.py
```diff
@@ -147,65 +222,63 @@ def compute(self, task_key_logger):
...
+    requests have been satisfied).
+
     Valid next stages: [PENDING]
     """
-    NEW = auto()
+    COMPLETED = auto()
 
     """
-    Entry is waiting to be processed.
+    The entry is waiting to be processed. We know there's work to be done on it, but
```
namanjain, line 252: We create entries only when we need to process them. Since they start from `COMPLETED` state, we move them to `PENDING` to process them as soon as they are created.

How do you feel about making `PENDING` the starting stage? Or do you see a case in the future where we would create entries without knowing whether we need to process them? If not, I think it's better if we start from `PENDING` and make the change later on.

bionic/core/task_execution.py
```diff
@@ -505,7 +571,7 @@ def set_up_caching_flags(self, bootstrap):
             which may cause the values to be subtly different.
             To avoid this warning, disable persistence for the decorators
             by {disable_message}."""
-            logging.warn(message)
+            logger.warn(message)
```
namanjain, line 574: This change does not belong in this PR but I just realized we are using `warnings.warn` in some places and `logger.warn` in other. We should make these warnings consistent.

bionic/core/flow_execution.py
```diff
@@ -210,7 +241,8 @@ def _mark_entry_blocked(self, blocked_entry):
         for blocking_tk in blocking_tks:
             self._blockage_lists_by_blocking_tk[blocking_tk].append(blockage)
 
-        self._mark_blocking_entries_pending(blocked_entry)
+        for blocking_entry in blocked_entry.blocking_dep_entries():
+            assert blocking_entry.stage != EntryStage.COMPLETED
```
namanjain, line 245: Nice!

bionic/core/task_execution.py
```diff
@@ -585,17 +651,31 @@ def _load_value_hashes(self):
                     Failed to load cached value (hash) for descriptor
                     {accessor.query.dnode.to_descriptor()!r}.
                     This suggests we did not successfully complete the task
```
namanjain, line 653: nit: `complete` -> `compute`

namanjain: Same for line 364

bionic/core/task_execution.py
jqmp: > We create entries only when we need to process them.

I'm not sure this is true. We create an entry when we need to express some requirement on it, but that requirement might already be met. If the task state for that entry is already primed or cached, we don't need to actually put the entry into a `PENDING` state.

More abstractly: When we create a new entry, it has no requirements at all, so it's "completed" by definition, right?

jqmp:
It looks like the tests are failing because of a change in the behavior of Dask. We have a test that checks that we handle a particular exception it throws, but it seems to have stopped throwing that exception. I'll fix that in a separate PR before merging this one.

bionic/core/task_execution.py
namanjain: Duh, you are right. I was still thinking of it in terms of how we did it before. Also, you abstract explanation makes a lot of sense and convinced me easily!

PR #219, opened 2020-08-03
Fix test failure due to Dask update

jqmp:
Older versions of Dask (before 2.22.0) couldn't read multi-index
dataframes from Parquet, so we have a check to detect this and throw an
error if it happens, and a test to make sure this check is working.
However, the latest version of Dask seems to have fixed this problem,
which is good but breaks our test. I've updated the test to allow but
not require the exception to be thrown.

### DIFF
--- a/tests/test_flow/test_protocols.py
+++ b/tests/test_flow/test_protocols.py
@@ -361,7 +361,7 @@ def test_dask_dataframe_index_cols(builder):
         Delhi,India,Asia,26
         Shanghai,China,Asia,24
         Sao Paulo,Brazil,South America,21
-        Mumbai,India,21
+        Mumbai,India,Asia,21
         Mexico City,Mexico,North America,21
         Beijing,China,Asia,20
         Osaka,Japan,Asia,20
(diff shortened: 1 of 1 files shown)


tests/test_flow/test_protocols.py
```diff
@@ -381,8 +381,12 @@ def raw_df():
     def counts_df(raw_df):
         return raw_df.groupby(["continent", "country"]).size().to_frame("count")
 
-    with pytest.raises(UnsupportedSerializedValueError):
+    try:
         builder.build().get("counts_df")
```
namanjain, line 385: For Dask >= 2.22.0, can you add an assertion here that checks a value from `counts_df`?

jqmp: So, it turns out that Dask's new behavior here is [non-deterministic](https://github.com/dask/dask/issues/6484), so I'll have to add a somewhat vague check. Oh well.

namanjain: Till you get an answer from dask developers, I think it's okay to merge your original change (without any value checks) and the other PR that depends on it, this blocks a lot of your other work. You can open another PR later once you hear back from them.

jqmp: I agree. I modified the PR to add a minimal check in the test, but I don't think it's worth getting blocked any further on this.

BTW, based on the discussion so far in the linked thread, it sounds like it make more sense to use `fastparquet` for Dask DataFrames, but that can happen in a different PR.

PR #220, opened 2020-08-06
Introduce FunctionAttributes class

jqmp:
I moved the `code_fingerprint` and `changes_per_run` attributes into
a small data class, which `TaskState` can use instead of keeping a
reference to the original `Provider`. This makes no real difference now,
but it gives us the option of simplifying our task stripping code,
because `FunctionAttributes` is picklable but `Provider` is not. I think
this will be useful for fixing tuple persistence.

I also added a TODO for a further refactor that would move
`FunctionAttributes` onto the `Task` itself.

### DIFF
--- a/bionic/core/task_execution.py
+++ b/bionic/core/task_execution.py
@@ -307,13 +307,13 @@ class TaskState:
     intermediate state and the deriving logic.
     """
 
-    def __init__(self, task, dep_states, case_key, provider, entity_defs_by_dnode):
+    def __init__(self, task, dep_states, case_key, func_attrs, entity_defs_by_dnode):
         assert len(entity_defs_by_dnode) == len(task.keys)
 
         self.task = task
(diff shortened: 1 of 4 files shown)


jqmp:
Good point about the other TODO. Happy to discuss this more, but to me it seems pretty compelling to put this data on the Task object, for the following reasons:

1. It directly describes the Task's `compute` function, so it makes sense to put it as close to that function as possible.
2. Well, technically, it describes the user-provided function, which might be underneath several layers of wrapping. Ideally the function attributes would evolve as part of the wrapping process, so that's another reason to put it on the Task.

That said, if we accept that this data should ultimately end up on the Task, that still leaves the question of how it gets there. It's a little tricky because a decorator like `@version` can be appear at any point in the decorator chain, but it is intended to modify the function at the bottom. This might require some extra plumbing in the Provider interface.

PR #223, opened 2020-08-10
Use Code objects to analyze bytecode

namanjain:
Bionic looked at bytecode instructions to look for any changes in
functions to invalidate the cache. This change moves us away from that
approach in favor of analyzing the code objects directly instead.

This change is a no-op for the user. But in the future, we can extend
this to check references of the functions and any changes to class
objects.


bionic/code_analyzer.py
```diff
@@ -0,0 +1,134 @@
+"""
+This module implements the logic to analyze your code objects and
```
jqmp, line 2: Can we give a higher-level description of what's happening here? Why do we need to analyze code objects?

bionic/code_analyzer.py
```diff
@@ -0,0 +1,134 @@
...
+"""
+
+import inspect
+import json
+import warnings
+
+from .utils.misc import oneline
+
+
+DEFAULT_NAME = b"zA)T+ndu/Ygd%JH=eux)[F9:<>)zND]}u4x,c49tWH<yd5+#+4Xg{=J<[S5/'XG+"
```
jqmp, line 13: What is this?

bionic/code_analyzer.py
```diff
@@ -0,0 +1,134 @@
...
+    assert inspect.isroutine(func)
+
+    analyzer = CodeAnalyzer()
+    analyzer.analyze(func)
+    return analyzer.get_analyzed_bytes()
+
+
+class CodeAnalyzer:
+    """
+    This class contains the logic to analyze a code object. For now, it
```
jqmp, line 26: "Analyze" doesn't really say anything about what this class does. What is it useful for?

bionic/code_analyzer.py
```diff
@@ -0,0 +1,134 @@
...
+        if obj_id in self._id_to_hash:
+            return self._id_to_hash[obj_id]
+
+        # If the obj is already hashing, break the circular ref and return early.
+        if obj_id in self._ids_being_hashed:
+            return get_obj_name(obj, DEFAULT_NAME)
+        self._ids_being_hashed.add(obj_id)
+
+        # Otherwise, hash the object.
+        obj_hash = self._analyze(obj)
```
jqmp, line 53: What is happening here? Does `_analyze` return anything?

For that matter, should this function return something?

bionic/code_analyzer.py
```diff
@@ -0,0 +1,134 @@
...
+    def _update_bytes(self, data):
+        self._analyzed_bytearray += data
+
+    def _analyze(self, obj):
+        if isinstance(obj, bytes) or isinstance(obj, bytearray):
+            self._update_bytes(obj)
+
+        elif obj is None or isinstance(obj, (int, float, str, bool)):
+            json_bytes = json.dumps(obj).encode()
+            self._update_bytes(json_bytes)
```
jqmp, line 66: Does this mean that `b"1"` (a byte array) and `1` (an integer) will produce the same bytes? Will they also produce the same hash value?

bionic/code_analyzer.py
```diff
@@ -0,0 +1,134 @@
...
+class CodeAnalyzer:
+    """
+    This class contains the logic to analyze a code object. For now, it
+    only analyzes the function's bytecode and constants. But in the
+    future, it will be able to analyze its references, including class
+    types.
+    """
+
+    def __init__(self):
+        self._id_to_hash = {}
```
jqmp, line 33: Don't we usually use the convention `hashes_by_id`?

bionic/code_analyzer.py
```diff
@@ -0,0 +1,134 @@
...
+
+        elif inspect.isclass(obj):
+            # TODO: Add support to analyze class objects.
+            self.analyze(obj.__name__)
+
+        else:
+            message = oneline(
+                f"""
+                Found complex object {obj!r} of type {type(obj)!r} while analyzing
+                code for caching, probably as a global variable or a function
```
jqmp, line 95: Do we check globals and defaults right now?

bionic/code_analyzer.py
```diff
@@ -0,0 +1,134 @@
...
+def get_obj_id(obj):
+    if obj is None or isinstance(obj, (int, float, str, bool)):
+        return obj
+
+    return id(obj)
+
+
+def get_obj_name(obj, default):
+    """
+    Returns the name of an object if it contains one. Otherwise, returns the
```
jqmp, line 128: If there's no name, why not return `None` and let the caller worry about the default behavior? It seems like we only call this in one place anyway.

bionic/code_analyzer.py
```diff
@@ -0,0 +1,134 @@
...
+
+def analyze_func(func):
+    assert inspect.isroutine(func)
+
+    analyzer = CodeAnalyzer()
+    analyzer.analyze(func)
+    return analyzer.get_analyzed_bytes()
+
+
+class CodeAnalyzer:
```
jqmp, line 24: This class seems like a good candidate for tests.

bionic/code_analyzer.py
```diff
@@ -0,0 +1,134 @@
...
+        consts = [
+            const
+            for const in code.co_consts
+            if not isinstance(const, str) or not const.endswith(".<lambda>")
+        ]
+        self.analyze(consts)
+
+
+def get_obj_id(obj):
+    if obj is None or isinstance(obj, (int, float, str, bool)):
```
jqmp, line 120: Why the special case here? Doesn't this mean that two objects could share the same "id"?

Also, if we're going to return something that's different from `id`, we should have a different name.

bionic/code_analyzer.py
namanjain: I added more information in the module and changed the language. Lmk if you think I should add more information.

bionic/code_analyzer.py
namanjain: This method converts the objects into bytes and adds them to the result. I added some documentation to this method as well.

bionic/code_analyzer.py
namanjain: Ah, we don't. I have that on another branch and I probably didn't vet this warning before breaking this change from that branch. I'll remove the "probably as a global variable or a function default" bit and add it back when we start looking at them.

bionic/code_analyzer.py
namanjain: Streamlit calls it key, I guess we can call it the same.

bionic/code_analyzer.py
namanjain: You are right! I completely missed this one. I guess we can prefix bytes / bytearray with a placeholder value to differentiate it from integers.

The first thought that came into my mind after this was, what if it clashes with strings too? Thankfully, we use json for dumping strings first, which always adds a quote around the string. Wdyt?

bionic/code_analyzer.py
namanjain: Will do!

namanjain: Done!

bionic/code_analyzer.py
```diff
@@ -0,0 +1,203 @@
...
+"""
+
+import inspect
+import json
+import warnings
+
+from .utils.misc import oneline
+
+
+# We use object names for circular references when analyzing code. For
```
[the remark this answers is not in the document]
namanjain: That doesn't work since object names aren't detecting circular refs. But I get your point that this by itself isn't giving enough info. I'll change it to, "We analyze objects names instead of the objects when they are circularly referenced."

bionic/code_analyzer.py
```diff
@@ -0,0 +1,203 @@
...
+                    self.analyze(d)
+
+            self._analyze_code(obj.__code__)
+
+        elif inspect.iscode(obj):
+            self._analyze_code(obj)
+
+        elif inspect.isclass(obj):
+            # TODO: Add support to analyze class objects.
+            self.analyze(obj.__name__)
```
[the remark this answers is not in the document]
namanjain: It does, but none of our current code paths would hit this if block. We will encode class objects once we start going through the references.

Are you saying that till we actually support class objects, treat them as complex objects and emit the warning?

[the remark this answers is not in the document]
jqmp: Well, if we're not going to encounter this code path, we can always just put an `assert False` or something here. But if we're going to go ahead and implement some logic for this case, it should be correct logic, right?

bionic/code_analyzer.py
```diff
@@ -0,0 +1,203 @@
...
+        # Add a prefix for obj type so we can differentiate between
+        # different types that look the same, like a list and a tuple,
+        # or an int and bytes.
+        self._add_prefix_for_type(obj)
+
+        self._keys_being_hashed.add(obj_key)
+        self._analyze(obj)
+        self._keys_being_hashed.remove(obj_key)
+
+    def _add_prefix_for_type(self, obj):
```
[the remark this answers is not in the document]
namanjain: That's how I started it, but then changed it to this since I thought keeping the prefixes together would be easier (and we won't use the same prefix by mistake). But if you think it's cleaner when kept with their types during analysis, I can move it back.

bionic/code_analyzer.py
```diff
@@ -0,0 +1,203 @@
...
+        else:
+            prefix = b"Z"
+
+        self._analyzed_bytearray += prefix
+
+    def _add_results(self, obj, result):
+        self._results_by_key[get_obj_key(obj)] = result
+        self._analyzed_bytearray += result
+
+    def _analyze(self, obj):
```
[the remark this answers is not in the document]
namanjain: How about `analyze` and `_encode`?

bionic/code_analyzer.py
```diff
@@ -0,0 +1,203 @@
...
+        # or an int and bytes.
+        self._add_prefix_for_type(obj)
+
+        self._keys_being_hashed.add(obj_key)
+        self._analyze(obj)
+        self._keys_being_hashed.remove(obj_key)
+
+    def _add_prefix_for_type(self, obj):
+        if isinstance(obj, bytes):
+            prefix = b"A"
```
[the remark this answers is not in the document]
namanjain: I did that too for this one, but then I used `B` for both bytes and bool. But if we move them with their encoding logic, I think we should have some mnemonic relationship. Maybe I can change it to 2 bytes instead of 1. Wdyt?

bionic/code_analyzer.py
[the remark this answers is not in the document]
namanjain: I'll add assert False for now.

bionic/code_analyzer.py
namanjain: I feel so stupid to miss this. I can remove the special case. I thought it would be nice to skip encoding the same integer or string multiple times, but that should be quick anyway.

bionic/code_analyzer.py
```diff
@@ -0,0 +1,203 @@
...
+            self._analyze_code(obj)
+
+        elif inspect.isclass(obj):
+            # TODO: Add support to analyze class objects.
+            self.analyze(obj.__name__)
+
+        else:
+            message = oneline(
+                f"""
+                Found complex object {obj!r} of type {type(obj)!r}
```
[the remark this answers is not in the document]
namanjain: I don't think there are any constants that we don't handle, but I left it here just in case there is one that I missed. It will be easy for users to notice a problem and contact us.

tests/test_code_analyzer.py
```diff
@@ -0,0 +1,159 @@
...
+from bionic.code_analyzer import CodeAnalyzer
+
+
+def get_encoded_value(obj):
+    analyzer = CodeAnalyzer()
+    analyzer.analyze(obj)
+    return analyzer.get_analyzed_bytes()
+
+
+def test_bytes():
```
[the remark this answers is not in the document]
namanjain: There was only one partition that had different objects with the same encoding, so I made that particular one a different test. Otherwise, I combined all the tests into one and validated that all objects have different encoding.

bionic/persistence.py
```diff
@@ -926,7 +926,7 @@ def valid_filename_from_query(query):
     return query.dnode.to_descriptor().replace(" ", "-")
 
 
-CACHE_SCHEMA_VERSION = 7
+CACHE_SCHEMA_VERSION = 8
```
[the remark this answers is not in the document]
namanjain: I added release notes for cache changes but not sure where I should put this warning.

bionic/code_analyzer.py
```diff
@@ -0,0 +1,181 @@
...
+
+        # If the obj is already being hashed, break the circular ref by
+        # analyzing the depth of the value instead.
+        if obj_id in self._object_depth_by_id:
+            self._hash.update(CodeHashPrefix.CIRCULAR_REF.value)
+            obj = self._object_depth_by_id[obj_id]
+            obj_id = id(obj)
+
+        self._object_depth_by_id[obj_id] = depth
+        self._update(obj, depth)
```
jqmp, line 66: Wouldn't it be easier to increment the depth here, rather than every place where you call `_update_with_checks`?

bionic/code_analyzer.py
```diff
@@ -0,0 +1,181 @@
...
+        elif inspect.iscode(obj):
+            self._hash.update(CodeHashPrefix.CODE.value)
+            self._analyze_code(obj, depth)
+
+        else:
+            self._hash.update(CodeHashPrefix.DEFAULT.value)
+            message = oneline(
+                f"""
+                Found a constant {obj!r} of type {type(obj)!r} that
+                Bionic cannot hash. This is most likely a miss from the
```
jqmp, line 130: "a miss" and "the developers" are both vague. Let's just say "a bug in Bionic". Also, how should they contact us?

Finally, can we add a TODO to find out what the actual list of constant types is? I think this is a case where just doing the right thing (finding the list of constant types) is almost as easy as writing a clear explanation of the almost-right thing.

bionic/code_analyzer.py
```diff
@@ -0,0 +1,181 @@
...
+                developers. Please contact one of the code owners of
+                Bionic to fix this issue.
+                """
+            )
+            warnings.warn(message)
+
+    def _analyze_code(self, code, depth):
```
jqmp, line 137: Should this be `_hash_code` now?

bionic/code_analyzer.py
jqmp: Actually, do we even need to track `depth`, or is it the same as the size of `_object_depth_by_id`?

bionic/code_analyzer.py
```diff
@@ -0,0 +1,181 @@
...
+        self._update_with_checks(consts, depth + 1)
+
+
+class CodeHashPrefix(Enum):
+    """
+    Represents a unique value for each type that CodeHasher hashes that
+    is prefixed to avoid collision between same encoded values.
+
+    If you change the prefix of any type, or add a new prefix, it can
+    change the encoding of objects. Since Bionic use to encoding makes
```
jqmp, line 156: "Since Bionic use to encoding": are there some missing words here?

tests/test_code_analyzer.py
```diff
@@ -0,0 +1,112 @@
...
+
+def test_code_hasher():
+    def barray(value):
+        return bytearray(value, "utf8")
+
+    circular_dict = {"k1": "v1"}
+    circular_dict["k2"] = circular_dict
+
+    def f1():
+        v = 10
```
jqmp, line 15: Is this testing something different from just writing `return 10`?

tests/test_code_analyzer.py
```diff
@@ -0,0 +1,112 @@
...
+        f2,
+        # Simple functions
+        inc,
+        dec,
+        # Functions with different return types
+        r1,
+        r2,
+    ]
+
+    idx_by_hash_value = {}
```
jqmp, line 83: Do we ever add to this dict?

tests/test_code_analyzer.py
```diff
@@ -0,0 +1,112 @@
...
+        v = 20
+        return v
+
+    def inc(x):
+        return x + 1
+
+    def dec(x):
+        return x - 1
+
+    def r1():
```
jqmp, line 28: Isn't this the same as `f1`, but with a different name?

tests/test_code_analyzer.py
```diff
@@ -0,0 +1,112 @@
...
+        v = 10
+        return v
+
+    def r2():
+        v = "10"
+        return v
+
+    values = [
+        # Bytes
```
jqmp, line 37: It's not clear to me that these comments are adding much.

tests/test_code_analyzer.py
```diff
@@ -0,0 +1,112 @@
+import pytest
+import threading
+
+from bionic.code_analyzer import CodeHasher, CodeHashPrefix
+
+
+def test_code_hasher():
```
jqmp, line 7: How about the following?
- strings that are just numbers
- the same data in both string and bytes format
- the empty string (and bytes)
- infinity and NaN
- more complex nestings
- circular references at different depths

bionic/code_analyzer.py
```diff
@@ -0,0 +1,181 @@
...
+    DICT = b"AJ"
+    ROUTINE = b"AK"
+    CODE = b"AL"
+    CIRCULAR_REF = b"AM"
+    DEFAULT = b"ZZ"
+
+
+def get_size_in_bytes(byte_obj):
+    assert isinstance(byte_obj, (bytes, bytearray))
+    return str(sys.getsizeof(byte_obj)).encode()
```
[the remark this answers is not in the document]
namanjain: Oh I didn't realize we can do this. I searched this on stack overflow and this was the accepted answer. I tried len and it works, I'll replace it with len.

bionic/code_analyzer.py
```diff
@@ -0,0 +1,181 @@
...
+changes, we will make this the default behavior.
+"""
+
+import hashlib
+import inspect
+import json
+import sys
+import warnings
+
+from enum import Enum
```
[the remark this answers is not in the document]
namanjain: I thought `import x` and `from foo import bar` are different groups but I can merge them by removing the newline? Is there a particular ordering that you prefer?

bionic/code_analyzer.py
```diff
@@ -0,0 +1,181 @@
...
+
+    If you change the prefix of any type, or add a new prefix, it can
+    change the encoding of objects. Since Bionic use to encoding makes
+    caching backwards-incompatible.
+
+    Update CACHE_SCHEMA_VERSION to update cache scheme if you change
+    the encoding.
+    """
+
+    BYTES = b"AA"
```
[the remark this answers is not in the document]
namanjain: They don't have to be. In case we add more than 26 cases and start using `ZA`, `ZB`, and so on, I think it'd be nice to have the prefix's length consistent.

tests/test_code_analyzer.py
namanjain: Whoops, missed this!

tests/test_code_analyzer.py
namanjain: I realized that after adding to the `idx_by_hash_value` dict.

tests/test_code_analyzer.py
namanjain: I think I lost direct return method when refactoring test. Added it back to check that returning directly vs returning a variable are treated differently.

bionic/code_analyzer.py
namanjain: We need to track depth since we are hashing the depth of the reference, not the current obj. What I mean is, for a chain like `a -> b -> c -> b`, depth of `b` when hashing `c` should be 1 (depth of `a` is 0). If we use the size of `_object_depth_by_id`, depth will be 2, which is the depth of `c`, not `b`. 

In short, we need to track the depth. If we don't, then `a -> b -> c -> b` and `a -> b -> c -> a` will return the same hash.

bionic/code_analyzer.py
namanjain: I'm so sure I deleted a line while fixing indentation. Fixed.

bionic/code_analyzer.py
namanjain: Added a test for this.

bionic/code_analyzer.py
```diff
@@ -0,0 +1,181 @@
+"""
```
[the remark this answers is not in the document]
namanjain: I'll make this change at the end before merging.

bionic/code_analyzer.py
jqmp: Sorry, "track `depth`" was ambiguous. I agree we need to track the depth of every in-progress object. However, do we need to pass around the current depth? It seems like it should always be basically equal to the number of in-progress objects.

bionic/code_analyzer.py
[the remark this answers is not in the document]
jqmp: I think Google/StackOverflow might have given you something misleading. `len` will give you the length of any sequence, and for a `bytes` object that's just the number of bytes. `sys.getsizeof` will give you the *total size in bytes* of a *Python object* (if possible). If you call `sys.getsizeof` on a `bytes` object, you will get a much bigger number than `len`, because it includes the overhead of the array object in addition to the bytes themselves.

    > len(b'1234')
    4
    > sys.getsizeof(b'1234')
    37

bionic/code_hasher.py
```diff
@@ -0,0 +1,201 @@
...
+                type_prefix=TypePrefix.DICT, byte_obj=str(len(obj)).encode(),
+            )
+            for key, elem in obj.items():
+                self._update_with_checks(key)
+                self._update_with_checks(elem)
+
+        elif inspect.isroutine(obj):
+            self._write_to_hash_object(type_prefix=TypePrefix.ROUTINE)
+            if obj.__defaults__ is not None:
+                for d in obj.__defaults__:
```
jqmp, line 136: Hmm, this seems like a place where we should be explicitly encoding the number of defaults. Can we just call `_update_with_checks(obj.__defaults__)` instead?

bionic/code_hasher.py
```diff
@@ -0,0 +1,201 @@
...
+        ]
+        self._update_with_checks(consts)
+
+
+class TypePrefix(Enum):
+    """
+    Represents a unique value for each type that CodeHasher hashes that
+    is prefixed to avoid collision between same encoded values.
+
+    If you change the prefix of any type, or add a new prefix, it can
```
jqmp, line 176: Doesn't this warning really apply to the entire file?

tests/test_code_hasher.py
```diff
@@ -0,0 +1,126 @@
...
+import threading
+
+from bionic.code_hasher import CodeHasher, TypePrefix
+
+
+def test_code_hasher():
+    def barray(value):
+        return bytearray(value, "utf8")
+
+    circular_dict_1_a = {"k11": "v1"}
```
jqmp, line 11: super nit: It seems a little weird that 1a, 2a, 3a are all part of one dict, while 1b, 2b, and 3b are all part of another. I'd have expected it to be 1a, 1b, 1c and 2a, 2b, 2c.

tests/test_code_hasher.py
```diff
@@ -0,0 +1,126 @@
...
+        ), f"{values[idx]} and {values[idx_by_hash_value[hash_value]]} have the same hash"
+        idx_by_hash_value[hash_value] = idx
+
+
+def test_complex_type_warning():
+    val = threading.Lock()
+    with pytest.warns(
+        UserWarning, match="Found a constant",
+    ):
+        assert CodeHasher.hash(val) == CodeHasher.hash(TypePrefix.DEFAULT)
```
jqmp, line 112: Is the fact that we use `TypePrefix.DEFAULT` an important property that we should be testing?

tests/test_code_hasher.py
```diff
@@ -0,0 +1,126 @@
+import pytest
+import threading
+
+from bionic.code_hasher import CodeHasher, TypePrefix
+
+
+def test_code_hasher():
```
jqmp, line 7: Other test cases I might have missed before: Empty lists/tuples/dicts? Sets? Lambdas?

bionic/code_hasher.py
```diff
@@ -0,0 +1,201 @@
...
+class CodeHasher:
+    """
+    This class hashes code objects into values that uniquely identify
+    those objects. If two code objects have different behavior, they
+    should ideally have different hashes. However, for now, we only
+    examine the function bytecode and associated constants. We will add
+    more functionality like checking function references and any
+    changes to class objects.
+
+    Since it doesn't examines the references and class objects, it
```
jqmp, line 37: s/examines/examine/

Also, it's confusing to use "it" here without making it clear what the referent is.

bionic/code_hasher.py
```diff
@@ -0,0 +1,201 @@
...
+    LIST = b"AH"
+    TUPLE = b"AI"
+    DICT = b"AJ"
+    ROUTINE = b"AK"
+    CODE = b"AL"
+    CIRCULAR_REF = b"AM"
+    DEFAULT = b"ZZ"
+
+
+def get_size_in_bytes(byte_obj):
```
jqmp, line 199: The name `get_size_in_bytes` makes me think this will return an integer indicating the size of an object, measured in bytes. `get_size_as_bytes` might be clearer?

bionic/code_hasher.py
```diff
@@ -0,0 +1,201 @@
...
+
+    def _hash_code(self, code):
+        # TODO: Find references for the code and analyze references.
+        self._update_with_checks(code.co_code)
+
+        # Hash constants used in the code object, but ignore names of lambdas.
+        consts = [
+            const
+            for const in code.co_consts
+            if not isinstance(const, str) or not const.endswith(".<lambda>")
```
jqmp, line 166: Why do we ignore string constants?

tests/test_code_hasher.py
```diff
@@ -0,0 +1,126 @@
...
+        return v
+
+    def f3():
+        v = "10"
+        return v
+
+    def f4():
+        return "10"
+
+    def inc(x):
```
jqmp, line 40: Should we test functions that have a wider range of constant values?

tests/test_code_hasher.py
```diff
@@ -0,0 +1,126 @@
...
+def test_complex_type_warning():
+    val = threading.Lock()
+    with pytest.warns(
+        UserWarning, match="Found a constant",
+    ):
+        assert CodeHasher.hash(val) == CodeHasher.hash(TypePrefix.DEFAULT)
+
+
+def test_same_func_different_names():
+    def f1():
```
jqmp, line 116: Should we test whether docstrings affect the hash?

tests/test_code_hasher.py
jqmp: Also, should we be testing functions with default values? What about functions containing more complex constructions?

namanjain:
Addressed all unresolved comments. Thanks for sticking with this PR, I'll keep my future PRs smaller to make incremental progress.

namanjain:
> Looks great! I'm really excited about this approach.
> 
> BTW, if you add one more comment to this PR, we should hit 100. :)

💯

PR #226, opened 2020-08-12
Model entry requirements explicitly

jqmp:
I added an explicit system of requirements between `TaskRunnerEntry`s,
allowing us to indicate that one entry can't make progress until another
entry reaches a certain milestone. The requirements are represented by
an new `EntryRequirement` class, while progress is represented by an
`EntryLevel` enum.

This system replaces the old concepts of "requests" and `EntryBlockage`s. In
addition to probably being simpler, it also allows entries to have
requirements on non-dependency entries; this should be useful for
implementing followup tasks, or any situation where we want to compute
multiple entries remotely in one batch.

### DIFF
--- a/bionic/core/flow_execution.py
+++ b/bionic/core/flow_execution.py
@@ -3,12 +3,16 @@ This module contains the logic to execute tasks and their dependencies
 to completion.
 """
 
-from collections import defaultdict
 from concurrent.futures import wait, FIRST_COMPLETED
 
 import logging
 
-from .task_execution import EntryBlockage, EntryStage, TaskRunnerEntry
+from .task_execution import (
+    EntryLevel,
+    EntryStage,
+    EntryRequirement,
+    TaskRunnerEntry,
+)
 from ..utils.misc import SynchronizedSet
 
 
@@ -56,7 +60,9 @@ class TaskCompletionRunner:
         self._entries_by_task_key = {}
         self._pending_entries = []
         self._in_progress_entries = {}
-        self._blockage_lists_by_blocking_tk = defaultdict(list)
+        # TODO This is a set but _in_progress_entries is a dict. Could they both be
+        # sets?
+        self._blocked_entries = set()
 
     @property
     def _parallel_execution_enabled(self):
@@ -70,7 +76,9 @@ class TaskCompletionRunner:
             for state in states:
                 state.set_up_caching_flags(self._bootstrap)
                 entry = self._get_or_create_entry_for_state(state)
-                self._request_entry_be_cached(entry)
+                self._add_requirement(
+                    src_entry=None, dst_entry=entry, level=EntryLevel.CACHED
+                )
 
             while self._has_pending_entries():
                 entry = self._activate_next_pending_entry()
@@ -79,7 +87,7 @@ class TaskCompletionRunner:
 
             assert len(self._pending_entries) == 0
             assert len(self._in_progress_entries) == 0
-            assert len(self._get_all_blocked_entries()) == 0
+            assert len(self._blocked_entries) == 0
 
             results = {}
             for state in states:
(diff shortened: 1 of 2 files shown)


bionic/core/flow_execution.py
```diff
@@ -56,7 +60,9 @@ def __init__(self, bootstrap, flow_instance_uuid, task_key_logger):
         self._entries_by_task_key = {}
         self._pending_entries = []
         self._in_progress_entries = {}
-        self._blockage_lists_by_blocking_tk = defaultdict(list)
+        # TODO This is a set but _in_progress_entries is a dict. Could they both be
+        # sets?
+        self._blocked_entries = set()
```
namanjain, line 65: This might be my gap in python knowledge, but doesn't all elements of a set need to be hashable? Would entries always be hashable, even when the state or the cached values aren't?

bionic/core/flow_execution.py
```diff
@@ -104,39 +110,44 @@ def _process_active_entry(self, entry):
...
         self._set_up_entry_dependencies(entry)
         for dep_entry in entry.dep_entries:
-            self._request_entry_be_primed(dep_entry)
-        if not entry.all_dependencies_satisfy_requests:
+            self._add_requirement(entry, dep_entry, EntryLevel.PRIMED)
+        if not entry.all_outgoing_reqs_are_met:
             self._mark_entry_blocked(entry)
             return
 
+        # Now that ew have the dependency digests, we can initialize our task state.
```
namanjain, line 130: nit: ew -> we

bionic/core/flow_execution.py
```diff
@@ -104,39 +110,44 @@ def _process_active_entry(self, entry):
...
         entry.state.initialize(self._bootstrap, self._flow_instance_uuid)
 
-        # If this entry is persistable, let's see if we can load it from the cache ...
+        # If this entry is persistable, we may be able to load it from the persistent
+        # cache, which will immediately get us to the CACHED level.
         if entry.state.should_persist:
             entry.state.attempt_to_access_persistent_cached_values()
             if entry.is_cached:
```
namanjain, line 137: I think we should move away from this direct check and use `entry.all_incoming_reqs_are_met` instead.

In addition, we should make `is_cached` and `is_primed` properties private to `TaskRunnerEntry`.

bionic/core/flow_execution.py
```diff
@@ -104,39 +110,44 @@ def _process_active_entry(self, entry):
...
                 self._mark_entry_completed(entry)
                 return
+
+        # Otherwise, if it's not persistable, then initializing it will have gotten it
+        # to the PRIMED level. If that's all we needed, we're done.
         else:
-            # ... otherwise, if the digest was all we needed, calling initialize() was
-            # sufficient.
-            if not entry.is_requested_to_be_cached:
+            assert entry.level >= EntryLevel.PRIMED
```
namanjain, line 144: There is a similar assert at line 113 that checks level in a list of acceptable levels. I personally like this way better (using `>=` operator) over the list way (using the `in` operator on the list), but I'm fine with either approach. Can you pick one and make it consistent?

bionic/core/flow_execution.py
namanjain: After this change, this part becomes the same in both `if` and `else` blocks. Can you take them out of if else and put it after?

bionic/core/task_execution.py
```diff
@@ -96,6 +77,27 @@ def stage(self, stage):
...
+        if self.is_cached:
+            return EntryLevel.CACHED
+        elif self.is_primed:
+            return EntryLevel.PRIMED
+        else:
+            return EntryLevel.CREATED
+
+    @property
+    def required_level(self):
+        return max(req.level for req in self.incoming_reqs)
```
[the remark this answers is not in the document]
jqmp: Oh, great point! We should definitely do that.

bionic/core/task_execution.py
```diff
@@ -37,38 +38,21 @@ class TaskRunnerEntry:
...
-    with (a), for a task with non-persistable output, this milestone is reached as
-    soon as we compute its provenance; for a task with persistable output, it's
-    reached only when the task is computed and its output is cached.
-
-    Thus, the operational definition of "primed" depends on whether the task's output is
-    persistable or not. If so, "primed" is equivalent to "cached"; if not, "primed"
-    represents a preliminary stage of work which happens to be sufficient for many
-    applications.
+    other entries. These relationships mostly take the form of `EntryRequirement` objects,
+    which indicate that one entry do any work until another entry reaches a certain
```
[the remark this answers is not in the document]
jqmp: Oh yeah, I accidentally the word "can't".

bionic/core/flow_execution.py
```diff
@@ -70,7 +76,7 @@ def run(self, states):
             for state in states:
                 state.set_up_caching_flags(self._bootstrap)
                 entry = self._get_or_create_entry_for_state(state)
-                self._request_entry_be_cached(entry)
+                self._add_requirement(None, entry, EntryLevel.CACHED)
```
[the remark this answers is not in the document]
jqmp: Yep, makes sense.

bionic/core/task_execution.py
```diff
@@ -280,21 +251,66 @@ class EntryStage(Enum):
...
+    Represents a requirement from one entry to another.
+
+    A requirement indicates that one entry (`src_entry`) can't make any further progress
+    until another entry (`dst_entry`) has reached a certain level of progress (`level`).
     """
 
-    def __init__(self, blocked_entry, blocking_tks):
-        self.blocked_entry = blocked_entry
-        self._blocking_tks = set(blocking_tks)
+    src_entry = attr.ib()
```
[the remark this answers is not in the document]
jqmp: I originally used `requirer_` and `requiree_`, but the code felt busy and hard to read. I agree `src` and `dst` are not great names though. Maybe there are other options?

bionic/core/flow_execution.py
jqmp: By default, Python classes are hashable and the hash function is based on their `id`. (However, some classes, like mutable containers, explicitly disable their hash function.) For our purposes, I think the default behavior should be fine. I do prefer to have a clearer hash key, but it didn't seem worth the trouble of keying on `entry.state.task_keys[0]`. WDYT? Also, in the future, when every entry has exactly one task key, the balance might shift a little.

bionic/core/flow_execution.py
jqmp: That's a nice point. I kept it as `is_cached` because I think it has a clearer relationship to the previous line, but being able to move it out of the `if` block is very nice. And I agree it's better to use the new levels instead of `is_cached` and `is_primed`.

bionic/core/flow_execution.py
jqmp: I originally used `>=` in both places, but in the context of the first assert it seemed helpful to make it clear that there are only two levels of interest, since there are a couple places where we assume that being cached is the same as being done. However, your comment on `is_cached` removes one of them, so maybe if we add an `assert entry.all_incoming_reqs_are_complete` before the final `_mark_entry_completed`, that assumption isn't necessary anymore.

bionic/core/flow_execution.py
```diff
@@ -104,39 +110,44 @@ def _process_active_entry(self, entry):
 
         # In theory we could have an entry that doesn't require priming, but in practice
```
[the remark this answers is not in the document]
jqmp: I agree. Looks like we can change this back to `>= EntryLevel.PRIMED`, which should make the comment make sense again.

bionic/core/flow_execution.py
namanjain: TIL!

If it works, I think it's fine. I don't think we are ever going to disable hash function for entries.

bionic/core/task_execution.py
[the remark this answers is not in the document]
jqmp: I tried a few more variations (including `blocking_` and `blocked_`), but I think the common problem is that they all put the key distinction in the middle of a long string (`requestING_entry` vs `requestED_entry`), so it's hard to pick it out when reading the code. Let's talk more in our discussion today.

PR #228, opened 2020-08-14
Add a flow method for reloading in place

simonafk:


bionic/cache_api.py
```diff
@@ -18,13 +18,13 @@ class Cache:
     object. Use ``get_entries`` to iterate through the set of cache entries.
     """
 
-    def __init__(self, deriver):
+    def __init__(self, deriver_func):
```
jqmp, line 21: Rather than changing this interface, it might be easier to just re-create the Cache object when reloading happens.

bionic/flow.py
```diff
@@ -1557,6 +1557,19 @@ def ...
 
         return flow
 
+    def reloading_in_place(self):
```
jqmp, line 1560: Let's call this `reload`. Also, I'd suggest making this method the primary way of accessing this functionality -- that is, rather having this method's docstring refer to `reloading` for details, `reloading`'s should refer to this one. Similarly, we can make this method the first one to appear in the docs.

It might make sense to go as far as deprecating `reloading`, although we can always do that later.

tests/test_flow/test_reloading.py
```diff
@@ -0,0 +1,85 @@
...
+    '''documentation for {var_name}'''
+    return 2*x
+
+flow = builder.build()
+            """
+            )
+            tmp_file.flush()
+        # Needed to ensure that the file is loaded properly
+        # TODO: any way to avoid this?
+        time.sleep(1)
```
jqmp, line 34: As we discussed in the Hangout, it would be great to find out how to remove this `sleep`. If Naman or I can help with this, let us know.

tests/test_flow/test_reloading.py
```diff
@@ -0,0 +1,85 @@
...
+
+@pytest.fixture(scope="function")
+def make_tmp_flow_module():
+    # We want to be able to do import statement on the temporary module, so we
+    # create a temporary directory here and add it to sys.path.
+    tmp_dir = tempfile.TemporaryDirectory()
+    sys.path.insert(0, tmp_dir.name)
+
+    def _make_tmp_flow_module(var_name, var_value):
+        with open(f"{tmp_dir.name}/tmp_flow_module.py", "w") as tmp_file:
```
jqmp, line 16: If you just want to write some data to a file, you can do it like this:

    (tmp_dir / "tmp_flow_module.py").write_text("""...""")

bionic/flow.py
```diff
@@ -1557,6 +1557,19 @@ def ...
...
+        Attempts to reload all modules used directly by this flow and updates this flow instance.
+
+        Note that unlike other operations which are immutable, this method mutates the flow instance.
+
+        Please see ``Flow.reloading()`` for caveats.
+        """
+        flow = self.reloading()
+        self._state = flow._state
+        self._deriver = flow._deriver
+        self._uuid = flow._uuid
```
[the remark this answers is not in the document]
namanjain: Our documentation says that we compute an entity only once for a flow instance when it is decorated with `@changes_per_run`. Since we are not changing the instance and just mutating the current flow, I think we should preserve the old UUID and not compute it again.

I think testing it with `@changes_per_run` decorator is a good idea.

tests/test_flow/test_reloading.py
```diff
@@ -0,0 +1,85 @@
...
+def test_reloading(make_tmp_flow_module):
+    make_tmp_flow_module("y", 1)
+
+    from tmp_flow_module import flow
+
+    check_results(flow, {"x": 1, "y": 2})
+
+    make_tmp_flow_module("z", 2)
+    new_flow = flow.reloading()
+    check_results(new_flow, {"x": 2, "z": 4})
```
namanjain, line 63: Since we don't mutate the existing flow, can you also check that the results of `flow` didn't change after reloading it?

tests/test_flow/test_reloading.py
```diff
@@ -0,0 +1,85 @@
...
+
+@builder
+def {var_name}(x):
+    '''documentation for {var_name}'''
+    return 2*x
+
+flow = builder.build()
+            """
+            )
+            tmp_file.flush()
```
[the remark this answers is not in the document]
simonafk: Yeah I left that line by accident while trying various ways to avoid the 1 second sleep.

tests/test_flow/test_reloading.py
jqmp: Oh, I just noticed your `tmp_dir` is a `TemporaryDirectory` object, which might not support the `/` operator and so on. An easier way to get a temporary directory is to use the built-in `tmp_path` fixture, as in:

    def make_tmp_module(tmp_path):
        sys.path.insert(0, str(tmp_path))

        ...
        (tmp_path / "tmp_flow_module.py").write_text(...)
        ...

tests/test_flow/test_reloading.py
simonafk: Turns out that pytest rewrites certain modules to provide more details during failed assertions. https://docs.pytest.org/en/stable/assert.html. The rewritten modules are stored in a cache.

https://github.com/pytest-dev/pytest/blob/d69abff2c7de8bc65b7f1ef867dec5b5b9c564bd/src/_pytest/assertion/rewrite.py#L379
Pytest does a timestamp check on the module files with second precision to determine whether the cache is valid. This is why the sleep with 1 second works.

The caching mechanism can be disabled.

bionic/flow.py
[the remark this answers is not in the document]
simonafk: Added `@changes_per_run` tests

tests/test_flow/test_reloading.py
simonafk: Thanks!

bionic/flow.py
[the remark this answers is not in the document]
jqmp: Thinking about this more: although we are technically keeping the "same" flow instance, it's not obvious to me that we want to keep the same UUID. I believe most users will think of `@changes_per_run` as "recompute every time", but we have implemented it as "recompute once for each flow instance" because we want to keep each flow instance's in-memory cache consistent. However, in this case we are clearing the in-memory cache (by resetting `_deriver`) so this reason doesn't apply. I think resetting the UUID and allowing all the `@changes_per_run` entities to be recomputed may be a more intuitive behavior.

@namanjain WDYT?

tests/test_flow/test_reloading.py
simonafk: Updated the check methods so that it will remember and run previous checks if we are not reloading in place.

bionic/flow.py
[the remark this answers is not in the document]
namanjain: The way I think about it, users are probably used to the fact that unless the flow changes, changes_per_run entities are not recomputed. So I actually think that it would be confusing if we recompute them if the flow variable still points to the same object.

But now that I said it out loud, it would also be confusing that we'll recompute those entities if the input changes with reload (because the provenance will change). But we won't recompute them if the input doesn't change. To have consistent behavior, I'm now convinced that we should recompute them. 

I think it would be good to update changes per run documentation to reflect this. Wdyt?

tests/test_flow/test_reloading.py
```diff
@@ -0,0 +1,277 @@
...
+
+    return _write_flow
+
+
+def verify_flow_contents(flow, results_map):
+    assert {k: flow.get(k) for k in results_map.keys()} == results_map
+    assert {k: getattr(flow.get, k)() for k in results_map.keys()} == results_map
+
+
+@pytest.fixture(params=["reloading in place", "not reloading in place"])
```
namanjain, line 71: nit: You can also call these params `reload` and `reloading` to be concise.

tests/test_flow/test_reloading.py
```diff
@@ -0,0 +1,277 @@
...
+
+    return _check_flow
+
+
+def test_no_change(write_flow, check_flow):
+    write_flow(1)
+
+    from flow import flow
+
+    check_flow(flow, {"x": 1, "x_doubled": 2}, False)
```
jqmp, line 104: I like the thoroughness of these tests! On the other hand, `check_flow` is doing a lot of work and it's not clear at a glance what it's doing or what the arguments mean.

One way to handle this would be to have a stateful "tester" object with expressive methods, like
```
tester.expect_values(x=1, x_doubled=2)
tester.reload_flow()
tester.expect_values(x=1, x_doubled=2)
```

WDYT?

tests/test_flow/test_reloading.py
```diff
@@ -0,0 +1,277 @@
...
+@pytest.fixture(params=["reloading in place", "not reloading in place"])
+def reloading_in_place(request):
+    return request.param == "reloading in place"
+
+
+# If we are not reloading in place, we want to check not only the reloaded flow
+# but also the original flow to verify that the values have not changed.
+@pytest.fixture()
+def check_flow(reloading_in_place):
+    flow_to_expected = {}
```
namanjain, line 80: nit: for a map of x -> y, we use the convention `ys_by_x` instead of `x_to_y`. So this would be `expected_results_by_flow`.

tests/test_flow/test_reloading.py
```diff
@@ -0,0 +1,277 @@
...
+flow = builder.build()
+        """,
+        )
+
+    return _write_flow
+
+
+def verify_flow_contents(flow, results_map):
+    assert {k: flow.get(k) for k in results_map.keys()} == results_map
+    assert {k: getattr(flow.get, k)() for k in results_map.keys()} == results_map
```
jqmp, line 68: We could also test that `flow.all_entity_names()` matches the keys of `results_map`.

bionic/flow.py
```diff
@@ -1480,10 +1478,32 @@ def ...
...
+        doing the ``get()``.
 
-        This will reload the modules and use the most recent version of the
-        flow before doing the ``get()``.
+        Since the flow instance is kept, Bionic will avoid re-computation for
+        functions with the decorator ``@changes_per_run``
+        """
+        flow = self.reloading()
+
+        # Reusing the same UUID is needed for @changes_per_run to work
```
jqmp, line 1492: Super nit: the convention is to use complete sentences, including periods.

bionic/flow.py
```diff
@@ -1611,6 +1627,12 @@ def __init__(self):
 
         self.setting = SettingMethod()
 
+    def _set_state(self, state, uuid):
```
jqmp, line 1630: Since this does more than just set the `_state` attribute, maybe something like `_initialize` would be more accurate?

tests/test_flow/test_reloading.py
```diff
@@ -0,0 +1,277 @@
...
+            for f, r in flow_to_expected.items():
+                verify_flow_contents(f, r)
+
+    return _check_flow
+
+
+def test_no_change(write_flow, check_flow):
+    write_flow(1)
+
+    from flow import flow
```
jqmp, line 102: Rather than naming both things `flow`, maybe the first one could be `module` or something?

tests/test_flow/test_reloading.py
```diff
@@ -0,0 +1,277 @@
...
+
+    write_flow(2)
+    check_flow(flow, {"x": 2, "x_doubled": 4}, True)
+
+
+def test_change_entity_name(write_flow_file, check_flow):
+    def write_flow(z: int):
+        write_flow_file(
+            "flow",
+            f"""
```
jqmp, line 123: WDYT about using `dedent` to align this more cleanly?

docs/release-notes.rst
```diff
@@ -74,6 +74,12 @@ Upcoming Version (Not Yet Released)
    in the "stable" docs (corresponding to the last release) but will be visible in the
    "latest" docs (corresponding to the master branch).
 
+Reloading in place
```
jqmp, line 77: The section names here are standardized (see "Types of Changes" above), so this would fall under "New Features". Also, each individual item is a sentence that describes the change; so this could be something like "`Flow` now has a `reload` method, which acts like `reloading` but modifies the flow in place."

bionic/flow.py
```diff
@@ -1480,10 +1478,32 @@ def ...
...
+            flow.get('my_entity')
+
+        This will update the flow instance to use reloaded the modules before
+        doing the ``get()``.
 
-        This will reload the modules and use the most recent version of the
-        flow before doing the ``get()``.
+        Since the flow instance is kept, Bionic will avoid re-computation for
+        functions with the decorator ``@changes_per_run``
+        """
```
jqmp, line 1489: We should also mention that we return the flow.

docs/concepts.rst
```diff
@@ -955,6 +955,19 @@ a re-imported version of the flow.  (This is a fairly magical procedure -- in
 complicated cases, it may not be able to figure out how to do this.  In these
 cases it will try to throw an exception rather than fail silently.)
 
+You can also use the :meth:`Flow.reloading_in_place <bionic.Flow.reloading_in_place>`
```
jqmp, line 958: We'll want to update the documentation to reflect the new method name, and the fact that `reload` is now the primary way of doing a reload.

bionic/flow.py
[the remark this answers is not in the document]
jqmp: Copying my comment from Slack: since this is an interaction between `changes_per_run` and `reload`, and the latter is the weird special case, I think we should document it in `reload`.

tests/test_flow/test_reloading.py
simonafk: Renamed to `flow_module`

tests/test_flow/test_reloading.py
simonafk: Thanks for bringing it up. However, some of the tests do not check all the entities, only the relevant ones that change during reload.

tests/test_flow/test_reloading.py
```diff
@@ -0,0 +1,277 @@
...
+    # Ensure that the temporary module is not loaded by a future test
+    sys.path.remove(tmp_dir_name)
+
+    sys.dont_write_bytecode = dont_write_bytecode
+
+
+@pytest.fixture
+def write_flow_file(tmp_module_path):
+    modules = set()
+    counter = 0
```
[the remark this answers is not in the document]
simonafk: Good catch. Removed.

tests/test_flow/test_reloading.py
```diff
@@ -0,0 +1,277 @@
+import pytest
+
+import sys
+
+from bionic.exception import CodeVersioningError
+
+
+@pytest.fixture
+def tmp_module_path(tmp_path):
+    tmp_dir_name = str(tmp_path)
```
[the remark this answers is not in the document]
simonafk: `tmp_path` is unique per test invocation https://docs.pytest.org/en/stable/tmpdir.html, there should not be any collisions.

bionic/flow.py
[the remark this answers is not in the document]
simonafk: After our conversation, updated the method to use a new UUID. Entities marked with `changes_per_run` are recomputed.

bionic/flow.py
```diff
@@ -1480,10 +1478,27 @@ def ...
 
             from mymodule import flow
             ...
-            flow.reloading().get('my_entity')
+            flow.reload()
+            flow.get('my_entity')
+
+        This will update the flow instance to use reloaded the modules before
```
jqmp, line 1484: nit: s/reloaded the/the reloaded/

bionic/flow.py
```diff
@@ -1480,10 +1478,27 @@ def ...
...
+        doing the ``get()``.
 
-        This will reload the modules and use the most recent version of the
-        flow before doing the ``get()``.
+        Entities marked with ``@changes_per_run`` will be recomputed.
```
jqmp, line 1487: Oh, we probably should also mention that this resets the flow's in-memory cache. That will also give more context to this sentence, so we could say something like "Entities that have not been persisted to disk, or that have been marked with ``@changes_per_run``, will be recomputed."

bionic/flow.py
```diff
@@ -1570,9 +1585,10 @@ def __init__(self, state, _official=False):
                 "use one of the classmethod constructors"
             )
 
-        self._uuid = str(uuid4())
-        self._state = state
-        self._deriver = EntityDeriver(state, self._uuid)
+        uuid = str(uuid4())
+        deriver = EntityDeriver(state, uuid)
```
jqmp, line 1589: WDYT about creating the `EntityDeriver` and `Cache` inside `_initialize` so we don't have to pass them?

tests/test_flow/test_reload.py
```diff
@@ -0,0 +1,257 @@
...
+
+import sys
+from textwrap import dedent
+
+from bionic.exception import CodeVersioningError
+
+
+# Pytest rewrites modules and caches the compiled pyc file. The cache can
+# interfere with the tests below. Hence, we disable the cache here.
+@pytest.fixture(autouse=True)
```
jqmp, line 11: Not that it matters much, but I'm curious why you used `autouse` instead of just making `dont_write_bytecode` a dependency of `tmp_module_path`?

tests/test_flow/test_reload.py
```diff
@@ -0,0 +1,257 @@
...
+    yield
+    sys.dont_write_bytecode = dont_write_bytecode
+
+
+@pytest.fixture
+def tmp_module_path(tmp_path):
+    tmp_dir_name = str(tmp_path)
+    sys.path.insert(0, tmp_dir_name)
+
+    sys.dont_write_bytecode = True
```
jqmp, line 23: Is this intentional?

tests/test_flow/test_reload.py
```diff
@@ -0,0 +1,257 @@
...
+
+    yield tmp_path
+
+    # Ensure that the temporary module is not loaded by a future test.
+    sys.path.remove(tmp_dir_name)
+
+
+@pytest.fixture
+def write_flow_file(tmp_module_path):
```
jqmp, line 32: From the names `write_flow_file` and `write_flow_module`, I'd guess that they do related things, but one deals with *files* and the other deals with *modules*. However, the actual difference is that one deals with *arbitrary* modules and the other deals with *flow* modules. Maybe a better name for this function would be `write_module`?

docs/concepts.rst
```diff
@@ -947,13 +947,27 @@ be used:
...
-complicated cases, it may not be able to figure out how to do this.  In these
-cases it will try to throw an exception rather than fail silently.)
+This attempts to reload all modules associated with the flow, and then updates
+the flow instance to use the reloaded modules.  (This is a fairly magical
+procedure -- in complicated cases, it may not be able to figure out how to do
+this.  In these cases it will try to throw an exception rather than fail
+silently.) All future operations on the flow will reflect its updated status.
+
+You can also use the :meth:`Flow.reloading <bionic.Flow.reloading>` method to
+get a new copy of the flow that uses reloaded modules. The existing
```
jqmp, line 960: nit: the second sentence might be a little tighter as ", without modifying the original flow instance."

bionic/flow.py
jqmp: Also, now that we know that merged flows don't work, that might be a good thing to note in this docstring.

tests/test_flow/test_reloading.py
jqmp: Any thoughts on this?

tests/test_flow/test_reloading.py
jqmp: Oh, good point!

tests/test_flow/test_reloading.py
jqmp: I can see that concern, but I also think it will be clear in the contexts where it matters.

Also, "not reloading in place" is a bit ambiguous -- it suggests to me that no reloading is happening at all. WDYT about `reload_in_place` and `reloading_copy`?

tests/test_flow/test_reload.py
simonafk: Oops, this is meant to be part of `dont_write_bytecode()`

bionic/flow.py
simonafk: Hmm, we already have "has never been modified (i.e., isn't derived from another Flow)" as one of the safety requirements.

jqmp: Ah, I see what you mean. That text predates the merging functionality, but was intended to refer to flows derived by calling methods on an existing flow (like `flow.setting`); i.e., you can only call `reload()` on a flow that was returned by `builder.build()`. You're right that the text can also be reasonably interpreted to disallow merging as well, but technically if you merge in a flow before calling `build()`, then the *resulting* flow object "has never been modified". So it might be worthwhile to make this text slightly more explicit.

tests/test_flow/test_reloading.py
simonafk: Updated to use tester method

tests/test_flow/test_reload.py
simonafk: Moving the contents of this method to `write_module`, where it belongs.

bionic/flow.py
simonafk: It makes the method much simpler. `EntityDeriver` and `Cache` are re-created, but there is little cost in doing so.

tests/test_flow/test_reloading.py
simonafk: Actually, on further thought, my concern can be resolved by renaming the pytest method to `reload_method`. Renamed the params to use `reload` and `reloading`

tests/test_flow/test_reload.py
simonafk: Renamed module writing function to `write_module` and flow writing function to `write_flow_module`.

tests/test_flow/test_reload.py
```diff
@@ -0,0 +1,258 @@
...
+# and verify that it gives the same result after.
+@pytest.fixture
+def flow_tester(reload_method):
+    expected_results_by_flow = {}
+
+    def _verify_flow_contents(flow, results_map):
+        assert {k: flow.get(k) for k in results_map.keys()} == results_map
+        assert {k: getattr(flow.get, k)() for k in results_map.keys()} == results_map
+
+    class FlowTester:
```
jqmp, line 81: I think it's a little clearer and more idiomatic to put all your state and functions in the class, rather having a mixture of class state (`flow` and `expected_results`) and closure state (`reload_method` and `_verify_flow_contents`).

jqmp: Similarly, it's probably more straightforward to just have an `import_flow` method on the tester object than to require each test to construct this class. (Also, if you have a class called `FlowTester` and a fixture called `flow_tester`, I'd expect the latter to be an instance of the former.)

tests/test_flow/test_reload.py
```diff
@@ -0,0 +1,258 @@
...
+
+    modules = set()
+
+    def _write_module(name, contents):
+        (tmp_module_path / f"{name}.py").write_text(dedent(contents))
+        modules.add(name)
+
+    yield _write_module
+
+    # Invalidate caches
```
jqmp, line 36: convention nit: We use complete sentences, with periods, for comments.

tests/test_flow/test_reload.py
```diff
@@ -0,0 +1,258 @@
...
+        write_module(
+            "flow_module",
+            f"""
+                import bionic as bn
+
+                builder = bn.FlowBuilder('flow')
+                builder.set('core__versioning_mode', 'assist')
+
+                @builder
+                def total():
```
jqmp, line 203: Why is this called `total` and not, say, `x`?

tests/test_flow/test_reload.py
simonafk: The flow module needs to be written before `import flow_module` works. Hence the `FlowTester` class is constructed inside of the test.

bionic/flow.py
simonafk: Added safety requirements about flow merging.

tests/test_flow/test_reload.py
simonafk: Renamed

tests/test_flow/test_reload.py
jqmp: What I mean is, it might be clearer to leave the flow un-initialized and then initialize it like this:

```
def test_change_entity_value(write_flow_module, flow_tester):
    write_flow_module(1)
    tester.import_flow()
    tester.expect_values(x=1, x_doubled=2)
```

Normally I would agree that it's not great to require this sort of extra initialization, but in this case I think it makes the test both clearer and shorter. Up to you, though!

simonafk: I tried having `FlowTester` initialized with a pytest fixture. The `test_assist_versioning_mode` test fails in pycharm, but passes in pytest. This is very strange. I am wondering if it is related to howhow modules are loaded / cached differently in pycharm compared to pytest. One effect of having `FlowTester` be part of a pytest fixture is that it changes the scope of the `flow_module` module.

Issue #229, opened 2020-08-18
Caching fails when directory names contain spaces


jqmp:
Here's how to reproduce:

```
builder = bn.FlowBuilder('spaces')
builder.set('core__persistent_cache__global_dir', 'dir with spaces')

@builder
def x():
    return 7

flow = builder.build()

flow.get('x')
```

This returns an exception like this:

```
InvalidCacheStateError: Cached data may be in an invalid state; this should be impossible but could have resulted from either a bug or a change to the cached files. You should be able to repair the problem by removing all cached files under ('file:///Users/janek/Development/bionic/dir%20with%20spaces/spaces/inventory',).
```

I suspect this is a bug in the conversion between paths and URLs. It's also suspicious that the path is being reported as a tuple, but I'm not sure if that's a symptom or just a bug in our reporting. I will investigate.

PR #230, opened 2020-08-18
Unquote URI path values

namanjain:
This change fixes a bug in path to uri conversion that leads to quoted
uri value. Refer to #229 for an example of this issue.

Screen Shot 2020-08-18 at 11 54 04 AM

### DIFF
--- a/bionic/utils/urls.py
+++ b/bionic/utils/urls.py
@@ -4,7 +4,7 @@ Utilities for working with URLs.
 
 import os
 from pathlib import Path
-from urllib.parse import urlparse
+from urllib.parse import unquote, urlparse
 
 FILE_SCHEME = "file"
 GCS_SCHEME = "gs"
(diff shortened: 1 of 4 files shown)


bionic/utils/urls.py
```diff
@@ -36,7 +36,7 @@ def path_from_url(url):
 
 
 def url_from_path(path):
-    return Path(path).as_uri()
+    return unquote(Path(path).as_uri())
```
jqmp, line 39: If we unquote it here, then it's no longer a valid URL, right? Wouldn't it make more sense to unquote it in `path_from_url`?

namanjain: I think unquoted urls are technically valid urls too. But it makes sense to unquote it from `path_from_url` instead!

jqmp: I was going based on [this post](https://stackoverflow.com/questions/497908/is-a-url-allowed-to-contain-a-space), which states that spaces must be encoded in URLs.

BTW, I just remembered we also need a release note.

namanjain: TIL!

PR #232, opened 2020-08-20
Add code references

namanjain:
This is the next step towards hashing code references. The
implementation is copied from Streamlit, which is a reasonably popular
framework. But I added some basic tests to be sure that it works
correctly.

It also integrates the code references logic in Bionic's caching
strategy. It gets us a step closer to having a robust automatic caching
strategy.

Here is the list of things left after this change:

- hashing classes
- dedup references
- maybe return references that are stored in local variable
- caching individual object hashes for a CodeHasher run
- investigate if we can hash module or package versions
- verify that we hash all Python constant types
- version.suppress_bytecode_warnings TODO
- skip referenced code objects
- Only call getattr for modules TODO
- Throw an exception for unhandled bytecode instruction type


bionic/code_references.py
```diff
@@ -0,0 +1,124 @@
...
+            which means any changes to the reference won't invalidate
+            the cache.
+
+            This is most likely a bug in Bionic. Please raise a new
+            issue at https://github.com/square/bionic/issues to let us
+            know.
+            """
+            )
+            message = message + "\n" + str(e)
+            warnings.warn(message)
```
namanjain, line 122: I wasn't sure if I should throw an error or warning here, but I ended up using a warning. I think it's really bad if a user can't use the flow at all. Maybe we should give more information like how to delete / invalidate the cache entries manually when they see this warning. That way, the users won't be completely blocked. Wdyt? 

FWIW, Streamlit throws an error here.

jqmp: It sounds like we're going to keep iterating through the remaining opcodes. Does that make sense? Will our state still be valid?

bionic/code_references.py
```diff
@@ -0,0 +1,124 @@
...
+        nonlocal tos
+        if tos is not None:
+            # Hash tos so we support reading multiple objects
+            refs.append(tos)
+        tos = t
+
+    # Our goal is to find referenced objects. The problem is that co_names
+    # does not have full qualified names in it. So if you access `foo.bar`,
+    # co_names has `foo` and `bar` in it but it doesn't tell us that the
+    # code reads `bar` of `foo`. We are going over the bytecode to resolve
```
jqmp, line 63: Do we have a test case that illustrates this `foo.bar` case and how it's useful? I see the example with `[MyClass, "log_val"]`, but it's not clear to me why we care about `"log_val"`, since presumably we can get that just from hashing the bytecode.

bionic/code_references.py
namanjain: Now that I had some more time to think about it, I think I should change the catch block and add the tos to references if it contains one.

bionic/code_references.py
```diff
@@ -0,0 +1,124 @@
...
+        )
+
+    varnames = {}
+    if inspect.ismethod(func):
+        varnames = {"self": func.__self__}
+
+    return CodeContext(globals=func.__globals__, cells=cells, varnames=varnames)
+
+
+def get_referenced_objects(code, context):
```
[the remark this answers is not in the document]
jqmp: 1. What is the definition of "major"? And why?
2. If we don't catch every kind of reference, why not?
3. How do we know that it will catch every major reference? I'd like to be more confident than "I tried as many tests as I could think of and they passed". We should be able to read the code and be convinced that it is correct. What can you tell the reader that would convince them?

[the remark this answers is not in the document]
namanjain: I added docstring to this method. But I think it's really hard to justify `3.` without explaining how python bytecode instructions work.

bionic/code_references.py
namanjain: After thinking about it a bit more (and talking on slack), I changed it to an error (like Streamlit).

bionic/code_references.py
```diff
@@ -0,0 +1,159 @@
...
+            message = oneline(
+                f"""
+            Bionic found a code reference in file ${code.co_filename}
+            at line ${lineno} that it cannot hash when hashing
+            ${code.co_name}.This should be impossible and is most
+            likely a bug in Bionic. Please raise a new issue at
+            https://github.com/square/bionic/issues to let us know.
+            """
+            )
+            raise CodeVersioningError(message) from e
```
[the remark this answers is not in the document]
namanjain: Like we discussed a while back, I think we should give users the option to ignore bytecode of an entity. Since we don't have that yet, I haven't added anything related to it in the error message. But I can add a code comment (or a TODO) so we don't forget about it.

bionic/code_references.py
```diff
@@ -0,0 +1,159 @@
...
+                    set_tos(context.globals[op.argval])
+                else:
+                    set_tos(op.argval)
+            elif op.opname in ["LOAD_DEREF", "LOAD_CLOSURE"]:
+                set_tos(context.cells[op.argval])
+            elif op.opname == "IMPORT_NAME":
+                try:
+                    set_tos(importlib.import_module(op.argval))
+                except ImportError:
+                    set_tos(op.argval)
```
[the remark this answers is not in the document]
namanjain: I think this is just defensive coding. I'll remove it from our implementation.

tests/test_code_references.py
```diff
@@ -0,0 +1,164 @@
+import pytest
```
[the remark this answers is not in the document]
namanjain: Streamlit doesn't have any unit tests for finding code references, they only have integration tests. But looking at their [integration tests](https://github.com/streamlit/streamlit/blob/develop/lib/tests/streamlit/hashing_test.py#L649), I think I'm covering all of the cases they do.

I'll add all the examples in the appendix of the Code References doc here to have more complex cases.

bionic/code_references.py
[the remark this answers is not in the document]
namanjain: I added some more details in the function docstring.

bionic/code_references.py
```diff
@@ -0,0 +1,124 @@
...
+
+def get_code_context(func) -> CodeContext:
+    code = func.__code__
+
+    # Mapping from variable name to the value if we can resolve it.
+    # Otherwise map to the name.
+    cells = {}
+
+    for var in code.co_cellvars:
+        cells[var] = var  # Instead of value, we use the name.
```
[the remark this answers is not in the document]
namanjain: Streamlit never explained it anywhere, but I left my understanding of this as a code comment.

bionic/code_references.py
```diff
@@ -0,0 +1,191 @@
...
+    Attributes
+    ----------
+    globals: dict
+        A reference to the dictionary that holds the function’s global
+        variables — the global namespace of the module in which the
+        function was defined.
+    cells: dict
+        A dictionary that tracks all free variables by name and cell
+        variable names by themselves.
+    varnames: dict from TaskKey to Task
```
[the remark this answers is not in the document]
namanjain: No... I don't know how that ended up there...

tests/test_code_references.py
```diff
@@ -0,0 +1,203 @@
...
+            cell_var = "cell_val"
+            import logging
+
+            logging.log(global_val, free_var, local_val, cell_var, func_def_var)
+            builder = FlowBuilder("x")
+            # more business logic
+            builder.build()
+
+            def inner_x():
+                import logging
```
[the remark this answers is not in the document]
namanjain: Yup, that doesn't matter. The purpose of the inner function was to test for cell variable. The `inner_x` function is also a constant for `enclosing_x` function which will be hashed separately, so any references inside `inner_x` will be taken care of then.

I think this would become apparent when we integrate code referencing logic with our hashing logic and tests for them.

tests/test_code_references.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+def test_multiple_references():
+    def get_val(dict, key):
+        return dict[key]
+
+    def enclosing_x():
+        free_var = "free_val"
+
+        def x(func_def_var="func_def_val"):
```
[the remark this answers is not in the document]
namanjain: We will go through bytecode instructions for every line of every control flow. I'll add a test for loops and if else. 👍🏽

bionic/code_references.py
```diff
@@ -0,0 +1,191 @@
...
+
+            if op.opname in ["LOAD_GLOBAL", "LOAD_NAME"]:
+                if op.argval in context.globals:
+                    set_tos(context.globals[op.argval])
+                else:
+                    set_tos(op.argval)
+            elif op.opname in ["LOAD_DEREF", "LOAD_CLOSURE"]:
+                set_tos(context.cells[op.argval])
+            elif op.opname == "IMPORT_NAME":
+                # Streamlit catches ImportError and returns op.argval
```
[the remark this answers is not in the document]
namanjain: For posterity, we discussed this off-GitHub. Importing a module like this only happens when someone uses import statements inside functions. If a user is importing modules inside a function, it's mostly because they don't want to import the module till function execution time.

In Bionic, we will change this and not execute import statements either. Instead, we will suggest users to make the import statement global if possible. We will also emit a warning and give users an option to either suppress the warning or turn off bytecode analysis altogether.

bionic/code_references.py
```diff
@@ -0,0 +1,191 @@
...
+                # as the reference instead. If we see errors with
+                # importing modules, consider adding that check.
+                set_tos(importlib.import_module(op.argval))
+            elif op.opname in ["LOAD_METHOD", "LOAD_ATTR", "IMPORT_FROM"]:
+                if tos is None:
+                    refs.append(op.argval)
+                elif isinstance(tos, str):
+                    tos += "." + op.argval
+                else:
+                    tos = getattr(tos, op.argval)
```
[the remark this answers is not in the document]
namanjain: I don't think it's possible for attribute to not exist. Python should throw an error during compile time in that case.

But yes, if the attribute is a property, it will run the property method code. I didn't find a way to detect whether a variable is a property.

bionic/code_references.py
```diff
@@ -0,0 +1,191 @@
...
+            # previous start_line (if any). This way when there's an exception we at
+            # least can point users somewhat near the line where the error stems from.
+            if op.starts_line is not None:
+                lineno = op.starts_line
+
+            if op.opname in ["LOAD_GLOBAL", "LOAD_NAME"]:
+                if op.argval in context.globals:
+                    set_tos(context.globals[op.argval])
+                else:
+                    set_tos(op.argval)
```
jqmp, line 147: Thinking about this more: here `op.argval` is just a string, right? So if later we have a `LOAD_ATTR` instruction, we will call `getattr` on this string? Is that what we want?

jqmp: Also, it might be helpful to have a comment explaining how this case arises. If we see a `LOAD_NAME` and it's not in the global names, what other places could it be in?

bionic/code_references.py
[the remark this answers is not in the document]
namanjain: I was very off mark about python compile time checks. For now, I'm going to gate this behind `hasattr` check.

We also discussed this off-GitHub but we might end up calling `getattr` only for modules. But that would require some more thought to make sure we don't miss anything. I'll leave that as a TODO.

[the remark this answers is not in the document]
namanjain: I'm gonna leave a TODO for properties too.

bionic/code_references.py
namanjain: If we get `LOAD_ATTR` when `TOS` is string, we concatenate the TOS with the attribute name.

From my understanding of Python and bytecode, I think this can only happen if the variable does not exist. If you look at the test `test_function_references`, you'll see that the last example in that test return a `"func_does_not_exist"` string because the function does not exist and cannot be found.

bionic/code_references.py
```diff
@@ -0,0 +1,214 @@
...
+    the returned reference would be ``"call"`` instead of ``MyClass.call``.
+
+    .. code-block:: python
+
+        def x():
+            my_cls = get_my_class("MyClass") # Returns class MyClass
+            my_cls.call()
+
+
+    On a higher level, for Smart Caching, this leaves some changes that Bionic
```
jqmp, line 109: I don't think "Smart Caching" is referenced anywhere else in our codebase. I don't think you need to go into a lot of detail here, but can we write this in a more standalone way?

bionic/code_references.py
```diff
@@ -0,0 +1,214 @@
...
+            # least can point users somewhat near the line where the error stems from.
+            if op.starts_line is not None:
+                lineno = op.starts_line
+
+            if op.opname in ["LOAD_GLOBAL", "LOAD_NAME"]:
+                if op.argval in context.globals:
+                    set_tos(context.globals[op.argval])
+                else:
+                    # This can happen if the variable does not exist. If we
+                    # cannot find the variable, we return it's name instead.
```
jqmp, line 151: nit: s/it's/its/

bionic/code_references.py
```diff
@@ -0,0 +1,214 @@
...
+            # previous start_line (if any). This way when there's an exception we at
+            # least can point users somewhat near the line where the error stems from.
+            if op.starts_line is not None:
+                lineno = op.starts_line
+
+            if op.opname in ["LOAD_GLOBAL", "LOAD_NAME"]:
+                if op.argval in context.globals:
+                    set_tos(context.globals[op.argval])
+                else:
+                    # This can happen if the variable does not exist. If we
```
jqmp, line 150: What does it mean for the variable to not exist? Are you saying this can only happen if the code is referencing an invalid name? It looks to me like `LOAD_NAME` is intended to look up more than just globals, so presumably this could be a valid reference even if it's not in `context.globals`?

bionic/code_references.py
```diff
@@ -0,0 +1,214 @@
...
+                # This instruction only appears if the code object imports a
+                # module using the import statement. If a user is importing
+                # modules inside a function, they probably don't want to import
+                # the module until the function execution time.
+                message = oneline(
+                    f"""
+                Bionic found a code reference in file ${code.co_filename} at line
+                ${lineno} that imports ${op.argval} module when hashing
+                ${code.co_name}. Changes inside the module will likely not be
+                detected. To detect changes inside the module, import it globally
```
jqmp, line 165: What if they can't import it globally for some reason? What should they do to avoid this warning?

bionic/code_references.py
```diff
@@ -0,0 +1,214 @@
...
+                set_tos(None)
+            elif op.opname in ["LOAD_METHOD", "LOAD_ATTR"]:
+                if isinstance(tos, str):
+                    tos += "." + op.argval
+                # TODO: Consider calling getattr only when TOS is a module.
+                # Doing so has risk of missing cases, and any missing case would
+                # be hard to detect for users. Before making this change,
+                # document all the cases that we won't be catching so that we are
+                # aware of them.
+                # TODO: Due to the same reason as why we don't import modules, we
```
jqmp, line 179: It sounds like we intend to complete this TODO before releasing this code -- is that right? If so, can we document this issue in a more prominent way to make sure we remember to fix it?

bionic/code_references.py
```diff
@@ -0,0 +1,214 @@
...
+                # should not call `getattr` on properties. We should either add a
+                # separate check for properties, or implement the above TODO to
+                # fix this.
+                elif hasattr(tos, op.argval):
+                    tos = getattr(tos, op.argval)
+                else:
+                    refs.append(op.argval)
+            elif op.opname == "DELETE_FAST" and tos:
+                del context.varnames[op.argval]
```
jqmp, line 188: I hadn't noticed before that we're mutating the context object here. Should we make this clear in the docstring? Would we ever continue using the context object after calling this function?

bionic/code_references.py
```diff
@@ -0,0 +1,214 @@
...
+            message = oneline(
+                f"""
+            Bionic found a code reference in file ${code.co_filename}
+            at line ${lineno} that it cannot hash when hashing
+            ${code.co_name}. This should be impossible and is most
+            likely a bug in Bionic. Please raise a new issue at
+            https://github.com/square/bionic/issues to let us know.
+            """
+            )
+            raise CodeVersioningError(message) from e
```
jqmp, line 212: Thinking about this more: should this be a `CodeVersioningError`? If this a bug in Bionic, maybe it should be an `AssertionError`?

bionic/code_references.py
namanjain: You're right, it looks `LOAD_NAME` also looks at the local variables in the frame object (which we don't have). I'll update this.

bionic/code_references.py
namanjain: We discussed this sometime back (before I started working on the integration logic). I think we decided to not delete variables, so I'm going to remove `DELETE_FAST` branch from here.

namanjain: I think we should `DELETE_FAST` anyway, but just to be sure that there is no side effect of mutating the context object during this analysis phase, I'll just make a copy of `code_context` at the beginning of this method.

tests/test_flow/test_persistence.py
```diff
@@ -747,6 +748,139 @@ def y():  # noqa: F811
...
+    builder.set("core__versioning_mode", "auto")
+
+    complex_number = complex(1.0, 0.0)
+
+    @builder
+    def x():
+        assert complex_number is not None
+        return 1
+
+    with pytest.warns(None):
```
jqmp, line 845: What is `warns(None)` doing? My understanding is that this just records the warnings for later use, but we're not using them, are we?

bionic/code_references.py
```diff
@@ -0,0 +1,248 @@
...
+        A dictionary that is used to track local variables by name.
+    can_find_references: bool, default True
+        Determines whether we can find references for the code object
+        using this reference.
+    """
+
+    globals = attr.ib(default={})
+    cells = attr.ib(default={})
+    varnames = attr.ib(default={})
+    can_find_references = attr.ib(default=True)
```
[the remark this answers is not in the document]
namanjain: I don't think there is anything useful we can do with the context when it is false. I was leaning towards keeping it None and asserting the value but this looked a bit cleaner when I tried. I can move back to None and assert though. Wdyt?

tests/test_flow/test_persistence.py
namanjain: Ohhh, I thought it asserts that the warning is `None`. But you are right, it records the warnings. I'll check that there are no recorded warnings with `pytest.warns(None)`.

tests/test_code_references.py
```diff
@@ -200,3 +199,118 @@ def x(my_class):
...
+
+def test_complex_function():
+    import inspect
+    import warnings
+    from bionic.code_hasher import CodeHasher, TypePrefix
+    from bionic.utils.reload import is_internal_file
+
+    hasher = CodeHasher(False)
+    # TODO: This makes me realize that maybe we should dedup the results.
+    assert get_references(hasher._ingest) == [
```
jqmp, line 212: If this test finds a bug in `CodeHasher`, then fixing the bug will also change the test, right?

namanjain: Changing `CodeHasher` code will likely change the returned references in this test too. Even if we don't add any new references, we will probably end up with a different order of the references. 

I tried a few methods from our test code as well (like the random tests), but it was referencing internal pytest methods and it was easier to use a complex method from Bionic code instead. I can always use a different method (replacing this or in addition to this) if you have something in mind.

Instead of just copying the returned references, I wrote all the references myself and then checked that the returned references were the same. This way I was able to make sure that we didn't miss a reference. I did add a TODO in this test which points that we might miss references from a conditional block statement. That's on my top of list of things to follow up on.

namanjain: Oh and if your concern is that the test is testing it's own source code, you shouldn't be worried about that. We are testing the logic for `get_code_references` by calling it on `CodeHasher`, which is separate from `get_code_references`.

bionic/code_references.py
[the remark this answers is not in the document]
jqmp: It seems to me that there are two separate issues:

1. If we don't have a valid code context, should we represent that with `None` or with a special flag on the context object? I think `None` is preferable to having a flag that means "don't use this object for anything".
2. If we don't have a valid code context when processing a code object, should we throw an exception or just skip analyzing the code? I'm not sure what the right answer is. What do you think?

bionic/code_references.py
```diff
@@ -0,0 +1,214 @@
...
+    varnames = {}
+    if inspect.ismethod(func):
+        varnames = {"self": func.__self__}
+
+    return CodeContext(globals=func.__globals__, cells=cells, varnames=varnames)
+
+
+def get_referenced_objects(code, context):
+    """
+    Returns referenced objects for a code object and name of the references when
```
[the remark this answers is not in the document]
namanjain: Sorry, forgot to reply to here. The example in the docstring will only return two references: 
* the function get_my_class, and
* the string `"call"`

The string `"MyClass"` will be a constant and hashed by `CodeHasher` from `func.__code__.co_consts` variable.

bionic/code_references.py
namanjain: I'm keeping track of things left to do separately. This TODO is just a reminder in case it slips through the cracks (which it shouldn't).

bionic/code_references.py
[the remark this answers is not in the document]
namanjain: 1. I was leaning towards `None` too but this seemed a bit cleaner in case we want to ever change it later. I can go back to `None`.

2. I'm on the fence for this one too. I went from throwing an exception to ignoring it, but we won't ever get to this point unless someone passes a code object to the hasher directly (which shouldn't happen) or defines a code object themselves (do people even do this?). I can be easily convinced otherwise, it is totally reasonable to throw an exception as a guard against any future bugs. But if you don't feel strongly, I think I'll keep ignoring it in case anyone ever defines a code object by themself.

tests/test_code_references.py
jqmp: Oh right, I was wrong to think that changing CodeHasher would affect this test.

That said, this does seem to have the potential to introduce a lot of annoying busywork whenever we change the code of `CodeHasher`. I don't think this needs to block this PR though.

bionic/datatypes.py
```diff
@@ -295,13 +298,30 @@ def str_from_version_value(value):
 class CodeVersion:
     """
     Contains the user-designated version of a piece of code, consisting of a
-    major and a minor version string.  The convention is that changing the
-    major version indicates a functional change, while changing the minor
-    version indicates a nonfunctional change.
+    major and a minor version string, and a boolean that indicates whether it
+    includes the bytecode.  The convention is that changing the major version
+    indicates a functional change, while changing the minor version indicates a
+    nonfunctional change.  includes_bytecode determines whether the bytecode is
```
jqmp, line 304: I would phrase the last sentence as, "If `include_bytecode` is True, then the major version is understood to implicitly include the bytecode of the code as well."

docs/concepts.rst
```diff
@@ -316,17 +316,44 @@ file as long as the ``major`` version matches.
...
-about versioning.
 
-However, if you do want to completely automate the versioning process, you
-can set Bionic to a "fully automatic" mode:
+Be aware that Bionic can't detect every change that can affect your code's
+behavior. Bionic will inspect the bytecode of each function, as well as any
+global variables and other functions it references [#bytecode_hash]_. However, it
+won’t detect changes of the following types:
+
+1. **Installed Modules**: Bionic won’t detect changes in any built-in modules or
```
jqmp, line 325: I guess this should say "Built-In or Installed Modules". Also, "built-in modules or installed modules" should probably be "built-in or installed modules or functions".

docs/concepts.rst
```diff
@@ -316,17 +316,44 @@ file as long as the ``major`` version matches.
...
+   installed modules (modules accessed through Python’s
+   `installation paths <https://docs.python.org/3/library/sysconfig.html#installation-paths>`_.
```
jqmp, line 327: We're missing a close parenthesis here.

bionic/code_references.py
```diff
@@ -0,0 +1,245 @@
...
+        varnames = {"self": func.__self__}
+
+    return CodeContext(globals=func.__globals__, cells=cells, varnames=varnames)
+
+
+def get_referenced_objects(code, context):
+    """
+    Attempts to return all objects referenced externally by a code object. In
+    some cases, these objects are
+    - replaced with strings: when we can't find a variable or when the reference
```
jqmp, line 93: Just realized this: doesn't returning strings as placeholders also introduce the possibility of collisions? Like if we change this code

    x = call_func() # Suppose we can't following this reference

to 

    x = "call_func"

Can a situation like this produce the same hash?

WDYT about wrapping these placeholder strings in a special class? This might also solve the issue of using the `tos` variable for both reference values and reference names.

jqmp (commented):
I took another pass through the code. Most of the remaining changes seem pretty small. However, it seems like there are still several followup changes we plan to make that will change the hashes, which means that if we make those changes after releasing this code then we'll have another cache-breaking release. There are also some unaddressed edge cases that I'm not sure we'd want to release without fixing.

Normally I think it would be good practice to just fix all these things before merging, but I know this PR has been outstanding for a while. Could you just add a list of all the stuff that should be fixed before release and add it as a comment or something?

bionic/code_references.py
```diff
@@ -0,0 +1,245 @@
...
+        variables — the global namespace of the module in which the
+        function was defined.
+    cells: dict
+        A dictionary that tracks all free variables by name and cell
+        variable names by themselves.
+    varnames: dict
+        A dictionary that is used to track local variables by name.
+    """
+
+    globals = attr.ib(default={})
```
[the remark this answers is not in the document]
namanjain: We don't need these now. I'll revert the changes.

bionic/code_references.py
```diff
@@ -0,0 +1,245 @@
...
+    - replaced with strings: when we can't find a variable or when the reference
+      is an attribute of the result of a function call, or
+    - omitted: when the reference is accessed using exec or is the result of a
+      function call.
+
+    An externally referenced object is any object used by a piece of code but not
+    defined in that code. These objects can include classes, functions, modules,
+    or any other variables. They can be referenced from various scopes: global,
+    local, "free" (when the variable is used in a code block, but is not defined
+    there or is not global), etc.; or as attributes of other
```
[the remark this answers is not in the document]
namanjain: Right, it should be "and". Good point about local, I'll remove that.

bionic/code_references.py
namanjain: I don't think an example like that will result in possible collisions because we also hash the `func.__code__.co_code` variable which is the bytecode hash of the function itself. Those two functions will have different `co_code`.

We can hash those strings in a separate class, but we will have to wait till we hash classes for it to work. I'll leave a TODO to change it once we have support for hashing classes.

bionic/code_references.py
```diff
@@ -0,0 +1,245 @@
...
+                # should not call `getattr` on properties. We should either add a
+                # separate check for properties, or implement the above TODO to
+                # fix this.
+                elif hasattr(tos, op.argval):
+                    tos = getattr(tos, op.argval)
+                else:
+                    refs.append(op.argval)
+            elif op.opname == "STORE_FAST" and tos:
+                context.varnames[op.argval] = tos
+                tos = None
```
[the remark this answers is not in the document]
namanjain: I think this is because it stores the `tos` in `context.varnames`. But I agree that maybe we should use `set_tos(None)` and return the current `tos` as a reference. This is exactly what I meant to do from this [TODO](https://github.com/square/bionic/pull/232/files#diff-9e61581018162d3520e8e837df29a47bb3bea1afc6e11643b25b66d4e67d4a03R254).

bionic/code_references.py
```diff
@@ -0,0 +1,245 @@
...
+
+    for op in dis.get_instructions(code):
+        try:
+            # Sometimes starts_line is None, in which case let's just remember the
+            # previous start_line (if any). This way when there's an exception we at
+            # least can point users somewhat near the line where the error stems from.
+            if op.starts_line is not None:
+                lineno = op.starts_line
+
+            if op.opname in ["LOAD_GLOBAL", "LOAD_NAME"]:
```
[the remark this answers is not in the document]
namanjain: Yeah, we can have a list of known instructions at the end and throw an error if we encounter an unknown instruction. This is becoming a theme now, but I'll leave another TODO.

bionic/code_references.py
```diff
@@ -0,0 +1,245 @@
...
+                You can also suppress this warning by removing the
+                `suppress_bytecode_warnings` override from the `@version`
+                decorator on the corresponding function.
+                """
+                )
+                warnings.warn(message)
+                set_tos(None)
+            elif op.opname in ["LOAD_METHOD", "LOAD_ATTR"]:
+                if isinstance(tos, str):
+                    tos += "." + op.argval
```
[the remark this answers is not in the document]
namanjain: We didn't and it's a good point. But like you said, your suggestion in [another comment](https://github.com/square/bionic/pull/232#discussion_r542581595) should fix the issue.

docs/concepts.rst
```diff
@@ -316,17 +316,42 @@ file as long as the ``major`` version matches.
...
+   .. code-block:: python
+
+    def f():
+        a = get_a()  # We will inspect `get_a`.
+        b = module.submodule.get_b()  # We will inspect `get_b`.
+        import new_module
+        c = new_module.get_c()  # We won’t inspect `get_c`.
+        d = exec("get_d()")  # We definitely won’t inspect `get_d`!
+
+If you're not worried about false negatives, you can set Bionic to a "fully
```
jqmp, line 353: nit: I would say "*these* false negatives" to connect back to the previous section.

jqmp (approved):

PR #235, opened 2020-08-26
Add follow-up tasks

jqmp:
As noted in the commit messages below, I'd prefer to have some more fuzz testing in place before merging this, but I think the code is correct.

### Add KeyedPriorityStack data structure
We'll use this to add priority to our stack of pending task entries.

### Add follow-up tasks for tuples
This adds the concept of "follow-up tasks": when a task is finished, if
it has any "follow-up" tasks, those tasks must also be completed
immediately. This allows us to guarantee that when a tuple-returning
task is computed, all of the individual entities in the tuple are
persisted immediately.

Implementation-wise, this includes the following changes:
- Modifying TaskState creation to include these followups (which can now
introduce circular relationships between tasks)
- Modifying the TaskCompletionRunner to:
  - use a "priority stack" instead of a simple list
  - queue up follow-ups, at high priority, at the appropriate time
  - when dispatching tasks remotely, include any additional follow-ups
    that need to be computed as well

I think this code is correct, but I'd like to find a way to add more
fuzz tests before merging it.

### DIFF
--- a/bionic/core/flow_execution.py
+++ b/bionic/core/flow_execution.py
@@ -4,15 +4,17 @@ to completion.
 """
 
 from concurrent.futures import wait, FIRST_COMPLETED
-
 import logging
 
 from .task_execution import (
     EntryLevel,
+    EntryPriority,
     EntryStage,
     EntryRequirement,
+    RemoteSubgraph,
     TaskRunnerEntry,
 )
+from ..utils.keyed_priority_stack import KeyedPriorityStack
 from ..utils.misc import SynchronizedSet
 
 
@@ -23,9 +25,9 @@ from ..utils.misc import SynchronizedSet
 logger = logging.getLogger(__name__)
 
 
-def run_in_subprocess(task_completion_runner, state):
-    task_completion_runner.run([state])
-    return state.task.keys[0]
+def run_in_subprocess(task_completion_runner, states):
+    task_completion_runner.run(states)
+    return [state.task_keys[0] for state in states]
 
 
 class TaskCompletionRunner:
@@ -58,10 +60,10 @@ class TaskCompletionRunner:
 
         # These are used for caching and tracking.
         self._entries_by_task_key = {}
-        self._pending_entries = []
-        self._in_progress_entries = {}
-        # TODO This is a set but _in_progress_entries is a dict. Could they both be
-        # sets?
+        self._pending_entries_kps = KeyedPriorityStack()
+        self._in_progress_entries_by_task_key = {}
+        # TODO This is a set but _in_progress_entries_by_task_key is a dict. Could they
+        # both be sets?
         self._blocked_entries = set()
 
     @property
@@ -83,10 +85,10 @@ class TaskCompletionRunner:
             while self._has_pending_entries():
                 entry = self._activate_next_pending_entry()
                 self._process_active_entry(entry)
-                assert entry.stage != EntryStage.ACTIVE
+                assert entry.stage != EntryStage.ACTIVE, entry
 
-            assert len(self._pending_entries) == 0
-            assert len(self._in_progress_entries) == 0
+            assert len(self._pending_entries_kps) == 0
+            assert len(self._in_progress_entries_by_task_key) == 0
             assert len(self._blocked_entries) == 0
 
             results = {}
@@ -110,14 +112,13 @@ class TaskCompletionRunner:
 
         assert entry.stage == EntryStage.ACTIVE
 
-        # In theory we could have an entry that doesn't require priming, but in practice
-        # we always require it when we create one.
-        assert entry.required_level >= EntryLevel.PRIMED
+        # In theory we could have an entry that doesn't require initializing, but in
+        # practice we always require at least that much when we create one.
+        assert entry.required_level >= EntryLevel.INITIALIZED
 
         # First, if we've already met our requirements, we can stop immediately without
         # needing to process any dependencies.
-        if entry.all_incoming_reqs_are_met:
-            self._mark_entry_completed(entry)
+        if self._mark_entry_completed_if_possible(entry):
             return
 
         # Otherwise, we can't do anything more without the digests of our dependencies,
@@ -125,12 +126,15 @@ class TaskCompletionRunner:
         self._set_up_entry_dependencies(entry)
         for dep_entry in entry.dep_entries:
             self._add_requirement(entry, dep_entry, EntryLevel.PRIMED)
-        if not entry.all_outgoing_reqs_are_met:
-            self._mark_entry_blocked(entry)
+        if self._mark_entry_blocked_if_necessary(entry):
             return
 
         # Now that we have the dependency digests, we can initialize our task state.
         entry.state.initialize(self._bootstrap, self._flow_instance_uuid)
+
+        # If that was all we needed, we're done.
+        if self._mark_entry_completed_if_possible(entry):
+            return
 
         # If this entry is persistable, we may be able to load it from the persistent
         # cache, which would immediately get us to the CACHED level.
@@ -143,8 +147,7 @@ class TaskCompletionRunner:
             assert entry.level >= EntryLevel.PRIMED
 
         # If that was all we needed, we're done.
-        if entry.all_incoming_reqs_are_met:
-            self._mark_entry_completed(entry)
+        if self._mark_entry_completed_if_possible(entry):
             return
 
         # At this point we'll need to actually compute the entry. If possible, we
@@ -155,8 +158,67 @@ class TaskCompletionRunner:
             and not entry.state.task.is_simple_lookup
         )
         if can_compute_remotely:
-            new_state_for_subprocess = entry.state.strip_state_for_subprocess()
-            # We want serial execution in the subprocesses.
+            # When we run an entry remotely, we also need to run all of its immediate
+            # non-persistable ancestors, since their values can't be shared between
+            # processes. Those ancestors may have follow-up tasks, so we'll need to
+            # run those too; and if those follow-ups are persistable (but not actually
+            # persisted yet), their values will be usable everywhere, so we'll want to
+            # track the fact that they're running. So really we're going to run a
+            # collection of persistable entries (including the original entry), plus
+            # their immediate non-persistable ancestors.
+            remote_subgraph = RemoteSubgraph(entry.state, self._bootstrap)
+            target_entries = [
+                self._get_or_create_entry_for_state(persistable_subgraph_state)
+                for persistable_subgraph_state in remote_subgraph.persistable_but_not_persisted_states
+            ]
+            assert entry in target_entries
+
+            # We want all of our entries to be initialized before we try running
+            # them remotely. We'll just require this of the persistable ones, since
+            # all the other ones are ancestors of them.
+            for target_entry in target_entries:
+                if target_entry == entry:
+                    assert target_entry.level >= EntryLevel.INITIALIZED
+                else:
+                    self._add_requirement(
+                        src_entry=entry,
+                        dst_entry=target_entry,
+                        level=EntryLevel.INITIALIZED,
+                    )
+            if self._mark_entry_blocked_if_necessary(entry):
+                return
+
+            for target_entry in target_entries:
+                # Now we'll also require that each entry be CACHED, which guarantees
+                # that it won't be COMPLETED. (If we didn't do this, the entry's
+                # required level might only be INITIALIZED, in which case it would
+                # already be COMPLETED. However, we know it's not already CACHED,
+                # because it's `persistable_but_not_persisted`.)
+                self._add_requirement(
+                    src_entry=None,
+                    dst_entry=target_entry,
+                    level=EntryLevel.CACHED,
+                )
+                if target_entry.stage == EntryStage.PENDING:
+                    self._mark_entry_active(target_entry)
+                # The entry should now be ACTIVE, because:
+                # - It can't be COMPLETED because of the requirement we added above.
+                # - It can't be PENDING because we would have just activated it above.
+                # - It shouldn't be BLOCKED, because:
+                #   - we already required the entry to be at least INITIALIZED, and
+                #   - it should also be eligible for remote computation and part of the
+                #     same subgraph as this one, so it shouldn't have any additional
+                #     requirements compared to the original entry.
+                # - It shouldn't be IN_PROGRESS, because if it's a target entry for our
+                #    our original entry, then the reverse should also be true (it's a
+                #    symmetric relationship), so our original entry would already be
+                #    IN_PROGRESS or COMPLETED too.
+                assert target_entry.stage == EntryStage.ACTIVE
+
+            stripped_target_states = [
+                remote_subgraph.get_stripped_state(target_entry.state)
+                for target_entry in target_entries
+            ]
             new_bootstrap = self._bootstrap.evolve(executor=None)
             new_task_completion_runner = TaskCompletionRunner(
                 bootstrap=new_bootstrap,
@@ -166,21 +228,22 @@ class TaskCompletionRunner:
             future = self._bootstrap.executor.submit(
                 run_in_subprocess,
                 new_task_completion_runner,
-                new_state_for_subprocess,
+                stripped_target_states,
             )
-            self._mark_entry_in_progress(entry, future)
+
+            for target_entry in target_entries:
+                self._mark_entry_in_progress(target_entry, future)
             return
 
         # Otherwise we'll compute this entry locally. In that case, it's not enough
         # for our dependencies to be primed; they also need to have cached values.
         for dep_entry in entry.dep_entries:
             self._add_requirement(entry, dep_entry, EntryLevel.CACHED)
-        if not entry.all_outgoing_reqs_are_met:
-            self._mark_entry_blocked(entry)
+        if self._mark_entry_blocked_if_necessary(entry):
             return
+
         entry.compute(self.task_key_logger)
-        assert entry.all_incoming_reqs_are_met
-        self._mark_entry_completed(entry)
+        assert self._mark_entry_completed_if_possible(entry)
 
     def _set_up_entry_dependencies(self, entry):
         if entry.dep_entries is not None:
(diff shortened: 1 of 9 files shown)


bionic/core/task_execution.py
```diff
@@ -59,6 +59,7 @@ def __init__(self, state):
         self.state = state
         self.future = None
         self.results_by_dnode = None
+        self._stage = None
         self.stage = EntryStage.COMPLETED
```
namanjain, line 63: You don't need this anymore.

bionic/core/flow_execution.py
```diff
@@ -155,18 +162,70 @@ def _process_active_entry(self, entry):
...
+
+            target_entries = [
+                persistable_subgraph_entry
+                for persistable_subgraph_entry in persistable_subgraph_entries
+                if persistable_subgraph_entry.stage
+                not in (EntryStage.COMPLETED, EntryStage.IN_PROGRESS)
+            ]
+            for target_entry in target_entries:
+                if target_entry.stage == EntryStage.PENDING:
+                    self._mark_entry_active(target_entry)
```
namanjain, line 204: This means that more than one entry can be active now. We should change the documentation of `EntryStage.ACTIVE` to reflect this.

bionic/core/flow_execution.py
```diff
@@ -155,18 +162,70 @@ def _process_active_entry(self, entry):
...
+                    )
+            if not entry.all_outgoing_reqs_are_met:
+                self._mark_entry_blocked(entry)
+                return
+
+            target_entries = [
+                persistable_subgraph_entry
+                for persistable_subgraph_entry in persistable_subgraph_entries
+                if persistable_subgraph_entry.stage
+                not in (EntryStage.COMPLETED, EntryStage.IN_PROGRESS)
```
namanjain, line 200: I'm not entirely convinced about the remote computation working if we exclude the in progress entries. When can such an entry be in progress? If it is, can we get to a state in remote computation (that this block starts) where some other entry might be blocked on this entry, sees that it is not computed and also starts computing it?

bionic/provider.py
```diff
@@ -99,6 +99,18 @@ def entity_names(self):
...
+    # (like a tuple), then we'll need to set some followup tasks to make sure all the
+    # aggregated descriptors get persisted immediately. This field is gross, but once we
+    # have file descriptors it can go away, because the "crude" or "not-yet-persisted"
+    # values will have their own descriptors, and it will be obvious which tasks need
+    # to have followups.
+    # TODO In the meantime, if we start moving more attributes directly onto the Task
+    # class, this would be another good candidate.
+    @property
+    def task_output_may_need_followups_for_persistence(self):
+        return True
```
namanjain, line 112: I don't understand what providers would use this property to be true? Is there an easy way where we can figure out those providers and set it to `True` only for them, while keeping the default value `False`?

bionic/utils/keyed_priority_stack.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def push(self, key, value, priority):
+        """
+        Adds a value to the stack with associated key and priority.
+        """
+
+        if key is None:
+            raise KeyError("Attempted to add None as key to priority stack")
+        if key in self._unremoved_entries_by_key:
+            raise KeyError(f"Attempted to add duplicate key to priority stack: {key!r}")
```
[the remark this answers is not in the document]
jqmp: Oh, good point. Yeah, maybe `ValueError` is more appropriate here.

bionic/utils/keyed_priority_stack.py
```diff
@@ -0,0 +1,108 @@
...
+        If no key is provided, removes and returns the highest-priority element (or
+        the last-added such element, if there is a tie).
+
+        If a key is provided, removes and returns the element with that key.
+        """
+
+        if key is not None:
+            if key not in self._unremoved_entries_by_key:
+                raise KeyError(f"Key not found in priority stack: {key!r}")
+            entry = self._unremoved_entries_by_key.pop(key)
```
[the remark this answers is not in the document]
jqmp: Yep, that was the goal.

bionic/utils/keyed_priority_stack.py
```diff
@@ -0,0 +1,108 @@
...
+            entry.is_removed = True
+            self._n_unremoved_entries -= 1
+            return entry.value
+
+        else:
+            while True:
+                if self._n_unremoved_entries == 0:
```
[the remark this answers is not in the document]
jqmp: Can it? It seems to me that `_n_unremoved_entries` could be non-empty, but all the entries could turn out to be `is_removed`, and then we'd want this `if` statement to trigger.

bionic/utils/keyed_priority_stack.py
```diff
@@ -0,0 +1,108 @@
...
+                    raise IndexError("Attempted to get item from empty priority stack")
+                entry = heapq.heappop(self._heap)
+                if entry.is_removed:
+                    continue
+                self._n_unremoved_entries -= 1
+                if entry.key is not None:
```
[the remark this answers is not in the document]
jqmp: No, I think that's left over from when the API was more flexible. Nice catch!

bionic/protocols.py
```diff
@@ -642,7 +642,7 @@ def write(self, value, path):
         self._protocol_for_value(value).write(value, path)
 
     def __repr__(self):
-        return f"CombinedProtocol{tuple(self._subprotocols)!r}"
+        return "CombinedProtocol(...)"
```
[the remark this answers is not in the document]
jqmp: Oh, good question. I found the original version super long and not very informative, because we always use the same combination of protocols. It just takes up a lot of screen space anytime you print a protocol (because most things use the same default protocol, which is a big CombinedProtocol).

bionic/utils/keyed_priority_stack.py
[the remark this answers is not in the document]
namanjain: I think it can. You only pop the entries from heap before continue, `_n_unremoved_entries` never changes (only `_heap` size does) and the if conditional would always resolve to the same value. That said, maybe the part after the continue should be moved out as well, so that it's clear what exactly we are looping on.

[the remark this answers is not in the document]
jqmp: Oh yeah, you're right!

bionic/core/task_execution.py
```diff
@@ -695,50 +768,71 @@ def _get_digests_by_dnode(self):
...
-            task_state.task = None
-        else:
-            task_state.dep_states = [
-                dep_state.strip_state_for_subprocess(new_task_states_by_key)
-                for dep_state in task_state.dep_states
-            ]
+        if stripped_state.should_persist:
+            assert len(stripped_state.followup_states) == 0
+
+            if stripped_state.is_cached_persistently:
```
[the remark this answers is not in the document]
jqmp: Yeah, I figured we might as well be explicit, since being *persistently* cached is specifically what we care about here.

bionic/provider.py
jqmp: We discussed this in a call, but to summarize: this should be true for any provider than can provide a tuple, except `TupleConstructionProvider`.

In a sense, the underlying problem here is that tuples are sometimes upstream of their components and sometimes downstream, which creates a bunch of logical issues, including this one. Once we have file descriptors that will be resolved: crude tuples will always be upstream; regular tuples will always be downstream.

bionic/core/task_execution.py
```diff
@@ -695,50 +768,71 @@ def _get_digests_by_dnode(self):
...
+    of their follow-up tasks (which need to be computed immediately after their
+    parents). It also maintains a copy of this subgraph with unnecessary data pruned;
+    these "stripped" TaskStates can be safely serialized with cloudpickle and sent to
+    the other process.
+    """
 
-        Parameters
-        ----------
+    def __init__(self, target_state, bootstrap):
+        self._target_state = target_state
```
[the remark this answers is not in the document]
jqmp: Nope!

bionic/core/task_execution.py
jqmp: I think we do, because when we set `stage` on the next line, the implementation checks the value of `_stage`.

bionic/core/flow_execution.py
jqmp: We discussed this in a call. I don't think this can happen, but I agree we should be clearer about how much to depend on that assumption.

bionic/core/task_execution.py
namanjain: But won't `self.stage` be overridden by the property you defined later? Did you mean to initialize `_stage` with `COMPLETED` or is this some python magic where you can initialize a property this way?

jqmp: By the time we call `__init__`, all the methods defined in the class already exist, so when we do `self.stage = EntryStage.COMPLETED`, that calls the property setter.

bionic/core/task_execution.py
```diff
@@ -695,50 +768,71 @@ def _get_digests_by_dnode(self):
...
-        if task_state.is_cached:
-            task_state.dep_states = []
-            task_state.task = None
-        else:
-            task_state.dep_states = [
-                dep_state.strip_state_for_subprocess(new_task_states_by_key)
-                for dep_state in task_state.dep_states
-            ]
+        if stripped_state.should_persist:
+            assert len(stripped_state.followup_states) == 0
```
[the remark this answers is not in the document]
jqmp: Yes, it does seem pretty important. I feel like it's valuable to have the assertion here because it's related to work we're doing, but I'd be happy to also put it in `set_up_caching_flags`.

jqmp:
Now that the new random tests have been merged, I've added a commit here that extends them to use tuple outputs as well. They should all be passing now, which gives me more confidence that the code in this PR is correct.

bionic/core/flow_execution.py
```diff
@@ -155,8 +165,66 @@ def _process_active_entry(self, entry):
...
+                        src_entry=entry,
+                        dst_entry=target_entry,
+                        level=EntryLevel.INITIALIZED,
+                    )
+            if not entry.all_outgoing_reqs_are_met:
+                self._mark_entry_blocked(entry)
+                return
+
+            for target_entry in target_entries:
+                # Now we'll also require that each entry be cached, which guarantees
```
namanjain, line 200: This is very neat (and a little tricky)! I understand now why we need it, but I'm sure I'll be a little confused when I see this code later down the road. Can you also add a line or two about how it can be `COMPLETED` without this requirement (which IIUC, is because the follow up dependency might not have anyone consuming it yet)?

bionic/core/task_execution.py
```diff
@@ -59,22 +59,29 @@ def __init__(self, state):
...
 
         # This is initially set to None to avoid eagerly recursing through the entire
         # DAG. We set this once we start processing the entry.
         self.dep_entries = None
 
         self.incoming_reqs = set()
         self.outgoing_reqs = set()
+        self.priority = EntryPriority.DEFAULT
+
+        self.stage = EntryStage.COMPLETED
```
namanjain, line 72: This is mostly for my understanding of setters, is there a reason to do it at the end? Like is it supposed to be treated like an initialize method that we would call at the end (once all the variables are initialized)?

tests/test_flow/test_persistence_random.py
```diff
@@ -371,3 +425,13 @@ def test_random_flow_varied(harness, n_bindings, n_iterations, p_persist, p_mult
         p_multi_out=p_multi_out,
     )
     RandomFlowTester(harness, config).run()
+
+
+@pytest.mark.slow
+def test_random_flow_no_tuples(harness):
```
namanjain, line 431: Since this a slow test, we probably want to remove it once we "activate" tuple descriptors. Does this sound right? If it does, we should leave a TODO to remove this test and clean up this logic.

tests/test_flow/test_persistence_random.py
```diff
@@ -248,6 +281,8 @@ class RandomFlowTestConfig:
     n_iterations = attr.ib()
     p_persist = attr.ib(default=0.5)
     p_multi_out = attr.ib(default=0.3)
+    p_three_out_given_multi = attr.ib(default=0.3)
```
namanjain, line 284: Since you already have documentation for the older variables, it would be nice to add it for the new variables as well.

tests/test_flow/test_persistence_random.py
```diff
@@ -343,11 +382,26 @@ def new():
...
+
+        include_three_entities = self._random_bool(
+            p_true=self._config.p_three_out_given_multi
+        )
+        if not include_three_entities:
             return f"{new()}, {new()}"
 
+        return self._rng.choice(
+            [
+                lambda: f"{new()}, {new()}, {new()}",
```
namanjain, line 397: nit: so many brackets 😵. Can you add a small example as a comment after each choice? I mean something like `a, b, c`, `(a, b), c`, etc.

tests/test_flow/test_persistence_random.py
```diff
@@ -108,15 +110,28 @@ def _add_binding_to_flow(self, binding):
 
         joined_dep_entity_names = ", ".join(binding.dep_entity_names)
 
+        # TODO Could this logic live inside ModelBinding instead?
```
namanjain, line 113: I was going to suggest this for `compute_func` at least. But I see why you have it here now because of the output decorator.

bionic/core/flow_execution.py
```diff
@@ -110,12 +112,13 @@ def _process_active_entry(self, entry):
...
-        # In theory we could have an entry that doesn't require priming, but in practice
-        # we always require it when we create one.
-        assert entry.required_level >= EntryLevel.PRIMED
+        # In theory we could have an entry that doesn't require initializing, but in
+        # practice we always require at least that much when we create one.
+        assert entry.required_level >= EntryLevel.INITIALIZED
 
         # First, if we've already met our requirements, we can stop immediately without
         # needing to process any dependencies.
+        self._unblock_entries_with_met_requirements(entry)
```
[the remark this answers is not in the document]
jqmp: Yes, this makes me uncomfortable too -- I think because of the implicit requirement that we do this every time an entry's level could have changed. Of course, this requirement is specific to `_process_active_entry`, which already has an idiosyncratic structure.

The `_attempt_to_mark_complete` idea is interesting -- I'll try it out. I wonder if we can combine the "check if complete", "check if blocked", and "check for met requirements" into some kind of `_check_if_processing_should_continue` method.

bionic/core/flow_execution.py
```diff
@@ -155,8 +165,66 @@ def _process_active_entry(self, entry):
...
+                        level=EntryLevel.INITIALIZED,
+                    )
+            if not entry.all_outgoing_reqs_are_met:
+                self._mark_entry_blocked(entry)
+                return
+
+            for target_entry in target_entries:
+                # Now we'll also require that each entry be cached, which guarantees
+                # that it won't be COMPLETED. (We already know it's
+                # `persistable_but_not_persisted`, so it can't be cached.)
```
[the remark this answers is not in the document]
jqmp: Actually this was intended to be "can't" as in "logically must not":

1. it's persistable
2. therefore CACHED == persisted
3. but it's not persisted
4. therefore it's not CACHED

I'll make this more explicit.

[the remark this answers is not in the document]
namanjain: Ahhhhh, that also makes sense. It might also be easier to say "it isn't cached" than adding more explanation.

bionic/core/flow_execution.py
jqmp: Yep, adding more discussion here makes sense.

Also, I'm really hoping we can simplify away a lot of this by introducing file descriptors.

tests/test_flow/test_persistence_random.py
jqmp: I agree we should rethink this once we activate tuple descriptors. I'm not certain we should totally remove it, since it's still useful to know that both `@bn.outputs` and `@bn.returns` work. But maybe we should just randomly use one or the other only when they're equivalent.

tests/test_flow/test_persistence_random.py
jqmp: Oh yeah! Thanks.

tests/test_flow/test_persistence_random.py
jqmp: Yes, I think this sentiment applies to both of these functions. Unfortunately they both need some global context (mainly `__descriptors_computed_by_XXX`), but that could be factored out.

bionic/core/flow_execution.py
```diff
@@ -155,8 +165,66 @@ def _process_active_entry(self, entry):
...
-            # We want serial execution in the subprocesses.
+            # When we run an entry remotely, we also need to run all of its immediate
+            # non-persistable ancestors, since their values can't be shared between
+            # processes. Those ancestors may have follow-up tasks, so we'll need to
+            # run those too; and if those follow-ups are persistable (but not actually
+            # persisted yet), their values will be usable everywhere, so we'll want to
+            # track the fact that they're running. So really we're going to run a
+            # collection of persistable entries (including the original entry), plus
+            # their immediate non-persistable ancestors.
+            remote_subgraph = RemoteSubgraph(entry.state, self._bootstrap)
```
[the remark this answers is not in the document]
jqmp: I thought so too, but we really need to get both the stripped states *and* the target entries. Or maybe I'm misunderstanding what your proposed method would look like?

bionic/core/task_execution.py
jqmp: Oh, no, there's no good reason. This was leftover from when I had some more debugging code in the `stage` property. I'll move it back.

bionic/core/flow_execution.py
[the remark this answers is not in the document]
namanjain: Nope, I missed that we use both stripped states and target entries.

bionic/core/flow_execution.py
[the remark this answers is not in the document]
jqmp: I pushed a commit that refactors this (among others). Let me know what you think.

bionic/core/flow_execution.py
```diff
@@ -192,9 +185,8 @@ def _process_active_entry(self, entry):
                         dst_entry=target_entry,
                         level=EntryLevel.INITIALIZED,
                     )
-            if not entry.all_outgoing_reqs_are_met:
-                self._mark_entry_blocked(entry)
-                return
+                if self._mark_entry_blocked_if_necessary(entry):
```
namanjain, line 188: Shouldn't this be outside the for loop (since it's replacing the if condition outside the loop)?

tests/test_flow/test_persistence_random.py
```diff
@@ -338,15 +381,34 @@ def _random_existing_entity_name(self):
...
+        )
+        if not include_three_entities:
+            template = "X, X"
+
         else:
-            return f"{new()}, {new()}"
+            # We'll randomly choose one of the following three-entity descriptors:
+            template = self._rng.choice(
+                [
+                    "X, X, X",
```
namanjain, line 403: This is so much better!

namanjain (approved):
I feel better with the new changes!

bionic/core/flow_execution.py
jqmp: Yep! That was a copy-and-paste error. (I guess it still works logically where it is now, but it's more efficient to put it outside the loop.)

tests/test_flow/test_persistence_random.py
jqmp: I agree! Thanks for pushing back on the original version.

PR #236, opened 2020-08-27
Reformat code using black

simonafk:


simonafk:
Added commit comments

PR #238, opened 2020-08-28
Update pre-commit config

namanjain:
Pre-commit package versions and my local versions are out of sync
and that's causing a lint error. This should fix the problem.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,11 +1,11 @@
 repos:
 -   repo: https://gitlab.com/pycqa/flake8
-    rev: 3.7.9
+    rev: 3.8.3
     hooks:
     -   id: flake8
 
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v2.4.0
+    rev: v3.2.0
     hooks:
     -   id: trailing-whitespace
     -   id: end-of-file-fixer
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -1,19 +1,19 @@
...
+    rev: v3.2.0
     hooks:
     -   id: trailing-whitespace
     -   id: end-of-file-fixer
     -   id: check-yaml
     -   id: check-added-large-files
 
 -   repo: https://github.com/psf/black
-    rev: stable
+    rev: 20.8b1
```
jqmp, line 16: Was `stable` not doing what we want here?

namanjain: That's what I thought too. I followed the pre-commit [autoupdate documentation](https://pre-commit.com/#pre-commit-autoupdate) and none of the option let me keep the `stable` rev.

We can overwrite the revision to stable after we run autoupdate. Wdyt?

jqmp: Well, fair enough. It's also a little confusing to have some of our pre-commit dependences based on a moving target and others pinned, so I guess I don't know if either solution is obviously better.

namanjain: Then I'll keep the default autoupdate behavior and pin all the revisions, so that we don't have to worry about the extra step to update it after running the command. Lmk if you think the other way is better (even if slightly).

PR #239, opened 2020-08-31
Add new random flow tester

jqmp:
This is an alternative implementation of the random testing ideas in
`test_persistence_fuzz.py`, but focused on persistence vs
non-persistence and multi-output functions instead of versioning. The
goal is to allow us to randomly test the tuple persistence capabilities
(both serial and parallel) that I've been working on.

I basically rewrote `test_persistence_fuzz.py` because I was finding it
hard to add support for the new features above. The new implementation
models a little more of what's going on in a flow, so it shouldn't be
too hard to support what the fuzzer does as well (versioning modes, GCS,
and non-determinism) and eventually remove the old fuzzer code
altogether.

The new implementation also has a different algorithm for generating
graphs which produces more realistic and efficient flows, which means we
can test larger flows in a reasonable amount of time.

Finally, I've avoided the term "fuzzer" in this new implementation
because what we're doing doesn't actually feel like "fuzz testing". I
think "random testing" or "model-based testing" are more accurate.

### DIFF
--- a/tests/test_flow/test_persistence_random.py
+++ b/tests/test_flow/test_persistence_random.py
@@ -1,0 +1,373 @@
+import pytest
+
+from textwrap import dedent
+
+import attr
+import numpy as np
+
+from bionic.descriptors.parsing import dnode_from_descriptor
+from bionic.utils.misc import single_unique_element
+import bionic as bn
+
+
+class ModelFlowHarness:
+    """
+    Manages a Bionic flow and tests its behavior against a "model" implementation.
+
+    This class wraps both a Bionic flow and a simplified model implementation of the
+    same flow. The flows are kept in sync, allowing their behavior to be checked against
+    each other. This allows large flows to be constructed and validated
+    programmatically.
+
+    The model flow implementation supports the following functionality:
+
+    - creating string-valued entities derived from a variable number of other entities
+    - making functional changes to the definition of any entity
+    - testing the value of an entity and the set of (user-provided) functions computed
+      in the process
+    - persisted and non-persisted entities
+    - functions with multiple entity outputs
+
+    This class is similar to the ``SimpleFlowModel`` ("SFM") in ``test_persisted_fuzz``,
+    but with the following differences:
+
+    - SFM does not support non-persisted entities or multiple entity outputs
+    - this class does not support the following:
+      - non-deterministic (``changes_per_run``) entities
+      - non-functional updates
+      - alternative versioning modes
+      - cloud caching
+
+    This class has a more sophisticated implementation than SFM, so it should be
+    easier to port SFM's features to this class than vice versa. If we can port all of
+    the above features, we can remove SFM altogether.
+    """
+
+    def __init__(self, builder, make_list):
+        self._builder = builder
+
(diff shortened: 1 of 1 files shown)


tests/test_flow/test_persistence_random.py
```diff
@@ -0,0 +1,375 @@
...
+    def _random_existing_entity_name(self):
+        return self._rng.choice(self._harness.get_all_entity_names())
+
+    def _random_descriptor_of_new_entities(self, should_persist):
+        def new():
+            entity_name = self._harness.create_entity(should_persist)
+            self._child_counts_by_entity_name[entity_name] = 0
+            return entity_name
+
+        if self._random_bool(p_true=1 - self._config.p_multi_out):
```
namanjain, line 348: super nit: can you invert this if condition? Then, you won't have to complement the multi-output probability.

tests/test_flow/test_persistence_random.py
```diff
@@ -0,0 +1,375 @@
...
+        return tuple(
+            out_entity.compute_value(dep_values) for out_entity in self.out_entities
+        )
+
+    def _compute_entity_value(self, entity_name, dep_values):
+        return f"{entity_name}.{self.func_version}({','.join(dep_values)})"
+
+    def compute_and_save_values(self, dep_values):
+        values_by_entity_name = {
+            entity.name: self._compute_entity_value(entity.name, dep_values)
```
namanjain, line 202: You can replace `self._compute_entity_value` with `entity.compute_value` here. Once you do, you can then remove `_compute_entity_value` method and avoid duplicating that logic in two places.

tests/test_flow/test_persistence_random.py
```diff
@@ -0,0 +1,375 @@
...
+        if self.should_persist:
+            self.persisted_values_by_entity_name = values_by_entity_name
+            self.has_persisted_values = True
+        self.some_descendant_might_have_persisted_values = True
+        return values_by_entity_name
+
+    def invalidate_persisted_values(self):
+        if not self.some_descendant_might_have_persisted_values:
+            return
+        self.has_persisted_values = False
```
namanjain, line 214: IIUC, I think we should also clear `self.persisted_values_by_entity_name` with `self. has_persisted_values`.

tests/test_flow/test_persistence_random.py
jqmp: Yep, I agree this would improve it. I actually have a follow-up PR (adding support for tuple descriptors) that fixes this, so I may wait for that one if you don't mind.

tests/test_flow/test_persistence_random.py
jqmp: Oh yeah, you're right. That was the result of an incomplete refactor.

tests/test_flow/test_persistence_random.py
jqmp: My thinking here was that if `has_persisted_values` is false, then the value of `persisted_values_by_entity_name` is irrelevant, but I agree it's clearer to just set it to `None`.

PR #240, opened 2020-09-01
Fix visualization of tuple nodes

jqmp:
Pydot can break when it encounters special characters, so we need some
special handling when visualizing nodes that aren't just entities. I've
added a singleton tuple to our test fixture; once we activate tuple
descriptors everywhere, the `outputs` part of the test will also use a
tuple.

### DIFF
--- a/bionic/dagviz.py
+++ b/bionic/dagviz.py
@@ -102,17 +102,22 @@ def dot_from_graph(graph, vertical=False, curvy_lines=False, name=None):
     )
 
     def name_from_node(node):
-        return graph.nodes[node]["name"]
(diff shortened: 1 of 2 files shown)


namanjain (approved):
I didn't test it out myself but it looks correct. Do you mind adding some screenshots of a DAG before and after the fix?

jqmp:
> I didn't test it out myself but it looks correct. Do you mind adding some screenshots of a DAG before and after the fix?

Sure. The old behavior was that it would crash if any node name ended with a comma. The new behavior produces an image like this:

image

PR #242, opened 2020-09-02
Improve error reporting for multiple outputs

jqmp:
These commits introduce a new exception type for protocol validation failures; improve the reporting of one specific error case; and add some new tests for reporting errors with tuple descriptors.


namanjain (approved):
:shipit:

PR #243, opened 2020-09-03
test_reload.py use tmp cache dir

simonafk:
test_reload.py to use temporary pytest directory, which are unique for each test run, for caching so that tests do not interfere with each other.

### DIFF
--- a/tests/test_flow/test_reload.py
+++ b/tests/test_flow/test_reload.py
@@ -40,30 +40,8 @@ def write_module(tmp_module_path):
     sys.dont_write_bytecode = dont_write_bytecode
 
 
-@pytest.fixture
-def write_flow_module(write_module):
-    def _write_flow_module(x):
-        write_module(
-            "flow_module",
-            f"""
-                import bionic as bn
-
-                builder = bn.FlowBuilder('flow')
-                builder.assign('x', {x})
-
-                @builder
-                def x_doubled(x):
-                    return x*2
-
-                flow = builder.build()
-            """,
-        )
-
-    return _write_flow_module
-
-
 @pytest.fixture(params=["reload", "reloading"])
-def reload_method(request):
+def reload_method_name(request):
     return request.param
 
 
@@ -77,12 +55,32 @@ class FlowTester:
     before the reload.
     """
 
-    def __init__(self, reload_method):
-        from flow_module import flow
-
-        self.flow = flow
+    def __init__(self, reload_method_name, tmp_path, write_module):
+        self.flow = None
         self.expected_results_by_flow = {}
-        self.reload_method = reload_method
+        self.reload_method_name = reload_method_name
+        self.cache_dir = str(tmp_path / "BNTESTDATA")
+        self.write_module = write_module
+
+    def write_flow_module(self, body):
+        header = f"""
+            import bionic as bn
+
+            builder = bn.FlowBuilder('flow')
+            builder.set('core__persistent_cache__flow_dir', '{self.cache_dir}')
+        """
+        footer = """
+            flow = builder.build()
+        """
+        self.write_module(
+            "flow_module",
+            dedent(header) + "\n" + dedent(body) + "\n" + dedent(footer),
+        )
+
+        if self.flow is None:
+            from flow_module import flow
+
+            self.flow = flow
 
     def expect_values(self, **kwargs):
         expected = dict(kwargs)
@@ -91,10 +89,7 @@ class FlowTester:
             self._verify_flow_contents(f, r)
 
     def reload_flow(self):
-        if self.reload_method == "reload":
-            self.flow.reload()
-        else:
-            self.flow = self.flow.reloading()
+        self.flow = getattr(self.flow, self.reload_method_name)()
 
     def get(self, name):
         return self.flow.get(name)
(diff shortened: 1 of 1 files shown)


tests/test_flow/test_reload.py
```diff
@@ -40,216 +40,240 @@ def _write_module(name, contents):
...
+    flow_tester.expect_values(one=1, x=3, x_doubled=6)
 
 
 # Test nondeterministic function marked with the changes_per_run decorator
-def test_changes_per_run_decorator(write_module, reload_method):
+def test_changes_per_run_decorator(flow_tester):
     def write_flow_module():
-        write_module(
-            "flow_module",
+        flow_tester.write_tester_flow(
```
jqmp, line 259: WDYT about using `flow.setting('core__persistent_cache__flow_dir', ...)` to modify the flow *after* loading it? This would let you avoid this awkward two-part module writing.

(To elaborate: whenever you set or reload `self.flow` inside `FlowTester`, you also create a modified copy with the appropriate cache setting, and you use that for all your testing.)

simonafk: For `reload` to work, the flow needs to be the first flow built from the builder, and it has not been modified. `flow.setting` breaks this requirement.

tests/test_flow/test_reload.py
```diff
@@ -40,216 +40,240 @@ def _write_module(name, contents):
     sys.dont_write_bytecode = dont_write_bytecode
 
 
+@pytest.fixture(params=["reload", "reloading"])
+def reload_method(request):
```
namanjain, line 44: nit: I know we approved this before, but can you rename this to something like `reload_type`? It reads like we are passing around a method function.

tests/test_flow/test_reload.py
jqmp: Right, I was thinking you would store both the original flow and the modified flow. Whenever you need to call `reload`, you call it on the original flow, and then you use `setting` to get the modified reloaded version.

tests/test_flow/test_reload.py
```diff
@@ -40,216 +40,240 @@ def _write_module(name, contents):
...
+            self.flow = None
+            self.expected_results_by_flow = {}
 
-                flow = builder.build()
-            """,
-        )
+        def write_tester_flow(self, contents):
+            write_module(
+                "flow_module",
+                f"""
```
namanjain, line 67: You can avoid this awkward indent by breaking this flow string into three dedented strings (before `contents`, `contents`, and after `contents`) and joining them.

namanjain: Or you can do what Janek suggested.

tests/test_flow/test_reload.py
simonafk: I have modified the `FlowTester` class to actually call the named method i.e. `getattr(.., reload_method)()`. This ensures that there is no typo.

namanjain: How about `reload_method_name`?

tests/test_flow/test_reload.py
```diff
@@ -40,216 +40,240 @@ def _write_module(name, contents):
...
+
 @pytest.fixture
-def write_flow_module(write_module):
-    def _write_flow_module(x):
-        write_module(
-            "flow_module",
-            f"""
-                import bionic as bn
+def flow_tester(reload_method, tmp_path, write_module):
+    class FlowTester:
```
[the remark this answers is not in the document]
simonafk: If the class is moved outside of the function, then the parameters `reload_method_name, tmp_path, write_module` will need to be passed in. I think the test is simpler having the `FlowTester` class being nested inside of the function?

tests/test_flow/test_reload.py
simonafk: Hmm, that would make the test less simple, and it would also change the nature of the flow as well; we are testing a flow that exist in the `test_reload.py`, not in the temporary `flow_module.py`.

tests/test_flow/test_reload.py
simonafk: I like this; it maintains the test's simplicity.

tests/test_flow/test_reload.py
jqmp: Personally I think it would make the tests simpler, because each test would be writing a complete module that defines its own `builder`, instead of having to know that `write_tester_flow` magically adds it to the code. And I don't think it changes the nature of the flow in a meaningful way; the key thing is whether you see the old or new behavior, and using `setting` to configure the cache won't change that.

That said, it's up to you!

tests/test_flow/test_reload.py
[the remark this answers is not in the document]
jqmp: I agree with Naman that it feels more natural to define the class outside of the fixture. It does add some annoying extra lines of code, but it makes it easier to keep track of the places we're storing state. Fundamentally, classes and closures are just alternative ways of associating state with functions, and it feels weird to use them together to store different parts of our state.

In the end, though, I think it's a pretty minor difference.

tests/test_flow/test_reload.py
```diff
@@ -40,216 +40,242 @@ def _write_module(name, contents):
...
+        new copy of the flow with reloaded modules. We need to verify that the
+        entities of the original flow instance give the same values when computed as
+        before the reload.
+        """
+
+        def __init__(self):
+            self.flow = None
+            self.expected_results_by_flow = {}
+
+        def write_tester_flow(self, body):
```
jqmp, line 64: nit: this method might be clearer as `write_flow_module`; that makes it a bit clearer how it relates to `write_module`, and of course the word `tester` is a bit redundant here.

jqmp (approved):

tests/test_flow/test_reload.py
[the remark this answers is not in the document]
simonafk: Good point on storing state in one place. Moved class outside of the fixture.

PR #244, opened 2020-09-03
reload add versioning documentation

simonafk:
In documentation for reload, add note about versioning since it is
important.

### DIFF
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -209,6 +209,8 @@ Bionic's caching can be seen in action in the `ML tutorial`_.
 
 .. _ML tutorial: tutorials/ml_workflow.ipynb
 
+.. _cache-invalidation-and-versioning:
+
 Cache Invalidation and Versioning
 .................................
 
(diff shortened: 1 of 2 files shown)


docs/concepts.rst
```diff
@@ -956,6 +958,11 @@ procedure -- in complicated cases, it may not be able to figure out how to do
 this.  In these cases it will try to throw an exception rather than fail
 silently.) All future operations on the flow will reflect its updated status.
 
+To ensure that any functional changes are detected properly, you will need to
```
jqmp, line 961: wording nits:

- you have "To ensure that any functional changes are detected properly," twice; I'd drop the first one.
- "the versioning feature" can probably just be "versioning".
- :"functional changes": "code changes" might be clearer
- "may use stale cache": may want to be precise here and say something like "may keep using cached values from previous versions of the code"

Likewise for the docstring above.

jqmp (approved):

jqmp (approved):
LGTM!

PR #245, opened 2020-09-03
reload for merged flows

simonafk:
Fix reload so that it works for flows which are merged from another.

### DIFF
--- a/bionic/flow.py
+++ b/bionic/flow.py
@@ -8,10 +8,8 @@ import shutil
 import warnings
 from collections import defaultdict
 from pathlib import Path, PosixPath
-from importlib import reload
 from textwrap import dedent
 from uuid import uuid4
-
 import pyrsistent as pyrs
 import pandas as pd
 
@@ -49,6 +47,7 @@ from .utils.misc import (
     check_at_most_one_present,
     oneline,
 )
+from .utils.reload import recursive_reload
 
 DEFAULT_PROTOCOL = protos.CombinedProtocol(
     protos.JsonProtocol(),
@@ -1461,7 +1460,6 @@ class Flow:
         * has never been modified (i.e., isn't derived from another Flow)
         * is assigned to a top-level variable in a module that one of its
           functions is defined in
-        * does not merge another Flow or FlowBuilder from another module.
 
         The most straightforward way to meet these requirements is to define
         your flow in a module as:
@@ -1526,28 +1524,39 @@ class Flow:
 
         self_name = self.name
 
-        module_names = set()
+        # Find the module that contains the flow.
+        candidate_flow_modules = set()
         for provider in state.providers_by_name.values():
             source_func = provider.get_source_func()
             if source_func is None:
                 continue
-            module_names.add(source_func.__module__)
+            module = module_registry[source_func.__module__]
+            if len(self._get_flows_from_module(module)) > 0:
+                candidate_flow_modules.add(module)
+        if len(candidate_flow_modules) == 0:
+            raise Exception(
+                oneline(f"Couldn't find the module that has flow {self_name!r}.")
+            )
+        if len(candidate_flow_modules) > 1:
+            raise Exception(
+                oneline(
+                    f"""
+                        Too many modules that contain flow {self_name!r},
+                        found: {len(candidate_flow_modules)}, wanted 1"""
+                )
+            )
+        (flow_module,) = candidate_flow_modules
 
+        flow_module = recursive_reload(flow_module)
+
+        flows = self._get_flows_from_module(flow_module)
         blessed_candidate_flows = []
         unblessed_candidate_flows = []
-        for module_name in module_names:
-            module = reload(module_registry[module_name])
-            for key in dir(module):
-                element = getattr(module, key)
-                if not isinstance(element, Flow):
-                    continue
-                flow = element
-                if flow.name != self_name:
-                    continue
-                if not flow._state.is_blessed:
-                    unblessed_candidate_flows.append(flow)
-                else:
-                    blessed_candidate_flows.append(flow)
+        for flow in flows:
+            if not flow._state.is_blessed:
+                unblessed_candidate_flows.append(flow)
+            else:
+                blessed_candidate_flows.append(flow)
 
         if len(blessed_candidate_flows) == 0:
             if len(unblessed_candidate_flows) > 0:
@@ -1563,7 +1572,7 @@ class Flow:
                     oneline(
                         f"""
                     Couldn't find any flow named {self_name!r}
-                    in modules {module_names!r}"""
+                    in module {flow_module.__name__!r}"""
                     )
                 )
         if len(blessed_candidate_flows) > 1:
@@ -1571,7 +1580,7 @@ class Flow:
                 oneline(
                     f"""
                 Too many flows named {self_name!r}
-                in modules {module_names!r};
+                in module {flow_module.__name__!r};
                 found {len(blessed_candidate_flows)}, wanted 1"""
                 )
             )
@@ -1639,6 +1648,18 @@ class Flow:
         builder = FlowBuilder._from_state(self._state)
         builder_update_func(builder)
         return Flow._from_state(builder._state)
+
+    def _get_flows_from_module(self, module):
+        flows = []
+        for key in dir(module):
+            element = getattr(module, key)
+            if not isinstance(element, Flow):
+                continue
+            flow = element
+            if flow.name != self.name:
+                continue
+            flows.append(flow)
+        return flows
 
 
 class ShortcutProxy:
--- a/bionic/utils/reload.py
+++ b/bionic/utils/reload.py
@@ -1,0 +1,45 @@
+import builtins
+import importlib
+from sys import modules as module_registry
+from fnmatch import fnmatch
+from sysconfig import get_paths as sysconfig_paths
+
+
+def recursive_reload(module):
+    """
+    Helper method to reload a module recursively. If a module imports a set of
+    modules, then the modules in the set are also reloaded and so on.
+
+    Modules that are part of the current python installation are not reloaded.
+    For example, modules part of python standard library or modules installed
+    through pip (or some other package manager that use distutils).
+
+    Also note that this method may not be able to handle dynamic imports that
+    only happens at runtime. For example, if a module imports another module
+    only when a certain method is executed, reloading the former module does
+    not guarantee that the latter module is reloaded.
+    """
+
+    original_import = builtins.__import__
+    already_reloaded = set()
+
+    def custom_import(name, globals=None, locals=None, fromlist=[], level=0):
+        if name in module_registry:
+            module = module_registry[name]
+            if name not in already_reloaded and not is_internal_module(module):
+                already_reloaded.add(name)
+                importlib.reload(module)
+        return original_import(name, globals, locals, fromlist, level)
+
+    try:
+        builtins.__import__ = custom_import
+        return importlib.reload(module)
+    finally:
+        builtins.__import__ = original_import
+
+
+def is_internal_module(module):
+    return not hasattr(module, "__file__") or any(
+        fnmatch(module.__file__, file_dir + "/*")
+        for file_dir in sysconfig_paths().values()
+    )
--- a/docs/release-notes.rst
+++ b/docs/release-notes.rst
@@ -99,11 +99,13 @@ Deprecated Features
 Bug Fixes
 .........
 
-- Fixed an `issue <https://github.com/square/bionic/issues/111>`_ where non-persistable
+- Fixed an `issue <https://github.com/square/bionic/issues/111>`__ where non-persistable
   entities could be spuriously recomputed even when their values weren't directly
   needed.
-- Fixed an `issue <https://github.com/square/bionic/issues/229>`_ where caching fails
+- Fixed an `issue <https://github.com/square/bionic/issues/229>`__ where caching fails
   when cache directory paths contain whitespaces.
+- Fixed an `issue <https://github.com/square/bionic/issues/233>`__ where reloading of
+  flows that merge other flows from other modules does not work correctly.
 
 Documentation
 .............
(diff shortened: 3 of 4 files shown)


bionic/module_reloader.py
```diff
@@ -0,0 +1,24 @@
+import builtins
```
jqmp, line 1: super nit: `module_reloader` is a bit of an unusual name for a module; usually they're named after concepts (like `persistence` or classes (like `flow`). In this case, a natural name might be `reloading`.

bionic/module_reloader.py
```diff
@@ -0,0 +1,24 @@
...
+from importlib import reload
+from sys import modules as module_registry
+
+
+def recursive_reload(module):
```
jqmp, line 6: Maybe a docstring would be helpful here?

bionic/module_reloader.py
```diff
@@ -0,0 +1,24 @@
...
+                already_reloaded.add(name)
+                try:
+                    reload(module)
+                except Exception:
+                    pass
+        return original_import(name, globals, locals, fromlist, level)
+
+    builtins.__import__ = custom_import
+    module = reload(module)
+    builtins.__import__ = original_import
```
jqmp, line 23: Should we put this in a `finally` block, to make sure we clean up after ourselves even if `reload` throws an exception?

bionic/module_reloader.py
```diff
@@ -0,0 +1,24 @@
...
+    already_reloaded = set()
+
+    def custom_import(name, globals=None, locals=None, fromlist=[], level=0):
+        if name in module_registry:
+            module = module_registry[name]
+            if name not in already_reloaded and hasattr(module, "__file__"):
+                already_reloaded.add(name)
+                try:
+                    reload(module)
+                except Exception:
```
jqmp, line 17: What exceptions are we catching here? In general, if we're going to swallow all exceptions, we should probably have a comment explaining why.

tests/test_flow/test_reload.py
```diff
@@ -27,6 +27,7 @@ def write_module(tmp_module_path):
     modules = set()
 
     def _write_module(name, contents):
+        # (tmp_module_path / f"{name}.py").write_text(dedent(contents))
```
jqmp, line 30: Leftover comment?

bionic/module_reloader.py
```diff
@@ -0,0 +1,24 @@
...
+from importlib import reload
+from sys import modules as module_registry
+
+
+def recursive_reload(module):
+    original_import = builtins.__import__
+    already_reloaded = set()
+
+    def custom_import(name, globals=None, locals=None, fromlist=[], level=0):
+        if name in module_registry:
```
namanjain, line 11: Before checking `name` in `module_registry`, would you need to change it using `globals` and `locals` when the `level` is relative instead of being absolute?

namanjain: Similarly, when we have submodules in `fromlist`, should we just reload those submodules instead of reloading the entire module?

simonafk: My guess is that `__import__()` will be called on those submodules anyways, and `custom_import` will be called. Hence, there is probably no need to call `reload` on those submodules?

bionic/module_reloader.py
```diff
@@ -0,0 +1,24 @@
...
+            module = module_registry[name]
+            if name not in already_reloaded and hasattr(module, "__file__"):
+                already_reloaded.add(name)
+                try:
+                    reload(module)
+                except Exception:
+                    pass
```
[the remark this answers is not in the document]
simonafk: The goal is to reload as many modules as possible. Some modules simply cannot be reloaded. Added comments here.

bionic/flow.py
```diff
@@ -1540,21 +1538,52 @@ def reloading(self):
...
+                        Couldn't find the module that has flow {self_name}.
+                    """
+                )
+            )
+        (flow_module,) = candidate_flow_modules
+        if len(candidate_flow_modules) > 1:
+            raise Exception(
+                oneline(
+                    f"""
+                        Too many modules that contain flow {self_name},
```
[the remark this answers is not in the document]
namanjain: We talked about this offline, but for posterity, this works since we filter modules to the ones that have a flow with the same name.

bionic/module_reloader.py
namanjain: But if someone imports `a.b` using `from a import b`, would we also reload other submodules of `a` this way (that are not needed / used)?

bionic/utils/reload.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def custom_import(name, globals=None, locals=None, fromlist=[], level=0):
+        if name in module_registry:
+            module = module_registry[name]
+            if name not in already_reloaded and hasattr(module, "__file__"):
+                already_reloaded.add(name)
+                try:
+                    reload(module)
+                except Exception:
+                    # The goal is to reload as many modules as possible. If a
```
jqmp, line 31: Why can some modules not be reloaded? What kind of exception will they throw? What about swallowing legitimate exceptions -- how we will we deal with that?

For example, if the user has a bug in their new module that throws an exception, won't this swallow it?

bionic/utils/reload.py
```diff
@@ -0,0 +1,40 @@
...
+
+def recursive_reload(module):
+    """
+    Helper method to reload a module recursively. If a module imports a set of
+    modules, then the modules in the set are also reloaded and so on.
+
+    Not all modules can be reloaded. This method attempts to reload as many
+    modules as possible.
+
+    Also note that this method may not be able to handle dynamic imports that
```
jqmp, line 14: Nice observation.

bionic/utils/reload.py
```diff
@@ -0,0 +1,40 @@
...
+    only happens at runtime. For example, if a module imports another module
+    only when a certain method is executed, reloading the former module does
+    not guarantee that the latter module is reloaded.
+    """
+
+    original_import = builtins.__import__
+    already_reloaded = set()
+
+    def custom_import(name, globals=None, locals=None, fromlist=[], level=0):
```
jqmp, line 23: This will re-import all the third-party libraries and so on that the root module uses, right? That seems like it could be quite a lot. Do you think it could make sense to avoid reloading things like standard libraries and installed libraries?

bionic/flow.py
```diff
@@ -1526,28 +1524,50 @@ def reloading(self):
...
                 continue
-            module_names.add(source_func.__module__)
-
-        blessed_candidate_flows = []
-        unblessed_candidate_flows = []
-        for module_name in module_names:
-            module = reload(module_registry[module_name])
+            module = module_registry[source_func.__module__]
             for key in dir(module):
                 element = getattr(module, key)
```
jqmp, line 1535: We have this logic in two different places now. Would it make sense to factor it out into a `get_flow_from_module` function?

bionic/utils/reload.py
simonafk: It definitely makes sense. The problem is figuring out which modules are standard libraries and which ones are not. It would be nice if there is a simple heuristics to do this. Right now, there is a `hasattr(module, "__file__")` check that avoids reloading modules that does not contain python source code.

Maybe we can check file timestamp?

bionic/utils/reload.py
simonafk: That is a very good point.

simonafk: Hmm, the test actually passes without the try catch.

bionic/utils/reload.py
jqmp: Could we figure this out from the location of the module files? I think a human being could generally do this by examining the file paths, but I'm not sure exactly what the algorithm should be.

The file timestamp idea is also interesting; that does seem like a good way to filter out modules that don't need to be loaded.

bionic/utils/reload.py
jqmp: I'm curious what errors you originally had in mind when you wrote the try-catch. Are there some modules you don't think will reload properly?

bionic/utils/reload.py
jqmp: BTW, @namanjain do you know if Streamlit has any ideas that might be useful here?

damienrj:
All I have seen is https://github.com/streamlit/streamlit/pull/537

bionic/utils/reload.py
namanjain: Streamlit uses the same logic for reloading as caching. They check that the file should be in the dir where the script is running from or is in PYTHONPATH, and the file should not be blacklisted. This doc is not publicly available, but you can see this logic for caching in the Smart Caching doc [here](https://docs.google.com/document/d/1VxF-GC-Fk8JC-yTr1ACrAIjKBdevsH5sAnNp9DN6nA0/edit#heading=h.7obdz6b2lnx2). Simon and I briefly discussed that we should share that logic between smart caching and reloading, but we never reached a conclusion. I think it would be pretty nice to share that logic though.

You can find how Streamlit filters it [here](https://github.com/streamlit/streamlit/blob/develop/lib/streamlit/watcher/local_sources_watcher.py#L154).

simonafk: Thanks! Checking PYTHONPATH should work.

bionic/utils/reload.py
simonafk: I think some earlier iteration of this method may have failed to reload certain internal modules. Removed the try-catch as it is no longer necessary.

bionic/module_reloader.py
simonafk: I am not entirely sure, but I think the `importlib.reload` works on a per-module basis and hence it may update all variables that are associated with the reloaded module? This is not worse than before since we were also calling `reload`.

bionic/utils/reload.py
simonafk: Using streamlit's `file_in_pythonpath` method to skip over installed modules.
`file_in_pythonpath` uses environment variables and stays constant even if `sys.path` is modified. (The tests modify `sys.path`)

bionic/flow.py
simonafk: Refactored into `_get_flows_from_module`

bionic/utils/reload.py
simonafk: Also updated the comments in the method.

bionic/module_reloader.py
simonafk: Renamed to `reload.py`

bionic/utils/reload.py
jqmp: I'm confused. Installed libraries generally don't go in the PYTHONPATH; in fact, it's normal to run Python without a PYTHONPATH at all. And Streamlit seems to use PYTHONPATH the opposite way: it reloads files that *are* in the PYTHONPATH, right?

simonafk: Ah, I think I have misunderstood how `PYTHONPATH` works. Running python or pytest from command line shows that `PYTHONPATH` can be empty.

simonafk: I have reverted the changes. How about letting this method reload every module for now, while we figure out a solution on how to reduce the reload separately? Reloading everything is the simplest approach, is guaranteed to work and solves the main problem.

jqmp: That's true, but it also feels like a bit of a regression from the previous behavior. Is there something simple we could do, like not reload files that are within a separate `site-packages` directory? That seems like it might catch most installed libraries.

simonafk: Added the sysconfig-based solution that I have been looking into. Also added a test to check that internal modules are not reloaded.

bionic/utils/reload.py
```diff
@@ -0,0 +1,45 @@
...
+        return original_import(name, globals, locals, fromlist, level)
+
+    try:
+        builtins.__import__ = custom_import
+        return importlib.reload(module)
+    finally:
+        builtins.__import__ = original_import
+
+
+def is_internal_module(module):
```
jqmp, line 41: This looks pretty good! I wasn't familiar with `sysconfig`. Anything we should be worried about?

tests/test_flow/test_reload.py
```diff
@@ -280,3 +281,242 @@ def r():
...
+    write_builder_module(3)
+    flow_tester.reload_flow()
+    flow_tester.expect_values(x=3, x_plus_1=4, y=10, total=13)
+
+
+def test_internal_modules_not_reloaded(write_module, flow_tester):
+    write_module("test_module", "")
+    flow_tester.write_flow_module(
+        """
+            import sys
```
jqmp, line 495: Could we also test a third-party library, like `pandas`?

jqmp (approved):
This looks good to me! But don't forget to add something to the release notes as well.

bionic/utils/reload.py
simonafk: My initial concern was about whether this will still work in a different platform and package manager (e.g. Anaconda). From reading https://docs.python.org/3/library/sysconfig.html#installation-paths and looking into `distutils`, I think it should all work fine.

bionic/utils/file_util.py
```diff
@@ -0,0 +1,60 @@
+# Copyright 2018-2020 Streamlit Inc.
```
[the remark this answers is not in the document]
simonafk: This file has been removed as I am no longer using the `PYTHONPATH` approach.

tests/test_flow/test_reload.py
simonafk: Good idea

docs/release-notes.rst
```diff
@@ -99,11 +99,13 @@ Deprecated Features
...
 
-- Fixed an `issue <https://github.com/square/bionic/issues/111>`_ where non-persistable
+- Fixed an `issue <https://github.com/square/bionic/issues/111>`__ where non-persistable
   entities could be spuriously recomputed even when their values weren't directly
   needed.
-- Fixed an `issue <https://github.com/square/bionic/issues/229>`_ where caching fails
+- Fixed an `issue <https://github.com/square/bionic/issues/229>`__ where caching fails
   when cache directory paths contain whitespaces.
+- Fixed an `issue <https://github.com/square/bionic/issues/233>`__ where reloading of
+  flows that merge other flows from other modules do not work correctly.
```
jqmp, line 108: s/do not/does not/

PR #247, opened 2020-09-08
Garbage-collect entry-memoized values

jqmp:
Currently, if a value isn't set to `@persist` or `@memoize`, it's still
memoized (at the TaskRunnerEntry level) for the duration of the `get`
call. This isn't a big deal because it's pretty rare to disable both
persistence and memoization for an entity. However, once we start using
tuple descriptors, those tuples will have both persistence and
memoization disabled *and* may contain large values that we don't want
to keep in memory.

In particular, for any entity defined like this:

    @builder
    @bn.memoize(False)
    @bn.outputs('a', 'b')
    def _(...):
        ...

while `a` and `b` will be correctly persisted and not memoized, the
tuple `a, b` *will* be memoized for the duration of the `get` call. If
`a` or `b` is large, this may be a problem for users.

To address this, I've added logic to detect when these entry-memoized
values are no longer needed and delete them. This should allow us to
enable tuple descriptors without any memory-usage regressions, and also
gives us some memory savings in the rare case where both persistence and
memoization are manually disabled.

This is implemented by allowing requirements to be discarded when
they're no longer needed. The criteria for discarding a requirement
depend on which of three types it falls into, as documented on the
EntryRequirements class. This system of state and requirements is
getting a little rickety again. I'm open to ideas for simplifying it; or
maybe it'll be easier after we've done of the future refactoring

### DIFF
--- a/bionic/core/flow_execution.py
+++ b/bionic/core/flow_execution.py
@@ -78,9 +78,7 @@ class TaskCompletionRunner:
             for state in states:
                 state.set_up_caching_flags(self._bootstrap)
                 entry = self._get_or_create_entry_for_state(state)
-                self._add_requirement(
-                    src_entry=None, dst_entry=entry, level=EntryLevel.CACHED
-                )
+                self._add_permanent_requirement(entry, EntryLevel.CACHED)
 
             while self._has_pending_entries():
                 entry = self._activate_next_pending_entry()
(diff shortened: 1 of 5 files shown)


jqmp:
This PR doesn't work the way I expect when combined with https://github.com/square/bionic/pull/150, so I suspect there's a bug in here somewhere.

jqmp:
Update: I think this code will prematurely garbage-collect values in some cases. I'm going to close the PR until I figure it out.

PR #252, opened 2020-09-10
Refactor Tasks to have a single TaskKey

jqmp:
Until now each Task could have multiple TaskKeys, which were originally
used to support functions that returned mulitiple entity values (using
`@outputs`). However, this functionality is now implemented using tuple
descriptors, allowing one TaskKey to represent multiple entity values,
so all Tasks now have exactly one TaskKey. This commit removes support
for multiple TaskKeys, which simplifies a lot of things.

There should be no functional changes.

### DIFF
--- a/bionic/datatypes.py
+++ b/bionic/datatypes.py
@@ -45,8 +45,8 @@ class Task:
 
     Attributes
     ----------
-    keys: list of TaskKeys
-        Keys corresponding to the output values computed by this task.
+    key: TaskKey
+        Key corresponding to the output value computed by this task.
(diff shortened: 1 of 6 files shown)


bionic/datatypes.py
```diff
@@ -45,8 +45,8 @@ class Task:
 
     Attributes
     ----------
-    keys: list of TaskKeys
-        Keys corresponding to the output values computed by this task.
+    key: TaskKeys
```
namanjain, line 48: nit: TaskKey~s~

namanjain (approved):
so good

PR #253, opened 2020-09-10
Add AIP based execution

namanjain:
This change adds the capability in Bionic to use AIP based execution.
Unlike parallel execution, not all tasks are computed in parallel.
Instead, you specify which entities to compute on AIP using the
`@aip_process` decorator.

This change builds on top of #249 and uses the `aip` package from that
change. Since it is still a PoC, we haven't documented this behavior or
added it to the release notes.

I also ran the new test using AIP (instead of the test implementation)
and it worked!
Screen Shot 2020-09-10 at 4 31 31 PM

Note that stackdriver logging is not implemented yet. I'll add that in
a separate PR.

### DIFF
--- a/bionic/aip/task.py
+++ b/bionic/aip/task.py
@@ -11,10 +11,12 @@ from bionic.aip.future import Future
 
 
 @attr.s(auto_attribs=True)
-class Resource:
-    """Description of a computing resource, which can be a single machine or a cluster of machines
+class TaskConfig:
+    """
+    Contains configuration that can differ per task, which can be a
+    single machine or a cluster of machines.
 
-    In the future could be extended to support GPUs and similar
+    In the future could be extended to support GPUs and similar.
     """
 
     machine: str
@@ -23,8 +25,9 @@ class Resource:
 
 
 @attr.s(auto_attribs=True)
-class Job:
-    """A job groups together individual tasks, all of which share these properties
+class Config:
+    """
+    Contains configuration that remains the same across all tasks.
 
     Attributes
     ----------
@@ -34,7 +37,6 @@ class Job:
         The full address on gcr of the docker image for the tasks
     project: str
         The GCP project where the jobs will be run
-
     """
 
     uuid: str
@@ -47,32 +49,34 @@ class Job:
 @attr.s(auto_attribs=True)
 class Task:
     name: str
-    job: Job
     function: Callable
-    resource: Resource
+    config: Config
+    task_config: TaskConfig
 
     def job_id(self):
-        return f"{self.job.uuid}_{self.name}"
+        return f"{self.config.uuid}_{self.name}"
 
     def inputs_uri(self):
-        # In a future version it might be better to make this path specified by the flow,
-        # so that it can be inside a GCS cache location and different file types
-        return f"gs://{self.job.project}/bionic/{self.job.uuid}/{self.name}-inputs.cloudpickle"
+        # In a future version it might be better to make this path
+        # specified by the flow, so that it can be inside a GCS cache
+        # location and different file types.
+        return f"gs://{self.config.project}/bionic/{self.config.uuid}/{self.name}-inputs.cloudpickle"
 
     def output_uri(self):
-        # In a future version it might be better to make this path specified by the flow,
-        # so that it can be inside a GCS cache location and different file types
-        return f"gs://{self.job.project}/bionic/{self.job.uuid}/{self.name}-output.cloudpickle"
+        # In a future version it might be better to make this path
+        # specified by the flow, so that it can be inside a GCS cache
+        # location and different file types.
+        return f"gs://{self.config.project}/bionic/{self.config.uuid}/{self.name}-output.cloudpickle"
 
     def _ai_platform_job_spec(self):
         """Conversion from our task data model to a job request on ai platform"""
         output = {
-            "jobId": f"{self.job.uuid}_{self.name}",
-            "labels": {"job": self.job.uuid},
+            "jobId": f"{self.config.uuid}_{self.name}",
+            "labels": {"job": self.config.uuid},
             "trainingInput": {
-                "serviceAccount": self.job.account,
-                "masterType": self.resource.machine,
-                "masterConfig": {"imageUri": self.job.image_uri},
+                "serviceAccount": self.config.account,
+                "masterType": self.task_config.machine,
+                "masterConfig": {"imageUri": self.config.image_uri},
                 "args": ["python", "-m", "bionic.aip.task", self.inputs_uri()],
                 "packageUris": [],
                 "region": "us-west1",
@@ -80,47 +84,52 @@ class Task:
                 "scaleTier": "CUSTOM",
             },
         }
-        if self.job.network is not None:
-            output["trainingInput"]["network"] = self.job.network
+        if self.config.network is not None:
+            output["trainingInput"]["network"] = self.config.network
 
-        if self.resource.worker_count is not None and self.resource.worker_count > 0:
-            output["trainingInput"]["workerCount"] = self.resource.worker_count
-            output["trainingInput"]["workerType"] = self.resource.worker_machine
-            output["trainingInput"]["workerConfig"] = {"imageUri": self.job.image_uri}
+        if (
+            self.task_config.worker_count is not None
+            and self.task_config.worker_count > 0
+        ):
+            output["trainingInput"]["workerCount"] = self.task_config.worker_count
+            output["trainingInput"]["workerType"] = self.task_config.worker_machine
+            output["trainingInput"]["workerConfig"] = {
+                "imageUri": self.config.image_uri
+            }
 
         return output
 
+    def _stage(self):
+        # Scope the blocks import to this function to avoid raising for anyone not using it.
+        blocks = import_optional_dependency("blocks", raise_on_missing=True)
+        cloudpickle = import_optional_dependency("cloudpickle", raise_on_missing=True)
 
-def _stage(task):
-    # Scope the blocks import to this function to avoid raising for anyone not using it.
-    blocks = import_optional_dependency("blocks", raise_on_missing=True)
-    cloudpickle = import_optional_dependency("cloudpickle", raise_on_missing=True)
+        path = self.inputs_uri()
+        logging.info(f"Staging task {self.name} at {path}")
+        with blocks.filesystem.GCSNativeFileSystem().open(path, "wb") as f:
+            cloudpickle.dump(self, f)
 
-    path = task.inputs_uri()
-    logging.info(f"Staging task {task.name} at {path}")
-    with blocks.filesystem.GCSNativeFileSystem().open(path, "wb") as f:
-        cloudpickle.dump(task, f)
+    def submit(self) -> Future:
+        # Scope the import to this function to avoid raising for anyone not using it.
+        discovery = import_optional_dependency(
+            "googleapiclient.discovery", raise_on_missing=True
+        )
 
+        self._stage()
+        spec = self._ai_platform_job_spec()
 
-def submit(task) -> Future:
-    # Scope the import to this function to avoid raising for anyone not using it.
-    discovery = import_optional_dependency(
-        "googleapiclient.discovery", raise_on_missing=True
-    )
+        aip = discovery.build("ml", "v1", cache_discovery=False)
+        logging.info(f"Submitting {self.config.project}: {self}")
 
-    task._stage()
-    spec = task._ai_platform_job_spec()
-
-    aip = discovery.build("ml", "v1", cache_discovery=False)
-    logging.info(f"Submitting {task.job.project}: {task}")
-
-    request = (
-        aip.projects().jobs().create(body=spec, parent=f"projects/{task.job.project}")
-    )
-    request.execute()
-    url = f'https://console.cloud.google.com/ai-platform/jobs/{spec["jobId"]}'
-    logging.info(f"Started task on AI Platform: {url}")
-    return Future(task.job.project, task.job_id(), task.output_uri())
+        request = (
+            aip.projects()
+            .jobs()
+            .create(body=spec, parent=f"projects/{self.config.project}")
+        )
+        request.execute()
+        url = f'https://console.cloud.google.com/ai-platform/jobs/{spec["jobId"]}'
+        logging.info(f"Started task on AI Platform: {url}")
+        return Future(self.config.project, self.job_id(), self.output_uri())
 
 
 def run():
--- a/bionic/core/flow_execution.py
+++ b/bionic/core/flow_execution.py
@@ -5,6 +5,7 @@ to completion.
 
 from concurrent.futures import wait, FIRST_COMPLETED
 import logging
+import time
 
 from .task_execution import (
     EntryLevel,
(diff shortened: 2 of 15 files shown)


bionic/core/flow_execution.py
```diff
@@ -327,7 +340,15 @@ def _wait_on_in_progress_entries(self):
         futures = set(
             entry.future for entry in self._in_progress_entries_by_task_key.values()
         )
-        finished_futures, _ = wait(futures, return_when=FIRST_COMPLETED)
+        if self._parallel_execution_enabled:
+            finished_futures, _ = wait(futures, return_when=FIRST_COMPLETED)
+        else:
+            assert self._aip_execution_enabled
+            # TODO: This works for PoC, but maybe there is a better way
```
jqmp, line 347: Yeah, longer-term, maybe a cleaner way is to put each job in its own thread and wrap that thread in a regular concurrent future, so we can keep using `wait` here (and hopefully mix both parallel and AIP futures).

bionic/decorators.py
```diff
@@ -440,6 +441,31 @@ def returns(out_descriptor):
     )
 
 
+def aip_resource(resource):
```
jqmp, line 444: Let's add the same disclaimer as we have on `accepts` and `returns`:

    This decorator is currently experimental and does not have any additional
    user-facing documentation. It may change in non-backwards-compatible ways.

bionic/decorators.py
```diff
@@ -440,6 +441,31 @@ def returns(out_descriptor):
...
+    """
+    Indicates that the decorated function should be computed in AIP.
+    This decorator requires AIP based distributed execution to be enabled, which
+    can be done by setting ``core__aip_execution__enabled`` core entity.
+
+    Parameters
+    ----------
+    resource: bionic.aip.task.Resource
+        Computing resource that the entity function should use on AIP.
```
jqmp, line 453: nit: "The machine type that should be used to compute the function on AIP."

bionic/decorators.py
```diff
@@ -440,6 +441,31 @@ def returns(out_descriptor):
...
 
+def aip_resource(resource):
+    """
+    Indicates that the decorated function should be computed in AIP.
+    This decorator requires AIP based distributed execution to be enabled, which
+    can be done by setting ``core__aip_execution__enabled`` core entity.
+
+    Parameters
+    ----------
+    resource: bionic.aip.task.Resource
```
jqmp, line 452: Assuming users aren't going to use the `Resource` class for anything else, would it make more sense to just accept  arguments like `machine_type` and so on directly instead of making them instantiate a `Resource`?

bionic/executor.py
```diff
@@ -6,19 +6,83 @@
...
+        """
+        pass
+
+    def stop_logging(self):
+        """
+        Stops any logging work in the current process that is not needed anymore.
+        """
+        pass
+
+    def set_resource(self, resource):
```
jqmp, line 49: Hmmm.... it's not clear to me why this base class is better than just having different APIs for the different executor types. If AipExecutor requires a resource to run something remotely, why not just make that part of the API?

bionic/executor.py
```diff
@@ -6,19 +6,83 @@
...
+            # an alphabet and only accepts alphanumeric and underscore
+            # characters.
+            name="a" + str(uuid4()).replace("-", ""),
+            job=self._aip_job,
+            resource=self._resource,
+            function=partial(fn, *args, **kwargs),
+        ).submit()
+
+    def create_synchronized_set(self):
+        return SynchronizedSet()
```
jqmp, line 79: Does this work on AIP?

bionic/executor.py
```diff
@@ -6,19 +6,83 @@
...
+        self._aip_job = aip_job
+
+    def submit(self, fn, *args, **kwargs):
+        assert self._resource is not None
+        return Task(
+            # TODO: Use a better identifiable name, maybe a combination
+            # of entity name and case key.
+            # This is a temporary workaround. We are changing the random
+            # UUID in such a manner because AIP names have to start with
+            # an alphabet and only accepts alphanumeric and underscore
```
jqmp, line 70: s/an alphabet/a letter/

tests/test_flow/conftest.py
```diff
@@ -105,6 +116,16 @@ def gcs_builder(builder, tmp_gcs_url_prefix):
     return builder
 
 
+@pytest.fixture(scope="function")
+def aip_builder(gcs_builder, aip_execution_enabled):
+    assert aip_execution_enabled
+
+    gcs_builder.set("core__aip_execution__enabled", "True")
+    gcs_builder.set("core__aip_execution__gcp_project", "sq-mlf-dev")
```
jqmp, line 124: Should we pass this on the command line like we do with `--bucket`?

tests/test_flow/conftest.py
```diff
@@ -15,17 +15,28 @@
...
-# debugging and much clearer than parameterizing `parallel_execution_enabled`
-# which suffixes [TRUE] / [FALSE].
+# like test_name[PARALLEL], test_name[SERIAL], and test_name[AIP]. This is
+# super helpful while debugging and much clearer than parameterizing
+# `parallel_execution_enabled` which suffixes [TRUE] / [FALSE].
 @pytest.fixture(
     params=[
         pytest.param("serial", marks=pytest.mark.serial),
         pytest.param("parallel", marks=pytest.mark.parallel),
+        pytest.param("aip", marks=pytest.mark.aip),
```
jqmp, line 25: What's the purpose of doing this with parametrization if every AIP test is explicitly marked as such? It's not clear to me why we're doing this differently from how we do it with GCS access.

bionic/flow.py
```diff
@@ -1817,12 +1819,62 @@ def core__persistent_cache(
...
+        core__aip_execution__docker_image,
+    ):
+        if not core__aip_execution__enabled:
+            return None
+        if core__aip_execution__gcp_project is None:
+            raise AssertionError(
+                "core__aip_execution__gcp_project is None, " "but needs a value"
+            )
+        return Job(
+            # TODO Revisit uuid.
```
jqmp, line 1848: Revisit it how?

bionic/flow.py
```diff
@@ -1817,12 +1819,62 @@ def core__persistent_cache(
...
+    def core__aip_execution__job(
+        core__aip_execution__enabled,
+        core__aip_execution__gcp_project,
+        core__aip_execution__docker_image,
+    ):
+        if not core__aip_execution__enabled:
+            return None
+        if core__aip_execution__gcp_project is None:
+            raise AssertionError(
+                "core__aip_execution__gcp_project is None, " "but needs a value"
```
jqmp, line 1845: Should this be one single string?

Also, why can we explain why it needs a value?

bionic/flow.py
```diff
@@ -1817,12 +1819,62 @@ def core__persistent_cache(
...
         core__parallel_execution__worker_count,
     ):
         if not core__parallel_execution__enabled:
             return None
-        return Executor(core__parallel_execution__worker_count)
+        return ProcessExecutor(core__parallel_execution__worker_count)
+
+    builder.assign("core__aip_execution__enabled", False, persist=False)
+    builder.assign("core__aip_execution__gcp_project", None, persist=False)
+    builder.assign("core__aip_execution__docker_image", "bionic:latest", persist=False)
```
jqmp, line 1832: Is this image publicly accessible, or just within Square?

bionic/aip/task.py
```diff
@@ -101,16 +99,13 @@ def _ai_platform_job_spec(self):
...
-            output["trainingInput"]["hyperparameters"] = self.tune.spec()
-
         return output
 
     def submit(self) -> Future:
         self._stage()
         spec = self._ai_platform_job_spec()
 
-        if os.environ.get("TEST", "").lower() == "true":
+        if "PYTEST_CURRENT_TEST" in os.environ:
```
jqmp, line 108: I'm a little confused at how these tests are intended to work. If we're substituting a `_TestFuture` here, are we still hitting AIP in our tests?

Also, as I noted [here](https://github.com/square/bionic/pull/249/files#r486491423), I'd much rather avoid this kind of "are we in a test" check altogether.

bionic/utils/misc.py
```diff
@@ -365,6 +365,21 @@ def __init__(self):
         self.values = set()
         self.lock = threading.Lock()
 
+    def __getstate__(self):
```
jqmp, line 368: Does this actually work the way we'd expect, or is this just a hack to avoid throwing an exception?

tests/test_flow/conftest.py
```diff
@@ -105,6 +116,16 @@ def gcs_builder(builder, tmp_gcs_url_prefix):
     return builder
 
 
+@pytest.fixture(scope="function")
+def aip_builder(gcs_builder, aip_execution_enabled):
+    assert aip_execution_enabled
+
+    gcs_builder.set("core__aip_execution__enabled", "True")
```
jqmp, line 123: Should `"True"` be a string?

tests/test_flow/test_persistence_aip.py
```diff
@@ -0,0 +1,30 @@
+import pytest
+
+import bionic as bn
+
+
+# This is detected by pytest and applied to all the tests in this module.
+pytestmark = [pytest.mark.needs_aip, pytest.mark.needs_gcs]
+
+
+def test_aip_jobs(aip_builder):
```
jqmp, line 10: Is there a way we can test that this actually uses AIP?

bionic/executor.py
namanjain: I thought this abstraction made the caller code easier, but that is only when we use only one kind of executor, which means either parallel or aip (not both). But if we support both parallel and aip execution on the same flow, then this abstraction doesn't help in any way. Then it would be easier to do what you suggested, have different APIs for different executor types.

bionic/executor.py
namanjain: Depends on what you mean by "work". AIP would get the synchronized set, but the main process that's sending the remote jobs won't get back the updates on the set. As we discussed before, progress reporting would be broken with this PoC. 

I can add this as a code comment too.

bionic/aip/task.py
namanjain: Right, these tests don't hit AIP. It runs the task in a subprocess instead. I had to change this condition to actually run the test on AIP.

Re "are we in a test" check, this is part of my second commit (tagged with fixup) and I'm going to remove that commit altogether once we fix this in the base PR (the one with the linked comment).

tests/test_flow/conftest.py
namanjain: You're right, this was a mistake.

tests/test_flow/test_persistence_aip.py
namanjain: Once we have a different task implementation of test, we can run this test using both the implementations. We can run the test implementation by default, and run the AIP implementation on our local machines using some kind of flag. Like I mentioned in an earlier [reply](https://github.com/square/bionic/pull/253#discussion_r487122003), I'm waiting for the fix in the base PR, once that's done, I'll change the test to do as I described above.

I guess I could've made it clear in the description or waited till the base PR had the fixes in. My bad.

bionic/flow.py
namanjain: Ah right, this is the image you create using the dockerfile in the base PR. Maybe we should have instructions for people on how to create the docker image.

For reference, these are the commands I use:
```
docker build -t bionic:latest .
docker tag bionic:latest gcr.io/sq-mlf-dev/bionic:latest
docker push gcr.io/sq-mlf-dev/bionic:latest
```

bionic/flow.py
```diff
@@ -1817,12 +1819,62 @@ def core__persistent_cache(
...
+    @decorators.immediate
+    def core__executor(
+        core__aip_execution__enabled,
+        core__parallel_execution__enabled,
+        core__aip_executor,
+        core__process_executor,
+    ):
+        if core__aip_execution__enabled and core__parallel_execution__enabled:
+            raise AssertionError(
+                "Both AIP and parallel execution enabled, but only one can be enabled."
```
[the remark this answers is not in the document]
namanjain: Yes, we should remove it later. I added this for PoC since it makes the waiting logic a lot simpler, because we can keep using [concurrent.futures.wait](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.wait) for parallel processing. Added a TODO for it.

bionic/decorators.py
namanjain: I think I would need to change this anyway. @baxen this seems related to [your comment](https://github.com/square/bionic/pull/249#discussion_r488070772) on your recent PR. What fields would you want configurable? I initially thought we would only want to configure the resource, but it seems like I got it wrong.

bionic/flow.py
namanjain: I modeled this after `core__persistent_cache__cloud_store` to be consistent, but I can make it a single string. I'll add an explanation, but let's wait till we resolve the [decorator question](https://github.com/square/bionic/pull/253#discussion_r487072298). Depending on its result, we might not even create this here.

bionic/decorators.py
jqmp: What seems most natural to me is for `@aip_resource` to take a `machine_type` directly, and have the job config be set either via global config entities or (maybe, optionally) via a separate `@aip_job` decorator. This seems equally composable (you can always instantiate a decorator and reuse it in your code) but requires less boilerplate for the common case. @namanjain @baxen WDYT?

bionic/flow.py
jqmp: I'm guessing the string for `core__persistent_cache__cloud_store` used to be on separate lines, but they got moved by `black`. AFAICT there's no reason for it to use separate strings.

bionic/core/flow_execution.py
```diff
@@ -439,10 +478,15 @@ class TaskKeyLogger:
     def __init__(self, bootstrap):
         self._level = logging.INFO if bootstrap is not None else logging.DEBUG
 
-        executor = bootstrap.executor if bootstrap is not None else None
+        executor = bootstrap.process_executor if bootstrap is not None else None
         if executor is not None:
             self._already_logged_task_key_set = executor.create_synchronized_set()
         else:
+            # AIP would get the synchronized set, but the main process
```
jqmp, line 485: I'm not sure how to read this comment. The use of "would" suggests that we're talking about something hypothetical, but this seems like it's describing what will actually happen (in particular case).

It sounds like we're saying that some stuff won't work if AIP is enable. If so, let's be precise about
1. What circumstances will cause something weird to happen (AIP being enabled?)
2. Exactly what will happen (everything will work except that progress reporting will not include stuff happening on AIP?)
3. Whether this is something we need to come back and fix (TODO?)

bionic/executor.py
```diff
@@ -6,19 +6,49 @@
...
+            # UUID in such a manner because AIP names have to start with
+            # a letter and only accepts alphanumeric and underscore
+            # characters.
+            name="a" + str(uuid4()).replace("-", ""),
+            job=self._aip_job,
+            resource=self._resource,
+            function=partial(fn, *args, **kwargs),
+        ).submit()
+
+    def set_resource(self, resource):
```
jqmp, line 47: If we're using different interfaces for the two executors now, does it still make sense to have a separate `set_resource` method?

bionic/flow.py
jqmp: Bump?

jqmp (approved):
Aside from some nits around comments and the decorator UI, looks good to me!

bionic/decorators.py
namanjain: We talked offline about this. We'll continue keeping the global config the way it is, and explode the class as separate fields on this decorator. This is experimental anyway, so it won't hurt to try one way and change it before we GA.

bionic/flow.py
namanjain: Changed the prefix to use flow name.

PR #254, opened 2020-09-14
Refactor providers to output single dnode

jqmp:
This is a followup to #252; now that Tasks only have single-key outputs,
we can do the same thing for Providers.

There should be no functional changes.


namanjain (approved):
LGTM

simonafk (approved):

PR #255, opened 2020-09-16
Fix assisted versioning for non-persisted values

jqmp:
Looks like we had a bug where assisted versioning would not catch
versioning errors on non-persisted entities. This adds a fix by allowing
bytecode data to propagate through `Provenance.exact_hash`. This change
actually simplifies the matching model (instead of having four match
levels, we have three and they all propagate the same way), but it makes
it more challenging to report the source of the error when we find it,
since it could be somewhere upstream of where it's detected. To fix
*this*, each Provenance now has references to its dependency digests,
and each digest now has a reference to its provenance if it was created
from one, allowing us to search backward for the provenance with the
problem.

This also requires making ProvenanceDigests serializable as dicts, which
is getting a bit gross. I'm wondering if we can use a library like
`cattrs` to replace our bespoke object<->dict<->YAML serialization.


bionic/persistence.py
```diff
@@ -645,14 +635,6 @@ def _nominal_metadata_url_prefix_for_query(self, query):
             + minor_version_token
```
namanjain, line 635: now that we have a nominal hash, how do you feel about replacing minor version with `query.provenance.nominal_hash`? That way, you won't bleed the logic of how provenance hashes work in url creation. Although, it would also mean that our urls will be even longer.

bionic/core/task_execution.py
```diff
@@ -701,25 +701,65 @@ def _check_accessor_for_version_problems(self):
...
+                assert False
+
+            try:
+                locate_mismatched_provenances_and_raise(old_prov, new_prov)
+            except AssertionError as e:
+                message = """
+                Unable to complete assisted versioning check due to a bug in Bionic;
+                please report this to the developers
+                """
+                logger.warn(oneline(message), exc_info=e)
```
namanjain, line 762: Should this be an error instead? If there is a bug, the users cannot rely on assisted versioning anyway. Plus this warning is going to spam the users since we refresh the cache accessors a few times too (and all the descendents would log this error too).

bionic/persistence.py
```diff
@@ -602,16 +602,6 @@ def _find_best_match(self, query):
                 level="exact",
             )
 
-        samecode_url_prefix = self._samecode_metadata_url_prefix_for_query(query)
-        samecode_urls = [
-            url for url in equivalent_urls if url.startswith(samecode_url_prefix)
-        ]
-        if len(samecode_urls) > 0:
```
namanjain, line 609: For clarity, this wasn't doing anything in the current code, right? I don't see any place in Bionic where we create a file with `samecode_url_prefix`.

namanjain (approved):
Nice catch! Looks good to me once comments are addressed.

bionic/persistence.py
jqmp: Yeah, that seems like a good idea. It does give up a tiny bit of human-readable information, but it's more consistent and less coupled to the exact definition of `nominal_hash`.

bionic/core/task_execution.py
jqmp: Good question. I see your point, but here was my thinking:

1. Assisted versioning is already best-effort; users can't rely on it to be 100% accurate. Also, it's just a backstop to catch you if you forget to manually update a version.
2. If this does happen, there's probably still a versioning error; we just can't figure out exactly where it is. (Maybe we should make this more clear in the message.)
3. With the warning, the user has the option of filtering the warning and continuing; with an exception they will have to just give up.

Maybe I could expand the message to explain the user's options here (try to fix the error, clear the cache, or just stop using assisted versioning)?

bionic/persistence.py
jqmp: Yes, this seems to be vestigial. I actually have a separate commit for removing these unused bits, but it's not very clear from the PR message.

jqmp:
Just realized I need release notes too.

bionic/core/task_execution.py
namanjain: The updated message looks good!

PR #256, opened 2020-09-17
Refactor cache YAML serialization to use cattrs

jqmp:
This refactors our YAML serialization code to use `cattrs` instead of a
bespoke system of converting to and from dicts.

With our old system, every YAML-serializable class
(ArtifactMetadataRecord, Provenance, and ProvenanceDigest) had special
code to convert them to and from plain dicts, which could then be easily
serialized into YAML. This was awkward and meant these classes couldn't
refer to other classes unless those classes were also convertible to
dicts. However, now we use `attrs` to define many of our classes, and
any `attrs` class can be automatically (well, almost) converted to/from
dicts using the `cattrs` library.

This simplifies our code a lot, although it also introduce coupling
between some internal data model classes and our serialization format;
if any of the former change, we'll need to update our cache schema as
well.  However, if this becomes a problem we can always separate those
classes into "live" and "serialized" forms.


jqmp:
I added a workaround to make this work on Python 3.6.

bionic/persistence.py
```diff
@@ -1037,113 +1005,49 @@ def from_computation(
...
+                code_fingerprint.version.minor,
+            ]
         )
         exact_hash = hash_simple_obj_to_hex(
-            dict(
-                code=exact_code_dict,
-                deps=exact_deps_list,
-            )
+            [
+                nominal_hash,
```
simonafk, line 1035: Should this be `functional_hash`?

jqmp: The goal is for each hash to include all the information from the previous hashes -- so if `nominal_hash` changes, we want `exact_hash` to change as well. I can add a comment to make this more explicit.

simonafk: never mind, misread this.

simonafk (approved):

bionic/persistence.py
```diff
@@ -1027,6 +983,18 @@ class Provenance:
     must be a versioning error.
     """
 
+    descriptor: str = attr.ib()
+    # We don't use this for anything; it's just included to allow humans to distinguish
```
namanjain, line 987: This comment applies only to `case_key_elements`, right? If so, can you add a newline before `dep_digests`? Otherwise, it seems like the comment applies to both the fields with the newlines separating other fields.

bionic/persistence.py
```diff
@@ -1196,15 +1105,77 @@ def from_value_hash(cls, value_hash):
...
 
+
+# This converts the field Provenance.dep_digests from a forward reference to a real
+# type; if we don't do this, cattr will fail to handle it properly. (This could also be
+# resolved by adding a structure hook for ForwardReference('ProvenanceDigest'), but
+# that isn't available in Python 3.6. It can be accessed as _ForwardReference, but that
+# seemed to cause the interpreter to abort unpredictably in our tests so I didn't want
+# to use it. Similarly, we can use `from __future__ import annotations` to remove
+# the need forward references, but that also isn't available in 3.6.)
+attr.resolve_types(Provenance)
```
namanjain, line 1116: I'm glad at least this works!

namanjain (approved):
The code looks much cleaner with this. The only real question I have is whether we can use `frozen=True` for `Provenance`, but otherwise, LGTM!

bionic/datatypes.py
```diff
@@ -256,9 +261,9 @@ class CodeFingerprint:
     A collection of characteristics attempting to uniquely identify a function.
     """
 
-    version = attr.ib()
-    bytecode_hash = attr.ib()
-    orig_flow_name = attr.ib()
+    version: CodeVersion = attr.ib()
```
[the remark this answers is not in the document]
jqmp: Yep, `cattrs` depends on them to know how to deserialize these classes.

bionic/persistence.py
jqmp: Yep, good point!

bionic/persistence.py
jqmp: Me too!

bionic/persistence.py
```diff
@@ -924,65 +928,17 @@ def valid_filename_from_query(query):
...
-            Dumper=YamlDumper,
-        )
-
-    def __repr__(self):
-        return f"ArtifactMetadataRecord({self.descriptor!r})"
-
-
-# TODO Could we use something like the cattr library instead of our own custom dict
-# stuff?
+# We don't use frozen=True because Provenance has list fields, which are not hashable.
```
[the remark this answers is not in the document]
jqmp: You're right, `frozen=True` will still work when the fields aren't hashable, although it will make `attrs` define a hash method for you. (The method will still fail, but with a slightly more confusing error message than the default.) I guess I think of `frozen` as suggesting that the object should be hashable.

Overall I don't have strong feeling about whether to use `frozen=True` here; WDYT?

jqmp:
Note to self: I think `attr.resolve_types` is a pretty recent feature, so we should require the appropriate version of `attrs`.

bionic/persistence.py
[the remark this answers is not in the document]
namanjain: Ah, I see. I saw the documentation which said hash is only generated when both `frozen` and `eq` is set to `True`. What I didn't realize was that default `None` value for `eq` is I guess treated as `True`. 

I don't have strong feeling about it either. You can leave it as it is.

PR #257, opened 2020-09-21
Support versioning modes in random tests

jqmp:
This extends the new random test system to support the `assist` and
`auto` versioning modes, bringing us much closer to being able to get
rid of the old fuzz tests. (These new tests also would have caught the
failure of assisted versioning to work on non-persisted values,
presumably.)

This required the following supporting changes:
- CodeVersioningError now has a `bad_descriptor` field which identifies
the function that needs to be fixed.
- We now define each function in the flow under test by generating
Python code (`output_value_fragment`) instead of just calling out to a
separate function; this makes changes in the definition visible in the
Python bytecode.
- In the model flow, persisted values can be marked "stale", which will
cause them to trigger a CodeVersioningError.

### DIFF
--- a/tests/test_flow/test_persistence_random.py
+++ b/tests/test_flow/test_persistence_random.py
@@ -8,6 +8,7 @@ import re
 
 from bionic.descriptors import ast
 from bionic.descriptors.parsing import dnode_from_descriptor
+from bionic.exception import CodeVersioningError
 from bionic.utils.misc import single_element, single_unique_element
 import bionic as bn
 
@@ -47,23 +48,29 @@ class ModelFlowHarness:
 
     def __init__(self, builder, make_list):
         self._builder = builder
+        self._versioning_mode = "manual"
 
         self._entities_by_name = {}
 
         self._descriptors_computed_by_flow = make_list()
         self._descriptors_computed_by_model = make_list()
 
+    def set_versioning_mode(self, mode):
+        assert mode in ("manual", "assist", "auto")
+        self._versioning_mode = mode
+        self._builder.set("core__versioning_mode", mode)
+
     def get_all_entity_names(self):
         return list(self._entities_by_name.keys())
 
-    def create_entity(self, should_persist):
+    def create_entity(self, should_persist=True):
         name = f"e{len(self._entities_by_name) + 1}"
         entity = ModelEntity(name=name, should_persist=should_persist)
         assert name not in self._entities_by_name
         self._entities_by_name[name] = entity
         return name
 
-    def add_binding(self, out_descriptor, dep_entity_names, use_tuples_for_output):
+    def add_binding(self, out_descriptor, dep_entity_names, use_tuples_for_output=True):
         out_dnode = dnode_from_descriptor(out_descriptor)
         out_entities = list(
             map(self._entities_by_name.get, out_dnode.all_entity_names())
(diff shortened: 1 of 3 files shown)


tests/test_flow/test_persistence_random.py
```diff
@@ -194,21 +259,77 @@ class ModelBinding:
...
+    annotated_func_version = attr.ib(default=1)
 
     has_persisted_values = attr.ib(default=False)
     persisted_values_by_entity_name = attr.ib(default=None)
+    persisted_values_stale_ancestor = attr.ib(default=None)
+
     some_descendant_might_have_persisted_values = attr.ib(default=False)
+    some_descendant_might_have_fresh_persisted_values = attr.ib(default=False)
+
+    def update_func_version(self, change_func, update_annotation, versioning_mode):
```
namanjain, line 271: nit: This name is a little confusing now that don't always change the func or the annotation. Can we call this `update` or `update_binding`?

tests/test_flow/test_persistence_random.py
```diff
@@ -87,57 +94,105 @@ def add_binding(self, out_descriptor, dep_entity_names, use_tuples_for_output):
...
+        except CodeVersioningError as e:
+            model_exception = e
+
+        assert (flow_exception is None) == (model_exception is None)
+
+        if model_exception is None:
+            assert flow_value == model_value
+
+        else:
+            assert flow_exception.__class__ == model_exception.__class__
```
namanjain, line 133: Can you assert that the bad descriptors are also equal?

tests/test_flow/test_persistence_random.py
```diff
@@ -159,6 +214,9 @@ def _compute_model_value(self, entity_name, memoized_values_by_entity_name=None)
             self._descriptors_computed_by_model.append(binding.out_descriptor)
 
         else:
+            if binding.persisted_values_stale_ancestor is not None:
+                stale_ancestor = binding.persisted_values_stale_ancestor
```
namanjain, line 218: Can we also assert that the versioning mode is assisted?

tests/test_flow/test_persistence_random.py
```diff
@@ -179,8 +237,15 @@ class ModelEntity:
     binding = attr.ib(default=None)
     dependent_entities = attr.ib(default=attr.Factory(list))
 
-    def compute_value(self, dep_values):
-        return f"{self.name}.{self.binding.func_version}({','.join(dep_values)})"
+    def compute_value(self, func_version, dep_values):
```
namanjain, line 240: nit: Do you need to pass the func version here? I think it was cleaner when it was reading from binding directly.

tests/test_flow/test_persistence_random.py
```diff
@@ -194,21 +259,77 @@ class ModelBinding:
...
     use_tuples_for_output = attr.ib()
     func_version = attr.ib(default=1)
+    annotated_func_version = attr.ib(default=1)
 
     has_persisted_values = attr.ib(default=False)
     persisted_values_by_entity_name = attr.ib(default=None)
+    persisted_values_stale_ancestor = attr.ib(default=None)
+
     some_descendant_might_have_persisted_values = attr.ib(default=False)
+    some_descendant_might_have_fresh_persisted_values = attr.ib(default=False)
```
namanjain, line 269: I think it's clearer to say `stale_persisted_values` and flip the flag. The names are getting wordy but I don't have a better suggestion either.

namanjain (approved):
The random test code is getting a little complicated, but the functionality looks great!

tests/test_flow/test_persistence_random.py
```diff
@@ -87,57 +94,105 @@ def add_binding(self, out_descriptor, dep_entity_names, use_tuples_for_output):
...
+
+        assert (flow_exception is None) == (model_exception is None)
+
+        if model_exception is None:
+            assert flow_value == model_value
+
+        else:
+            assert flow_exception.__class__ == model_exception.__class__
+
+            # Now we konw that a CodeVersioningError has been thrown, we'll try to get
```
jqmp, line 135: s/konw/know/

tests/test_flow/test_persistence_random.py
jqmp: I think it's possible for there to be multiple possible bad descriptors, and which one we get will depend on the order in which we traverse the dependencies. I can add a comment making this clear.

(In theory, after we make all the fixes we could assert that the two sets of bad descriptors are the same, but I'm not sure it's worth reorganizing the code to make this possible.)

tests/test_flow/test_persistence_random.py
jqmp: Nope, there's no good reason for this. This is left over from a different approach. Thanks!

tests/test_flow/test_persistence_random.py
jqmp: So you'd call it `no_descendant_can_have_stale_persisted_values`? In that case I'd probably want to flip both of them; I'll try that and see if it looks better.

PR #258, opened 2020-09-21
Write AIP log entries to Stackdriver

namanjain:
This change writes the log entries from the AIP execution to
Stackdriver. It uses the google-cloud-logging library to write those
logs, similar to how Conductor (a Square internal library) does it. The
only difference is that we don't use environment variables for the
`job_id` or the `project`.

### DIFF
--- a/bionic/aip/main.py
+++ b/bionic/aip/main.py
@@ -1,0 +1,78 @@
+"""
+This module is run as main in order to execute a task on a worker.
+"""
+
+import logging
+import os
+import sys
+
+from bionic.deps.optdep import import_optional_dependency
+
+
+def run():
+    # Scope the import to this function to avoid raising for anyone not using it.
+    blocks = import_optional_dependency("blocks")
+    cloudpickle = import_optional_dependency("cloudpickle")
+
+    ipath = sys.argv[-1]
+    fs = blocks.filesystem.GCSNativeFileSystem()
+    with fs.open(ipath, "rb") as f:
+        task = cloudpickle.load(f)
+
+    # Now that we have the task, set up logging.
+    _set_up_logging(task.job_id(), task.config.project_name)
+    logging.info(f"Read task from {ipath}")
+
+    result = task.function()
+
+    opath = task.output_uri()
+    logging.info(f"Uploading result to {opath}")
+    blocks.pickle(result, opath)
+    return
+
+
+def _set_up_logging(job_id, project_id):
+    # TODO This is the ID of the hyperparameter tuning trial currently
+    # running on this VM. This field is only set if the current
+    # training job is a hyperparameter tuning job. Conductor uses this
+    # environment variable but AIP documentation suggests us to use
+    # TF_CONFIG. Check whether we need to update this env variable.
+    # Find more details on TF_CONFIG at this link:
+    # https://cloud.google.com/ai-platform/training/docs/distributed-training-details
+    trial_id = os.environ.get("CLOUD_ML_TRIAL_ID", None)
+
+    glogging = import_optional_dependency("google.cloud.logging")
+
+    client = glogging.Client(project=project_id)
+    resource = glogging.resource.Resource(
+        type="ml_job",
+        # AIP expects a default task_name for the master cluster. We
+        # use a placeholder value till we start using clusters. Once we
+        # do, it should be configured based on the cluster.
+        labels=dict(job_id=job_id, project_id=project_id, task_name="master-replica-0"),
+    )
+    labels = None
+    if trial_id is not None:
+        # Enable grouping by trial when present.
+        labels = {"ml.googleapis.com/trial_id": trial_id}
+
+    # Enable only the cloud logger to avoid duplicate messages.
+    handler = glogging.handlers.handlers.CloudLoggingHandler(
+        client, resource=resource, labels=labels
+    )
+    root_logger = logging.getLogger()
+    # Remote the StreamHandler. Any logs logged by it shows up as error
+    # logs in Stackdriver.
+    root_logger.handlers = []
+    # We should ideally make this configurable, but till then, let's
+    # set the level to DEBUG to write all the logs. It's not hard to
+    # filter using log level on Stackdriver so it doesn't create too
+    # much noise anyway.
+    root_logger.setLevel(logging.DEBUG)
+    root_logger.addHandler(handler)
+    for logger_name in glogging.handlers.handlers.EXCLUDED_LOGGER_DEFAULTS:
+        logging.getLogger(logger_name).propagate = False
+
+
+if __name__ == "__main__":
+    run()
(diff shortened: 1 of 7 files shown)


bionic/aip/main.py
```diff
@@ -0,0 +1,70 @@
...
+    # Scope the import to this function to avoid raising for anyone not using it.
+    blocks = import_optional_dependency("blocks", raise_on_missing=True)
+    cloudpickle = import_optional_dependency("cloudpickle", raise_on_missing=True)
+
+    ipath = sys.argv[-1]
+    fs = blocks.filesystem.GCSNativeFileSystem()
+    with fs.open(ipath, "rb") as f:
+        task = cloudpickle.load(f)
+
+    # Now that we have the task, setup logging.
```
jqmp, line 22: nit: I know this is from Conductor, but "setup" is a noun. The verb is "set up".

bionic/aip/main.py
```diff
@@ -0,0 +1,70 @@
...
+    logging.info("Starting execution")
+    result = task.function()
+
+    opath = task.output_uri()
+    logging.info(f"Uploading result to {opath}")
+    blocks.pickle(result, opath)
+    return
+
+
+def _setup_logging(job_id, project):
```
jqmp, line 35: nit: s/setup/set_up/

bionic/aip/main.py
```diff
@@ -0,0 +1,70 @@
...
+import logging
+import os
+import sys
+
+from bionic.deps.optdep import import_optional_dependency
+
+
+def run():
+    # Scope the import to this function to avoid raising for anyone not using it.
+    blocks = import_optional_dependency("blocks", raise_on_missing=True)
```
jqmp, line 14: WDYT about omitting the `raise_on_missing` parameter in these calls?

bionic/aip/main.py
jqmp, line 35: nit: Should `project` be `project_id`? Likewise for `trial`/`trial_id`?

bionic/aip/main.py
```diff
@@ -0,0 +1,70 @@
...
+    # Find more details on TF_CONFIG at this link:
+    # https://cloud.google.com/ai-platform/training/docs/distributed-training-details
+    trial = os.environ.get("CLOUD_ML_TRIAL_ID", None)
+
+    glogging = import_optional_dependency("google.cloud.logging", raise_on_missing=True)
+
+    client = glogging.Client(project=project)
+    resource = glogging.resource.Resource(
+        type="ml_job",
+        labels=dict(job_id=job_id, project_id=project, task_name="master-replica-0"),
```
jqmp, line 50: Where does `master-replica-0` come from?

bionic/aip/main.py
```diff
@@ -0,0 +1,70 @@
...
+    )
+    labels = None
+    if trial is not None:
+        # Enable the grouping by trial when present.
```
jqmp, line 54: nit: s/the//

bionic/aip/main.py
```diff
@@ -0,0 +1,70 @@
...
+        labels = {"ml.googleapis.com/trial_id": trial}
+
+    # Enable only the cloud logger to avoid duplicate messages.
+    handler = glogging.handlers.handlers.CloudLoggingHandler(
+        client, resource=resource, labels=labels
+    )
+    root_logger = logging.getLogger()
+    root_logger.handlers = []  # remove stream handler
+    root_logger.setLevel(logging.DEBUG)
```
jqmp, line 63: Hmmm... can we provide some more explanation of why setting this to `DEBUG` is good/necessary?

bionic/aip/main.py
```diff
@@ -0,0 +1,70 @@
...
+    root_logger.addHandler(handler)
+    for logger_name in glogging.handlers.handlers.EXCLUDED_LOGGER_DEFAULTS:
```
jqmp, line 65: What does this do?

jqmp (approved):
I have some nits and questions, but in general this looks pretty plausible!

bionic/aip/main.py
namanjain: I can't believe I missed this. Nice catch!

bionic/aip/main.py
```diff
@@ -0,0 +1,70 @@
...
+    cloudpickle = import_optional_dependency("cloudpickle", raise_on_missing=True)
+
+    ipath = sys.argv[-1]
+    fs = blocks.filesystem.GCSNativeFileSystem()
+    with fs.open(ipath, "rb") as f:
+        task = cloudpickle.load(f)
+
+    # Now that we have the task, setup logging.
+    _setup_logging(task.job_id(), task.config.project)
+    logging.info(f"Read task from {ipath}")
```
simonafk, line 24: This log seems redundant since we are printing another log immediately after?

bionic/aip/main.py
namanjain: ah, I didn't realize that the default is `True`. I'll remove it from everywhere in AIP.

bionic/aip/main.py
namanjain: We can rename `trial` to `trial_id`. For `project`, I think the more appropriate name would be `project_name` (instead of id). I'll rename this too.

namanjain: Oh but the google cloud logging `Resource` calls it `project_id`, so I guess I'll just call it `project_id` in this method but `project_name` elsewhere. I'm calling it `project_name` elsewhere because that's what the google API calls it (check this [create API](https://cloud.google.com/ai-platform/training/docs/reference/rest/v1/projects.jobs/create) that takes project).

namanjain: Their naming is all over the place honestly. While making this change, I found out that they also call it `project` in logging `Client`. I'll keep it `project_name`, but again wanted to point out how bad AIP documentation and naming is everywhere.

jqmp: https://www.youtube.com/watch?v=uaehB3PJfr0

bionic/aip/main.py
namanjain: This excludes loggers that the library uses itself to avoid infinite recursion. We do the same thing here as what the library provided [setup_logging](https://googleapis.dev/python/logging/1.14.0/_modules/google/cloud/logging/handlers/handlers.html) function does, but we also remove the StreamHandler as well to avoid those error logs (I added a comment in the code for this). Hence, we have this implementation here instead of using their provided method.

bionic/aip/main.py
namanjain: Sure, I can remove the next log for clarity.

bionic/aip/main.py
namanjain: It turns out that we do need to fill this with a default value because AIP expects one. This is just a placeholder for, till we start using clusters. I added a comment in the code explaining this too. Till then, I'm just gonna keep this placeholder value.

PR #259, opened 2020-09-22
Add fake AIP tests

namanjain:
This change adds a fake AIP test that executes AIP tasks using
subprocesses instead of dispatching them to AIP. Like the real AIP
tasks, the fake still uses GCS for caching and needs a bucket to run.

### DIFF
--- a/bionic/aip/main.py
+++ b/bionic/aip/main.py
@@ -32,6 +32,9 @@ def run():
 
 
 def _set_up_logging(job_id, project_id):
+    if os.environ.get("BIONIC_NO_STACKDRIVER", False):
+        return
+
     # TODO This is the ID of the hyperparameter tuning trial currently
     # running on this VM. This field is only set if the current
     # training job is a hyperparameter tuning job. Conductor uses this
--- a/tests/test_flow/conftest.py
+++ b/tests/test_flow/conftest.py
@@ -11,6 +11,7 @@ from ..helpers import (
     ResettingCallCounter,
 )
 
+from .fakes import FakeAipExecutor
 import bionic as bn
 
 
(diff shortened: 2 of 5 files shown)


tests/test_flow/conftest.py
```diff
@@ -105,6 +106,27 @@ def gcs_builder(builder, tmp_gcs_url_prefix):
...
+def fake_aip_builder(gcs_builder, gcs_url_stem):
+    gcs_builder.set("core__aip_execution__enabled", True)
+    # TODO: This is really a hack to make fake AIP tests work. We need
+    # to  specify a real project for input and output uri for AIP task.
+    # In future, we should change this so that the executor specifies
+    # the input and output uris instead. It could also be indicative of
+    # the fact that maybe we should use the gcs bucket for input and
+    # output uri instead of the AIP project (and use project only for
+    # AIP jobs).
+    project = gcs_url_stem[5:]
```
jqmp, line 119: Maybe we could `assert gcs_url_stem.startswith('gs://')` to make this clearer?

bionic/aip/main.py
```diff
@@ -32,6 +32,10 @@ def run():
 
 
 def _set_up_logging(job_id, project_id):
+    # We don't want to set up Stackdriver logging for local tests.
+    if os.environ.get("PYTEST_CURRENT_TEST", False):
```
jqmp, line 36: Having our real code check whether we're in a test seems hacky to me. Could we just have an explicit env variable or command line argument that configures whether to use stackdriver logging?

bionic/executor.py
```diff
@@ -44,6 +44,31 @@ def submit(self, task_config, fn, *args, **kwargs):
         ).submit()
 
 
+class FakeAipExecutor:
```
jqmp, line 47: This PR adds fake code across three different non-test files. Could we just put all this fake AIP stuff into one file? Possibly in the test directory?

Also, do we actually need a `FakeTask` class, or can we just inline the code from `submit`?

tests/test_flow/conftest.py
namanjain: I initially thought the same thing, but then saw that `gcs_url_stem` fixture already does that. But I can add it again for readability.

bionic/executor.py
namanjain: I'll move all the fake code in a fakes test file. 

I thought about `FakeTask` class more. I think it would be nice to keep `FakeTask` that inherits `Task` so that we can use the `Task._stage` method.

tests/test_flow/fakes.py
```diff
@@ -0,0 +1,54 @@
+from functools import partial
+import logging
+from uuid import uuid4
+
+from bionic.aip.future import Future as AipFuture
+from bionic.aip.task import Task as AipTask
+
+
+class FakeAipFuture(AipFuture):
```
jqmp, line 9: nit: docstring?

jqmp (approved):

PR #260, opened 2020-09-29
Introduce draft descriptors

jqmp:
This PR includes several small, self-contained preliminary commits, then one big commit with the major changes. It's probably easiest to read them separately, but I've put them all in one PR because the preliminary commits don't do anything useful on their own.

### DIFF
--- a/bionic/core/flow_execution.py
+++ b/bionic/core/flow_execution.py
@@ -3,6 +3,7 @@ This module contains the logic to execute tasks and their dependencies
 to completion.
 """
 
+import attr
 from concurrent.futures import wait, FIRST_COMPLETED
 import logging
 import time
@@ -15,6 +16,7 @@ from .task_execution import (
     RemoteSubgraph,
     TaskRunnerEntry,
 )
+from ..descriptors import ast
 from ..utils.keyed_priority_stack import KeyedPriorityStack
 from ..utils.misc import SynchronizedSet
 
@@ -472,7 +474,9 @@ class TaskKeyLogger:
 
         executor = bootstrap.process_executor if bootstrap is not None else None
         if executor is not None:
-            self._already_logged_task_key_set = executor.create_synchronized_set()
+            self._already_logged_entity_case_key_pairs = (
+                executor.create_synchronized_set()
+            )
         else:
             # TODO: When AIP execution is enabled, we will send this
             # SynchronizedSet to AIP jobs for logging task keys. AIP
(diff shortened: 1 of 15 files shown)


jqmp:
As mentioned in the commit message, the only user-visible change should be a slight difference in how DAGs are visualized (note the extra angle brackets around the tuples):
image
However, if you want to see the extra hidden nodes introduced by this change, you can use `flow.render_dag(_include_detail=True)`:
image

bionic/core/flow_execution.py
```diff
@@ -162,12 +164,15 @@ def _process_active_entry(self, entry):
...
-        if can_compute_remotely:
+        entry_can_be_computed_remotely = (
+            aip_execution_enabled_for_entry or self._parallel_execution_enabled
+        ) and entry.state.should_persist
+        if entry_can_be_computed_remotely:
+            remote_subgraph = RemoteSubgraph(entry.state, self._bootstrap)
+            if not remote_subgraph.all_states_can_be_serialized:
+                entry_can_be_computed_remotely = False
+
+        if entry_can_be_computed_remotely:
```
simonafk, line 175: Did you mean to remove this? The block above also has `entry_can_be_computed_remotely`

jqmp: I think this is what I intended; maybe it's confusing that the block above modifies `entry_can_be_computed_remotely`? The idea is that we need to meet all the initial conditions for remote computation, and then we need to also check that the subgraph is serializable. Maybe instead of modifying `entry_can_be_computed_remotely`, it would be better to introduce a new variable?

bionic/deriver.py
```diff
@@ -88,11 +97,31 @@ def export_dag(self, include_core=False):
...
+        # the same, but not when a user function returns a more complex descriptor.) For
+        # other nodes, we "collapse" them: we delete the node and connect each of its
+        # parents to each of its children.
+        def should_collapse_dnode(dnode):
+            if not include_core:
+                descriptor_includes_non_core = any(
+                    not entity_is_internal(entity_name)
+                    for entity_name in dnode.all_entity_names()
+                )
+                if not descriptor_includes_non_core:
```
namanjain, line 111: This means that all entities are core (or internal). I think you can simplify this by using `all` and removing `not`s (both inside `any` and outside here).

bionic/deriver.py
```diff
@@ -88,11 +97,31 @@ def export_dag(self, include_core=False):
...
+                    return True
+
+            if _include_detail:
+                return False
+
             if isinstance(dnode, ast.EntityNode):
-                return not entity_is_internal(dnode.to_entity_name())
+                return False
+
+            if isinstance(dnode, ast.DraftNode) and not isinstance(
```
namanjain, line 120: We will probably have non draft descriptors later once we allow users to fetch tuples, but this helped me understand that tuple descriptors are always draft descriptors in the current implementation. I think it would be good to document this somewhere, it took me some time to understand all this. Wdyt?

namanjain: Nevermind me, I see that you mentioned it in the comment on line 271.

bionic/deriver.py
```diff
@@ -147,41 +184,112 @@ def _register_static_providers(self):
...
+            else:
+                # The only remaining node type is an entity node, and there's no
+                # additional work for us to do on those.
+                assert isinstance(dnode, ast.EntityNode)
 
         for provider in set(self._flow_state.providers_by_name.values()):
             register_provider(provider)
 
         self._static_providers_by_dnode = static_providers_by_dnode
+        self._followup_dnode_lists_by_dnode = dict(followup_dnode_lists_by_dnode)
```
namanjain, line 292: Why do you need to cast this to a dict again?

bionic/deriver.py
```diff
@@ -281,12 +396,27 @@ def _obtain_entity_def_for_dnode(self, dnode):
...
+        elif isinstance(dnode, ast.DraftNode):
+            child_entity_def = self._obtain_entity_def_for_dnode(dnode.child)
+            doc_prefix = "(Intermediate value)"
+            if child_entity_def.doc is None:
+                doc = doc_prefix
+            else:
+                doc = doc_prefix + " " + child_entity_def.doc
+            return EntityDefinition(
+                name=dnode.to_descriptor(),
+                protocol=NonSerializableObjectProtocol(),
```
namanjain, line 414: nit: how do you feel about using a global instance for this protocol?

bionic/core/flow_execution.py
namanjain: I think adding a new variable would be cleaner than modifying the existing one.

namanjain (approved):
I particularly like how the follow up tasks are more generalized, but the edit method on descriptors gave me a pause. I think the change is a bit complicated and it took a little while to sink in, but it looks great overall!

bionic/deriver.py
jqmp: I just wanted to avoid any possible surprising behavior downstream, since other parts of the code may not realize that this is a `defaultdict`, not a `dict`.  (I don't anticipate any specific problems here, since the object's behavior will make sense either way, but it still seems safer to me.) I can add a comment.

bionic/deriver.py
jqmp: I don't feel strongly either way. Happy to change it.

bionic/deriver.py
jqmp: You're right, this looks nicer!

PR #263, opened 2020-10-01
AIP test that runs locally using fakes

simonafk:
Add fake AIP and GCS for testing of AIP locally and in CI.

### DIFF
--- a/bionic/gcs.py
+++ b/bionic/gcs.py
@@ -101,22 +101,35 @@ class GcsTool:
         obj_prefix = self._validated_object_name_from_url(url_prefix)
         return self._bucket.list_blobs(prefix=obj_prefix)
 
-    def gsutil_cp(self, src_url, dst_url):
-        args = [
-            "gsutil",
-            "-q",  # Don't log anything but errors.
-            "-m",  # Transfer files in paralle.
-            "cp",
-            "-r",  # Recursively sync sub-directories.
-            src_url,
-            dst_url,
-        ]
-        logger.debug("Running command: %s" % " ".join(args))
-        subprocess.check_call(args)
-        logger.debug("Finished running gsutil")
+    @staticmethod
+    def gsutil_cp(src_url, dst_url):
+        gsutil_cp(src_url, dst_url)
 
     def _validated_object_name_from_url(self, url):
         bucket_name, object_name = bucket_and_object_names_from_gs_url(url)
         assert bucket_name == self._bucket.name
         assert object_name.startswith(self._object_prefix)
         return object_name
+
+
+def gsutil_cp(src_url, dst_url):
+    """
+    This method is a proxy for _gsutil_cp which does the actual work. The proxy
+    exists so that _gsutil_cp can be replaced for testing.
+    """
+    _gsutil_cp(str(src_url), str(dst_url))
+
+
+def _gsutil_cp(src_url, dst_url):
+    args = [
+        "gsutil",
+        "-q",  # Don't log anything but errors.
+        "-m",  # Transfer files in parallel.
+        "cp",
+        "-r",  # Recursively sync sub-directories.
+        src_url,
+        dst_url,
+    ]
+    logger.debug("Running command: %s" % " ".join(args))
+    subprocess.check_call(args)
+    logger.debug("Finished running gsutil")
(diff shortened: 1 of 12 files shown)


jqmp (commented):
Being able to run AIP tests on CI will be a big improvement! A couple questions (not necessarily for this PR):

1. Can we apply this to our other GCS tests?
2. Would it make sense to have tests specifically to compare our fakes to the real GCS?

tests/test_flow/conftest.py
```diff
@@ -184,3 +165,58 @@ def tmp_gcs_url_prefix(session_tmp_gcs_url_prefix, request):
...
+    mock_gcs_client.get_bucket = FakeBucket
+
+    with patch("blocks.pickle") as mock_pickle, patch(
+        "blocks.unpickle"
+    ) as mock_unpickle, patch("blocks.filesystem.GCSNativeFileSystem") as mock_gcs_fs:
+        mock_gcs_fs().open = open_blob
+        mock_pickle.side_effect = block_pickle
+        mock_unpickle.side_effect = block_unpickle
+        try:
+            gcs._cached_gcs_client = mock_gcs_client
```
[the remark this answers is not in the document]
simonafk: Now using pytest parameters to configure whether to use fake or real GCP. All GCP / AIP tests are run twice; once locally using fake GCP, and another actually using GCP (skipped if the --project and/or --bucket parameters are not provided).

tests/test_flow/fakes.py
```diff
@@ -56,3 +65,164 @@ def submit(self, task_config, fn, *args, **kwargs):
...
+    def block_pickle(self, result, path):
+        self.blobs[path] = result
+
+    def block_unpickle(self, path):
+        return self.blobs[path]
+
+    def gsutil_cp(self, src_url, dst_url):
+        """
+        Emulate gsutil with the cp command.
+        """
```
simonafk, line 149: Turns out emulating `gsutil cp` is not straightforward. There are slightly different logic depending on whether the source / destination url is a file or directory.

jqmp (approved):
Aside from the question of global mocks vs injection, this looks reasonable to me. Does this mock out all of our existing GCS tests?

tests/conftest.py
```diff
@@ -90,6 +93,9 @@ def pytest_collection_modifyitems(config, items):
                 if "no_parallel" in item.keywords or not also_run_parallel:
                     continue
 
+            if "fake_gcp" in item.keywords:
```
namanjain, line 96: Why skip fake gcp tests for parallel execution?

bionic/gcs.py
```diff
@@ -101,22 +101,35 @@ def blobs_matching_url_prefix(self, url_prefix):
...
-            "-m",  # Transfer files in paralle.
-            "cp",
-            "-r",  # Recursively sync sub-directories.
-            src_url,
-            dst_url,
-        ]
-        logger.debug("Running command: %s" % " ".join(args))
-        subprocess.check_call(args)
-        logger.debug("Finished running gsutil")
+    @staticmethod
```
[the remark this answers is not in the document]
namanjain: If it's not too much work, I think creating a `Gsutil` type class would certainly make this better. I think that class would be equivalent to your `FakeGCS` class. But I agree that it might not be worth it if it's too much, given our inclination to move to fsspec.

tests/conftest.py
simonafk: In "parallel execution", bionic executes parts of the flow in a separate process. The fakes / mocks do not work in a separate process.

bionic/gcs.py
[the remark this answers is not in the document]
simonafk: Yeah, dependency injection may be neater, but global mocks are much easier to set up.

simonafk:
> Aside from the question of global mocks vs injection, this looks reasonable to me. Does this mock out all of our existing GCS tests?

Yeah, this should work with all non-parallel GCS tests.

tests/conftest.py
namanjain: Would it work if we use dependency injection instead of global mocks? If it does, I think it would be worth investing in it now. @jqmp wdyt?

simonafk: Hmm, that is a good point. But I am thinking of investigating FSSPEC first before looking into refactoring all the GCS calls to make it injectable.

jqmp: I agree that we should either use FSSPEC (if feasible) or use injection (if it will make the parallel tests work). I don't feel strongly about which one we should try first.

PR #265, opened 2020-10-06
Improve ergonomics of matching on DescriptorNodes

jqmp:
There are two commits here:

1. Add convenience methods to `DescriptorNode` to make it easier to write code that depends on what subtype you're dealing with. Basically I want to remove all the `isinstance` checks we have now.

2. Add a `flake8` plugin to catch situations where you're trying to exhaustively check all subtypes, but you miss one.

### DIFF
--- a/bionic/deriver.py
+++ b/bionic/deriver.py
@@ -121,12 +121,10 @@ class EntityDeriver:
             if _include_detail:
                 return False
 
-            if isinstance(dnode, ast.EntityNode):
+            if dnode.is_entity():
                 return False
 
-            if isinstance(dnode, ast.DraftNode) and not isinstance(
-                dnode.child, ast.EntityNode
-            ):
+            if dnode.is_draft() and not dnode.child.is_entity():
                 return False
 
             return True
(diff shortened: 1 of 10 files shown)


bionic/descriptors/ast.py
```diff
@@ -18,39 +18,22 @@
...
-#
-#    dnode.is_entity()
-
-# 3. Methods to assert that a node has a particular type:
-#
-#    name = dnode.as_entity().name
+from ..utils.misc import oneline
+
+
+class Subtype(Enum):
```
simonafk, line 29: Wondering about the naming, what about just calling it `Type`? `Subtype` seems to imply there is a `Type`.

jqmp: Yeah, good question. I was reluctant to call it a `Type` because it's not literally a Python type -- I feel like it would be misleading to have a `dnode.type` field whose value isn't actually a `type` object. I agree that "subtype" is not a great name either. I guess we could call the class `DescriptorNodeType` and make the field `dnode.type_enum`. WDYT?

jqmp:
Based on a conversation with @namanjain, I added a few extra comments to `check_dnode_match.py` to make it a little clearer what's going on.

bionic/flake8/check_dnode_match.py
```diff
@@ -0,0 +1,220 @@
...
+
+import attr
+
+from ..descriptors.ast import Subtype
+from ..utils.misc import oneline
+
+
+EXPECTED_TESTED_METHOD_NAMES = {f"is_{subtype.value}" for subtype in Subtype}
+N_EXPECTED_METHODS = len(EXPECTED_TESTED_METHOD_NAMES)
+EXPECATION_MSG = ""
```
namanjain, line 38: Do you need this?

namanjain (approved):
LGTM pending a couple of nits

bionic/flake8/check_dnode_match.py
jqmp: Nope, looks like it should have been deleted.

bionic/deriver.py
```diff
@@ -117,12 +117,10 @@ def should_collapse_dnode(dnode):
             if _include_detail:
                 return False
 
-            if isinstance(dnode, ast.EntityNode):
+            if dnode.is_entity():
```
simonafk, line 120: Wondering about the style: what about doing something like `if dnode.subtype() == ENTITY` ?

jqmp: That would work too, but you'd have to import `Subtype` and then write `if dnode.subtype == Subtype.ENTITY`, which I think is more annoying. (Note that Python doesn't have static imports so if you want to import just `ENTITY`, you need to separately do `ENTITY = Subtype.ENTITY`, and repeat it for each subtype.)

Python doesn't really have a great idiom for doing this kind of pattern matching (until [PEP 622](https://www.python.org/dev/peps/pep-0622/), maybe) so all of the options are a little clunky.

PR #267, opened 2020-10-07
Shard tests

jqmp:
I've separated our tests into "baseline" (the ones that run by default) and "non-baseline" (the ones activated by command line args), and I've sharded the non-baseline tests so that each one only runs on one Python version. This reduces the total latency of our CI by about 15 minutes.

More details in commit messages.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -46,6 +46,9 @@ def pytest_configure(config):
     add_mark("serial", "will run using serial execution")
     add_mark("parallel", "will run using parallel execution")
 
+    # This marker is added automatically based on other markers.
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -46,6 +46,9 @@ def add_mark(name, description):
     add_mark("serial", "will run using serial execution")
     add_mark("parallel", "will run using parallel execution")
 
+    # This marker is added automatically based on other markers
```
namanjain, line 49: super nit: missing full stop

namanjain (approved):

PR #268, opened 2020-10-07
Remove Query class

jqmp:
I've removed the Query class: everything that was previously using the
class now uses just the subset of components (TaskKey, Provenance,
Protocol) that it needs. In many cases, just the Provenance is
sufficient.

The primary motivation here is that I want to pull the logic for
serializing/deserializing values to/from disk out of the PersistentCache
class in order to have separate file descriptor nodes, at which point
the cache won't care about protocols and it won't make sense for them to
be part of its API. More generally, it was always a little weird for to
be passing a protocol around as part of a query, since it's only needed
in one place.

A secondary advantage is that this frees up the term "query" to be used
for something else -- maybe we can use "query" to describe one call to
`flow.get`? (As in, "this value is cached for the duration of the
query".)

### DIFF
--- a/bionic/persistence.py
+++ b/bionic/persistence.py
@@ -57,24 +57,23 @@ except AttributeError:
 
 class PersistentCache:
     """
-    Provides a persistent mapping between Queries (things we could compute) and
-    saved Results (computed Queries).  You use it by getting a CacheAccessor
-    for your specific query, and then performing load/save operations on the
+    Provides a persistent mapping between Provenances (things we could compute) and
+    saved Results (computed Provenances).  You use it by getting a CacheAccessor
+    for your specific Provenance, and then performing load/save operations on the
     accessor.
 
-    When looking up a Query, the cache searches for a saved artifact with a
-    matching Query.  The Query may not match exactly: each Query contains a
-    Provenance, which represents all the code and data used to compute a value,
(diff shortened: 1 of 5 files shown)


bionic/persistence.py
```diff
@@ -57,24 +57,24 @@
...
-    When looking up a Query, the cache searches for a saved artifact with a
-    matching Query.  The Query may not match exactly: each Query contains a
-    Provenance, which represents all the code and data used to compute a value,
-    and two Provenances can match at different levels of precision, from a
-    "functional" match to an "exact" one.  A functional match is sufficient to
-    treat two artifacts as interchangeable; the finer levels of matching are
-    only used by the "assisted versioning" system, which tries to detect
-    situations where a function's bytecode has changed but its version hasn't.
+    When looking up a Provenance, the cache searches for a saved artifact with a
+    matching Provenance. The Provenance may not match exactly: each which represents
```
namanjain, line 66: It seems like something is missing in this sentence. I think you need to replace "each which" with "each Provenance"?

namanjain (approved):

bionic/persistence.py
jqmp: Thanks, good catch. I've changed the sentence structure.

PR #270, opened 2020-10-07
Avoid redundant value hash

jqmp:
Removes some unnecessary hashing. This also removes some git-hook-inflicted damage from our cache test vectors; we need a better way to keep this from happening in the future.

See commit messages for more details.


namanjain (approved):
Nice catch!

PR #272, opened 2020-10-08
Bump version: 0.8.3 → 0.8.4

namanjain:
Screenshot of new release notes (everything looks correct)

Screen Shot 2020-10-08 at 2 23 00 PM

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.8.3
+current_version = 0.8.4
 commit = True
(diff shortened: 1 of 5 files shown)


jqmp (commented):
Since this change breaks cache compatibility, it should be a minor release (0.9.0), right?

namanjain:
> Since this change breaks cache compatibility, it should be a minor release (0.9.0), right?

I totally missed in release notes that cache compatibility change requires a minor release. I'll decline this and open a new PR for minor release.

PR #274, opened 2020-10-09
Fix AIP dispatch

jqmp:
This fixes a bug introduced by my earlier draft descriptors change.
When determining whether to send a task to AIP or not, we need to
consider not just whether that task has an AIP config, but also if anything
in its remote subgraph has one.

I don't have the real AIP tests working on my machine, but hopefully we can
get the fake AIP tests to the point where they can catch this bug.

There are also two small preliminary commits before the main fix.

### DIFF
--- a/bionic/core/flow_execution.py
+++ b/bionic/core/flow_execution.py
@@ -15,8 +15,9 @@ from .task_execution import (
     RemoteSubgraph,
     TaskRunnerEntry,
 )
+from ..exception import AttributeValidationError
 from ..utils.keyed_priority_stack import KeyedPriorityStack
-from ..utils.misc import SynchronizedSet
+from ..utils.misc import oneline, SynchronizedSet
 
 
 # TODO At some point it might be good to have the option of Bionic handling its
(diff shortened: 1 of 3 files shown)


bionic/core/flow_execution.py
```diff
@@ -67,11 +68,15 @@ def __init__(self, bootstrap, flow_instance_uuid, task_key_logger):
 
     @property
     def _parallel_execution_enabled(self):
-        return self._bootstrap.process_executor if self._bootstrap is not None else None
+        if self._bootstrap is None:
+            return False
+        return self._bootstrap.process_executor is not None
```
simonafk, line 73: What about `return self._bootstrap is not None and self._bootstrap.process_executor is not None` ?

bionic/core/flow_execution.py
```diff
@@ -67,11 +68,15 @@ def __init__(self, bootstrap, flow_instance_uuid, task_key_logger):
 
     @property
     def _parallel_execution_enabled(self):
```
namanjain, line 70: Thanks for changing this!

namanjain (approved):
LGTM

PR #275, opened 2020-10-10
Replace gsutil and sq-blocks with fsspec

namanjain:
This change replaces [gsutil](https://cloud.google.com/storage/docs/gsutil) and [sq-blocks](https://github.com/square/blocks) with [fsspec](https://filesystem-spec.readthedocs.io/en/latest/index.html). We had problems
with gsutil like breaking changes with Python 3.8 (GCP team asked us to
follow a [fix](https://github.com/GoogleCloudPlatform/gsutil/pull/1107) but now it is under question too), and we always wanted to
deprecate sq-blocks in favor of whatever rest of the Bionic code uses
to communicate with GCS.

fsspec has a really nice interface and has a lot of proven implementations
like [gcsfs](https://github.com/dask/gcsfs) and [s3fs](https://github.com/dask/s3fs). If or rather when we decide to support s3, it would be
easy to use fsspec at that point too.

### DIFF
--- a/bionic/aip/task.py
+++ b/bionic/aip/task.py
@@ -6,8 +6,9 @@ import attr
 from typing import Callable, Optional
 
 from bionic.aip.client import get_aip_client
-from bionic.deps.optdep import import_optional_dependency
 from bionic.aip.future import Future
+from bionic.deps.optdep import import_optional_dependency
+from bionic.gcs import get_gcs_fs_without_warnings
 
 
 @attr.s(auto_attribs=True, frozen=True)
@@ -100,13 +101,13 @@ class Task:
         return output
 
     def _stage(self):
-        # Scope the blocks import to this function to avoid raising for anyone not using it.
-        blocks = import_optional_dependency("blocks")
         cloudpickle = import_optional_dependency("cloudpickle")
 
         path = self.inputs_uri()
         logging.info(f"Staging task {self.name} at {path}")
-        with blocks.filesystem.GCSNativeFileSystem().open(path, "wb") as f:
+
+        gcs_fs = get_gcs_fs_without_warnings()
+        with gcs_fs.open(path, "wb") as f:
             cloudpickle.dump(self, f)
 
     def submit(self) -> Future:
--- a/bionic/deps/extras.py
+++ b/bionic/deps/extras.py
@@ -29,15 +29,17 @@ extras["standard"] = combine(extras["matplotlib"], extras["viz"])
 
 extras["dill"] = ["dill"]
 extras["dask"] = ["dask[dataframe]"]
-extras["gcp"] = ["google-cloud-storage", "fsspec", "gcsfs"]
+extras["gcp"] = ["fsspec", "gcsfs"]
 extras["parallel"] = ["cloudpickle", "loky"]
 extras["geopandas"] = ["geopandas"]
-extras["aip"] = [
-    "google-api-python-client",
-    "google-cloud-logging",
-    "sq-blocks",
-    "cloudpickle",
-]
+extras["aip"] = combine(
+    [
+        "google-api-python-client",
+        "google-cloud-logging",
+        "cloudpickle",
+    ],
+    extras["gcp"],
+)
 
 extras["examples"] = combine(extras["standard"], ["scikit-learn"])
 extras["full"] = combine(*extras.values())
--- a/bionic/deps/extras.py
+++ b/bionic/deps/extras.py
@@ -29,7 +29,7 @@ extras["standard"] = combine(extras["matplotlib"], extras["viz"])
 
 extras["dill"] = ["dill"]
 extras["dask"] = ["dask[dataframe]"]
-extras["gcp"] = ["google-cloud-storage"]
+extras["gcp"] = ["google-cloud-storage", "fsspec", "gcsfs"]
 extras["parallel"] = ["cloudpickle", "loky"]
 extras["geopandas"] = ["geopandas"]
 extras["aip"] = [
--- a/bionic/gcs.py
+++ b/bionic/gcs.py
@@ -14,7 +14,7 @@ _cached_gcs_fs = None
 
 
 def get_gcs_fs_without_warnings(cache_value=True):
-    # TODO It's not expensive to create gcs filesystem, but caching this enables
+    # TODO It's not expensive to create the gcs filesystem, but caching this enables
     # us to mock the cached gcs_fs with a mock implementation in tests. We should
     # change the tests to inject the filesystem in a different way and get rid of
     # this caching.
--- a/bionic/gcs.py
+++ b/bionic/gcs.py
@@ -42,44 +42,10 @@ def get_gcs_fs_without_warnings(cache_value=True):
         return fsspec.filesystem("gcs")
 
 
-_cached_gcs_client = None
-
-
-def get_gcs_client_without_warnings(cache_value=True):
-    # TODO This caching saves a lot of time, especially in tests.  But it would
-    # be better if Bionic were able to re-use its in-memory cache when creating
-    # new flows, instead of resetting the cache each time.
-    if cache_value:
-        global _cached_gcs_client
-        if _cached_gcs_client is None:
-            _cached_gcs_client = get_gcs_client_without_warnings(cache_value=False)
-        return _cached_gcs_client
-
-    gcs = import_optional_dependency("google.cloud.storage", purpose="caching to GCS")
-
-    with warnings.catch_warnings():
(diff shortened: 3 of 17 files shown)


bionic/persistence.py
```diff
@@ -872,38 +863,93 @@ class GcsFilesystem:
...
+        self._validate_object_name_from_url(url)
+        with self._fs.open(url, "wb") as f:
+            f.write(content_bytes)
 
     def read_bytes(self, url):
-        return self._tool.blob_from_url(url).download_as_string()
+        self._validate_object_name_from_url(url)
+        return self._fs.cat_file(url)
+
+    def get_dir(self, url, path):
```
jqmp, line 919: I'm a little confused about why we need both these new methods and the old methods.

tests/test_flow/fakes.py
```diff
@@ -67,132 +67,100 @@ def submit(self, task_config, fn, *args, **kwargs):
...
-        assert not prefix.startswith("gs://")
-        return [
-            FakeBlob(self.name, k.replace(f"gs://{self.name}/", "", 1), self.blobs)
-            for (k, v) in self.blobs.items()
-            if k.startswith(f"gs://{self.name}/{prefix}")
-        ]
-
-
-class FakeGCS:
+class FakeGcsFs:
```
jqmp, line 70: Is there anything about this fake filesystem that's specific to GCS? Could we just use another FSSPEC implementation, like a `MemoryFileSystem` (maybe modified to not be global)?

bionic/gcs.py
```diff
@@ -37,99 +35,33 @@ def get_gcs_client_without_warnings(cache_value=True):
...
-    def __init__(self, url):
+    _upload_to_gcs(path, url)
+
+
+def _upload_to_gcs(path, url):
+    fs = get_gcs_fs_without_warnings()
+    if path.is_dir():
+        fs.put(str(path), url, recursive=True)
+    else:
+        # If gcs url is a folder, we want to write the file in the folder.
```
jqmp, line 57: nit: "GCS URL" and "`put` API" should have the word "the" in front of them. In general, every noun needs an article unless it's a proper name, and these are not.

bionic/gcs.py
```diff
@@ -37,99 +35,33 @@ def get_gcs_client_without_warnings(cache_value=True):
...
-        "-m",  # Transfer files in parallel.
-        "cp",
-        "-r",  # Recursively sync sub-directories.
-        src_url,
-        dst_url,
-    ]
-    logger.debug("Running command: %s" % " ".join(args))
-    subprocess.check_call(args)
-    logger.debug("Finished running gsutil")
+            url = url + str(path).rsplit("/", 1)[1]
```
jqmp, line 66: As we discussed on Slack, `str(path).rsplit("/", 1)[1]` seems like it could just be `path.name`.

jqmp (commented):
FSSPEC looks great! Although I'm surprised by the `put` bug.

In a future PR, could we remove our `LocalFilesystem` and `GcsFilesystem` classes and just use the FSSPEC filesystem directly?

bionic/aip/task.py
```diff
@@ -100,13 +101,14 @@ def _ai_platform_job_spec(self):
         return output
 
     def _stage(self):
-        # Scope the blocks import to this function to avoid raising for anyone not using it.
-        blocks = import_optional_dependency("blocks")
+        # Scope the cloudpickle import to this function to avoid raising for anyone not using it.
```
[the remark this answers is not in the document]
namanjain: Probably not, we use this format at multiple places in Bionic.

bionic/gcs.py
```diff
@@ -2,30 +2,28 @@
...
-        return _cached_gcs_client
+        global _cached_gcs_fs
+        if _cached_gcs_fs is None:
+            _cached_gcs_fs = get_gcs_fs_without_warnings(cache_value=False)
+        return _cached_gcs_fs
 
-    gcs = import_optional_dependency("google.cloud.storage", purpose="caching to GCS")
+    fsspec = import_optional_dependency("fsspec", purpose="caching to GCS")
 
     with warnings.catch_warnings():
```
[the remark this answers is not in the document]
namanjain: Yup, it does

bionic/gcs.py
```diff
@@ -2,30 +2,28 @@
...
 
+_cached_gcs_fs = None
 
-def get_gcs_client_without_warnings(cache_value=True):
+
+def get_gcs_fs_without_warnings(cache_value=True):
     # TODO This caching saves a lot of time, especially in tests.  But it would
     # be better if Bionic were able to re-use its in-memory cache when creating
     # new flows, instead of resetting the cache each time.
     if cache_value:
```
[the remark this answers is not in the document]
namanjain: oh I should have checked this before, but no, it doesn't seem expensive. I can remove the cached client, but it would break unit tests. I can take a look at changing unit tests, but do you mind if I leave a TODO and do it in a separate PR? This PR is big as it is.

tests/test_flow/fakes.py
namanjain: No, you're right. I should have thought of using `MemoryFileSystem` implementation! I think I implemented part of it in my PR 🤦. But what do you mean by not be global?

bionic/persistence.py
namanjain: What do you mean by new and old methods here?

tests/test_flow/fakes.py
jqmp: From the `MemoryFileSystem` docs:

    This is a global filesystem so instances of this class all point to the same
    in memory filesystem.

We probably don't want this to be global if we're using it for testing. I suspect a subclass like this would work, but I could be wrong:

    class LocalMemoryFileSystem(MemoryFileSystem):
        def __init__(self):
            self.store = {}
            self.pseudo_dirs = []

bionic/persistence.py
jqmp: It looks like you've added wrappers for several FSSPEC methods, but you still have the original methods from our own filesystem interface. Is this class supposed to implement both interfaces? If so, why is that necessary?

namanjain: In the long run, it would be nice to make this a simple wrapper class over fsspec Filesystem class that simply does the validation of directory scope. But for this PR, I tried to make the changes in the least invasive way as possible. That means changing only the `GcsFilesystem` and and `GcsTool` classes.

tests/test_flow/fakes.py
namanjain: Actually, using `MemoryFileSystem` as a fake is getting annoying because things break with the gcs protocol `gs://` at the beginning of the url. I'll have to override all methods to remove / add protocol as required to make it work (I mean all and not only the ones that Bionic uses, since  `MemoryFileSystem` might be using some of those APIs underneath). There are some other small nuanced differences like for directories, `GcsFileSystem` return `True` whereas `MemoryFileSystem` returns false.

I think the current custom implementation is a better workaround than that.

namanjain: Due to nuanced differences, I don't think it would be easy to use fsspec like we initially thought either.

bionic/deps/extras.py
```diff
@@ -29,13 +29,12 @@ def combine(*dep_lists):
 
 extras["dill"] = ["dill"]
 extras["dask"] = ["dask[dataframe]"]
-extras["gcp"] = ["google-cloud-storage"]
+extras["gcp"] = ["fsspec", "gcsfs"]
 extras["parallel"] = ["cloudpickle", "loky"]
 extras["geopandas"] = ["geopandas"]
 extras["aip"] = [
```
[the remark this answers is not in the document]
namanjain: We discussed this on slack, but we don't have a check to assert that GCS caching is enabled for AIP. I'll add that in a separate PR.

tests/test_flow/fakes.py
```diff
@@ -67,132 +67,111 @@ def submit(self, task_config, fn, *args, **kwargs):
...
-            if k.startswith(f"gs://{self.name}/{prefix}")
-        ]
-
-
-class FakeGCS:
+# TODO It would be nice to use a different fsspec implementation here instead
+# of writing a custom one. We tried fsspec.implementations.memory.MemoryFileSystem
+# implementation, but that has a few problems.
+# 1) The glob method returns the protocol gs:// whereas gcsfs implementation doesn't
+# 2) There is a bug in the `rm` method and the tests fails to wipe the gcs cache.
```
namanjain, line 74: I opened an issue in fsspec for this bug as well https://github.com/intake/filesystem_spec/issues/449

jqmp (approved):

PR #278, opened 2020-10-13
fix aip computation logging

simonafk:
Temporary fix for AIP logging. Entities that use AIP will show `computed ...` instead of `Downloading ... from GCS`.

Note that this does not fix all problems. For example, if an AIP entity depends on another previously computed entity X, then it may not show the logs saying that it has downloaded X from GCS. This fix also does not show print all the logs from AIP.

One possible long-term fix is to pass these logging information through stackdriver.

### DIFF
--- a/bionic/aip/future.py
+++ b/bionic/aip/future.py
@@ -1,7 +1,7 @@
 import logging
 import pickle
 import time
-from concurrent.futures import Future as _Future, TimeoutError, CancelledError
+from concurrent.futures import Future as ConcurrentFuture, TimeoutError, CancelledError
 from enum import Enum, auto
 
 from bionic.aip.client import get_aip_client
@@ -42,7 +42,7 @@ class State(Enum):
         }
 
 
-class Future(_Future):
+class Future(ConcurrentFuture):
     """This future represents a job running on AI platform
 
     The result of the running job will be pickled, and this future can load that pickle.
@@ -56,6 +56,12 @@ class Future(_Future):
         self.job_id = job_id
         self.output = output
         self.aip = get_aip_client()
+        self.done_callbacks = []
+
+        # These are used for caching of the results when the remote job is
+        # finished.
+        self.state = None
+        self.error = None
 
     @property
     def name(self):
@@ -67,9 +73,22 @@ class Future(_Future):
         return True
 
     def _get_state_and_error(self):
+        if self.state is not None:
+            assert self.state.is_finished()
+            return self.state, self.error
+
         request = self.aip.projects().jobs().get(name=self.name)
         response = request.execute()
-        return State[response["state"]], response.get("errorMessage", "")
+        state, error = State[response["state"]], response.get("errorMessage", "")
+
+        if state.is_finished():
+            self.state = state
+            self.error = error
+            for fn in self.done_callbacks:
+                fn(self)
+            self.done_callbacks.clear()
+
+        return state, error
 
     def cancelled(self):
         # We may want to distinguish between canceling and cancelled on GCP.
@@ -118,3 +137,9 @@ class Future(_Future):
             raise CancelledError(f"{self.job_id}: " + error)
         if state is State.FAILED:
             return AipError(f"{self.job_id}: " + error)
+
+    def add_done_callback(self, fn):
+        if self.done():
+            fn(self)
+        else:
+            self.done_callbacks.append(fn)
--- a/tests/test_flow/fakes.py
+++ b/tests/test_flow/fakes.py
@@ -1,7 +1,7 @@
-from contextlib import contextmanager
-from functools import partial
 import io
 import logging
+from contextlib import contextmanager
+from functools import partial
 from pathlib import Path
 from unittest.mock import Mock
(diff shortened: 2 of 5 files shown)


tests/test_flow/test_logging.py
```diff
@@ -112,9 +119,48 @@ def log_unpickleable_value():
...
+    @builder
+    @bn.aip_task_config("n1-standard-8")
+    def x_plus_two(x_plus_one):
+        return x_plus_one + 1
+
+    flow = builder.build()
+
+    assert flow.get("x_plus_one") == 2
+
+    log_checker.contains_substrings(
```
jqmp, line 151: I assume we use substrings because we can't predict the full AIP url? Could we use regexes here to make it clear what part we can't predict?

tests/test_flow/test_logging.py
```diff
@@ -2,17 +2,24 @@
...
     def __init__(self, caplog):
         self._caplog = caplog
 
-    def expect(self, *expected_messages):
+    def expect_all(self, *expected_messages):
         actual_messages = [record.getMessage() for record in self._caplog.records]
         assert set(actual_messages) == set(expected_messages)
         self._caplog.clear()
 
+    def contains_substrings(self, *substrings):
```
jqmp, line 17: nit: can we name this in a way that's congruent with the other method? Such as `expect_superset`?

Also, do we expect the number of substrings to match the number of logged messages? Or could there be extra messages that don't match anything?

tests/test_flow/fakes.py
```diff
@@ -194,7 +194,9 @@ def run_in_fake_gcp(fake_gcs: FakeGCS):
     mock_gcs_client.get_bucket = FakeBucket
 
     def create_aip_job(body, parent):
-        run_aip(body["trainingInput"]["args"][3])
+        # AIP logs are not printed locally.
+        with caplog.at_level(logging.CRITICAL):
```
jqmp, line 198: Why do we need to change the log level?

tests/test_flow/test_logging.py
```diff
@@ -112,9 +119,48 @@ def log_unpickleable_value():
...
+
+    flow = builder.build()
+
+    assert flow.get("x_plus_one") == 2
+
+    log_checker.contains_substrings(
+        "Accessed   x(x=1) from definition",
+        "Uploading x(x=1) to GCS ...",
+        "Started task on AI Platform: https://console.cloud.google.com/ai-platform/jobs/",
+        "Computed   x_plus_one(x=1)",
```
jqmp, line 155: Maybe this should say `Computed    x_plus_one(x=1) on AIP` or something to make it clear that it happened remotely?

simonafk:
> Temporary fix for AIP logging. Entities that use AIP will show `computed ...` instead of `Downloading ... from GCS`.
> 
> Note that this does not fix all problems. For example, if an AIP entity depends on another previously computed entity X, then it may not show the logs saying that it has downloaded X from GCS. This fix also does not show print all the logs from AIP.
> 
> One possible long-term fix is to pass these logging information through stackdriver.

Mistyped this. Logging message that we want to remove is `Loaded     ... from disk cache`. The logging message `Downloading ... from GCS` is correct.

tests/test_flow/fakes.py
simonafk: Logs from running AIP do not show up in local instance, and we want to emulate it here. Added some comments.

jqmp: Ah, that makes sense -- thanks!

tests/test_flow/test_logging.py
simonafk: I am somewhat neutral to this. It seems redundant; we already have a bunch of logs for AIP when starting a job.

jqmp: We do, but the `Computed` message won't necessarily appear right after the `Started task` message; there could be several other messages in between, so it seems like it won't necessarily be obvious where the computation happened.

tests/test_flow/test_logging.py
simonafk: Renamed to `expect_regex`

> Also, do we expect the number of substrings to match the number of logged messages? Or could there be extra messages that don't match anything?

No, this is due to differences between running AIP in fake and real gcp. The message `Downloading ... from GCS` is logged in real GCP but not in fake. (Tried temporarily cleaning out `tmp_path` and doing a deep copy of the AIP job but it did not work, fixing this in fake GCP may be complicated.)

bionic/aip/future.py
```diff
@@ -55,6 +55,11 @@ def __init__(self, project_name: str, job_id: str, output: str):
         self.job_id = job_id
         self.output = output
         self.aip = get_aip_client()
+        self.done_callback = []
+
+        # Only set if the state is terminal
```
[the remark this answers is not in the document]
simonafk: Updated comments

tests/test_flow/test_logging.py
simonafk: Updated to show `Computed ... using AIP`

PR #279, opened 2020-10-13
Pull saving/loading code out of persistent cache

jqmp:
Now all serialization and deserialization happens outside the
PersistentCache class, which no longer deals with Results or Protocols,
just Artifacts and Provenances. This will make it possible to have
separate tasks for loading and saving data.

Unfortunately there are still a couple of awkward cases where we need to
call into the PersistentCache to handle some details. We could refactor
this further, but I think the path forward will become clearer as we
pull more stuff out of the cache. (For example, maybe we'll move to
separate local and cloud caches.)

### DIFF
--- a/bionic/core/task_execution.py
+++ b/bionic/core/task_execution.py
@@ -24,10 +24,16 @@ import logging
 
 from enum import auto, Enum, IntEnum
 
-from ..datatypes import Result
-from ..exception import CodeVersioningError
+from ..datatypes import Artifact, Result
+from ..exception import (
+    CodeVersioningError,
+    EntitySerializationError,
+    UnsupportedSerializedValueError,
+)
 from ..persistence import Provenance, ProvenanceDigest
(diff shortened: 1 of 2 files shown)


namanjain (commented):
The change itself looks good. I agree that we can refactor persistence classes further and that can be done incrementally. But how do you feel about moving the refactored logic into a separate class instead of moving it to `TaskState`?

jqmp:
> The change itself looks good. I agree that we can refactor persistence classes further and that can be done incrementally. But how do you feel about moving the refactored logic into a separate class instead of moving it to `TaskState`?

In the abstract, I think that would make sense. However, in this case I think the next step is to move `_local_artifact_from_value` and `_value_from_local_artifact` into two new Providers, which convert from `<X>` to `X/artifact` and from `X/artifact` to `X`, respectively. I'm working on that PR now, so I don't think the proposed separate class would be around very long.

namanjain (approved):

PR #281, opened 2020-10-21
fix aip double gcs download

simonafk:


bionic/core/task_execution.py
```diff
@@ -580,9 +581,9 @@ def sync_after_remote_computation(self):
         # do anything if it fails now.)
         self._cache_accessor.flush_stored_entries()
 
-        # Then, populate the value hashes.
-        if self._result_value_hash is None:
-            self._load_value_hash()
+        artifact = self._get_artifact()
+        self._cache_accessor.save_artifact(artifact)
```
namanjain, line 585: Nice catch! 

@jqmp Although the fix works, I think it can easily happen again in the future. What do you both think about changing the persistence `load_artifact` API to save the artifact locally when downloading from a remote source? I mean we are already downloading the artifact from remote, might as well save it.

tests/test_flow/test_persistence_aip.py
```diff
@@ -29,3 +29,33 @@ def c(a, b):
         return a - b  # 6 (2 * 3) - 5 (2 + 3)
 
     assert builder.build().get("c") == 1
+
+
+@pytest.mark.needs_aip
+def test_aip_not_persisted_dependency(aip_builder):
```
namanjain, line 35: I'm a little worried about having too many AIP tests. Each test takes 4-5 minutes when you run with AIP. If it's just this, can we modify the original test to contain a non persisted dep?

tests/test_flow/fakes.py
```diff
@@ -177,7 +212,13 @@ def create_aip_job(body, parent):
...
-        with caplog.at_level(logging.CRITICAL):
+        #
+        # Additionally, AIP executions run in an isolated environment and do not
+        # have access to disk cache from local bionic instance. It must retrieve
+        # the data from GCS. Emulate this behavior by temporarily removing all
+        # files in the tmp_path directory. The files are restored after the
+        # AIP computation is completed.
+        with caplog.at_level(logging.CRITICAL), temporary_clean_dir(tmp_path):
             run_aip(body["trainingInput"]["args"][3])
         return Mock()
```
namanjain, line 223: Does the AIP mocks here use the `FakeAipExecutor` (and other related fakes)? If it does not, should it?

tests/test_flow/fakes.py
```diff
@@ -177,7 +212,13 @@ def create_aip_job(body, parent):
...
         # only test against logs which are printed locally.
-        with caplog.at_level(logging.CRITICAL):
+        #
+        # Additionally, AIP executions run in an isolated environment and do not
+        # have access to disk cache from local bionic instance. It must retrieve
+        # the data from GCS. Emulate this behavior by temporarily removing all
+        # files in the tmp_path directory. The files are restored after the
+        # AIP computation is completed.
+        with caplog.at_level(logging.CRITICAL), temporary_clean_dir(tmp_path):
             run_aip(body["trainingInput"]["args"][3])
```
namanjain, line 222: Do you need to call this function here? Would it be easier to run `body["trainingInput"]["args"]` as a shell command instead? That way, you won't need to use this private method here.

tests/test_flow/fakes.py
```diff
@@ -166,8 +169,40 @@ def rm(self, url, recursive=False):
             del self._files_by_url[url]
 
 
+def move_files_in_dir(src_dir, dst_dir):
```
namanjain, line 172: I think it is doing the same thing as `bionic.utils.files.recursively_copy_path`. Can you use the util method if it works in place of this one?

tests/test_flow/fakes.py
```diff
@@ -166,8 +169,40 @@ def rm(self, url, recursive=False):
...
+                    shutil.rmtree(path)
+                else:
+                    os.remove(path)
+
+            move_files_in_dir(tmp_backup_dir, target_dir)
+
+
 @contextmanager
-def run_in_fake_gcp(fake_gcs_fs: FakeGcsFs, caplog):
+def run_in_fake_gcp(fake_gcs_fs: FakeGcsFs, caplog, tmp_path):
```
namanjain, line 205: Do we need to remove all files from the `tmp_path` or only from the cache dir? If it's the later, then wdyt about passing the cache dir here instead?

namanjain (commented):
Logically, this change looks great and the added tests gives me more confidence about the fix. But I'm not sure if the fix should be in execution layer or persistence layer. LGTM once this question is answered and rest of the comments are addressed.

namanjain:
Can you also add description to the PR and the commit message describing the change as well? I found [this post](https://chris.beams.io/posts/git-commit/) on writing commit messages super helpful.

tests/test_flow/fakes.py
```diff
@@ -166,8 +169,40 @@ def rm(self, url, recursive=False):
...
+
+
+@contextmanager
+def temporary_clean_dir(target_dir):
+    """
+    Temporarily remove all files in a directory for the duration of a context.
+    Once the context ends, the directory is restored to its original state. That
+    means any new file created during the duration of the context are deleted.
+    """
+    with TemporaryDirectory() as tmp_backup_dir:
```
namanjain, line 188: Oh, and if we only clean cache dir, then we can also use `tmp_path` here as well instead of using `TemporaryDirectory`.

bionic/core/task_execution.py
jqmp: I'm not sure I totally understand this bug -- as you mentioned elsewhere, it would be great for this PR to include a description of what's going wrong. But in theory the cache accessor should be storing a copy of any artifact object it creates. Is the problem that `accessor.load_artifact` is creating a local copy but not saving it? If so, I agree that we should probably ensure that calling `load_artifact` twice doesn't result in two downloads. Off the top of my head, I'm not sure if it makes more sense to actually save the local artifact, or just to maintain an in-memory copy of the local artifact object. (The former seems more robust, but also means that `load_artifact` can now mutate the cache state. Maybe that's not a big deal....)

As far as this happening in the future: I agree! Can we add a test that checks for multiple downloads? I see this PR updates the logging tests, but those don't seem like they'd catch duplicate messages.

simonafk: The logging tests were updated to check that the total count of the messages is the same as the expected. `assert len(actual_messages) == len(expected_patterns)`. But yes, I think a more robust double download check would be nicer.

tests/test_flow/fakes.py
simonafk: There is a difference. Firstly it is a move operation. Secondly, we want to move the contents of the source directory instead of moving the directory itself. For example if we have a file `src/a/b`, we want the resulting path to be `dest/a/b` instead of `dest/src/a/b`.

tests/test_flow/fakes.py
simonafk: If we run this in a separate process, we need to ensure that any writes to the GCS filesystem in the separate process is reflected here in the original process. If we have a fake GCS, then we need to pass a proxy of the fake GCS to the separate process. This is more involved.

bionic/core/task_execution.py
jqmp: Ah, sorry, I missed that change! That makes sense.

As you said, I do think a more explicit check would be good, but I guess you'll be the best judge of how practical it is.

tests/test_flow/fakes.py
namanjain: Ah, good point. I still think it's worthwhile and we should investigate if we can use our `PytestManager` to instantiate the fake gcs, but we don't have to do it in this PR.

tests/test_flow/test_persistence_aip.py
simonafk: Good point. I have combined the AIP tests into one.

simonafk:
> Logically, this change looks great and the added tests gives me more confidence about the fix. But I'm not sure if the fix should be in execution layer or persistence layer. LGTM once this question is answered and rest of the comments are addressed.

Hmm good question. The persistence layer understands how to load and save data, but it does not know whether the data is fresh or stale. I think there is logic that initiates flushing of in-memory data and copying of data, and I think the logic should be in the execution layer.

tests/test_flow/fakes.py
simonafk: Updated to use cache dir.

tests/test_flow/fakes.py
simonafk: I am using `TemporaryDirectory` here because it is much simpler. What do you think?

tests/test_flow/fakes.py
simonafk: No, it is not using the fake executor. I think faking / mocking AIP API would allow us to give the most test coverage to our code. Right now the fake AIP works by replacing a global variable; I think using injection would make things more robust. Can look into this in a future PR if needed.

jqmp:
> > Logically, this change looks great and the added tests gives me more confidence about the fix. But I'm not sure if the fix should be in execution layer or persistence layer. LGTM once this question is answered and rest of the comments are addressed.
> 
> Hmm good question. The persistence layer understands how to load and save data, but it does not know whether the data is fresh or stale. I think there is logic that initiates flushing of in-memory data and copying of data, and I think the logic should be in the execution layer.

@simonafk I have a comment replying to Naman's that addresses this (somewhat). Could you provide some more explanation of what is going wrong and why the fix should be in the execution layer? In particular, I'm not sure what you mean by "fresh or stale" here; what exactly is stale? We are already calling `_cache_accessor .flush_stored_entries()`, so if anything is still stale after that, that seems like it would be an error in the persistence layer.

Another question: does this bug happen when loading cached data from GCS? For example, if you run a Bionic flow that caches to GCS, then remove the local disk cache, and then and re-run it, do we seem the same double download? (I tried to test this myself but I ran into a problem, as noted on Slack.)

jqmp:
> Another question: does this bug happen when loading cached data from GCS? For example, if you run a Bionic flow that caches to GCS, then remove the local disk cache, and then and re-run it, do we seem the same double download? (I tried to test this myself but I ran into a problem, as noted on Slack.)

Okay, I was able to test this and it looks like we do have too many downloads from GCS, even when not using AIP. We can reproduce it like this:

    python -m example.ml_workflow_cli --bucket some-bucket
    rm -rf bndata
    python -m example.ml_workflow_cli --bucket some-bucket

Since the problem is not specific to AIP, I don't think it makes sense to fix it in `sync_after_remote_computation`. I think there's a more fundamental problem, probably introduced by my cache refactoring. I'll look into this some more.

BTW, thanks for finding this bug!

simonafk:
Closed this PR since the main problem, double gcs download, has been fixed in a different PR and other changes, such as checking the logs after running AIP, has been imported into other PRs.

PR #282, opened 2020-10-21
Run GCS tests using real GCS in CI

namanjain:
This change uses `setup-gcloud` GitHub Action to set up gcloud on CI
and runs the GCS tests using both fake and actual GCS.

This was a lot easier to use than I expected. You just need to add the
service account key to secrets and use it to authenticate. For more
information on how to use it, please check the
[readme](https://github.com/GoogleCloudPlatform/github-actions/blob/master/setup-gcloud/README.md).

I also verified that this works by checking that the CI build doesn't
skip GCS tests. I can also see the cached files inside the new bionic
bucket in our team's project when the CI is running.

Screenshot of docs change:
Screen Shot 2020-10-21 at 5 47 56 PM

### DIFF
--- a/docs/release-notes.rst
+++ b/docs/release-notes.rst
@@ -65,14 +65,19 @@ For each release, we list the following types of change (in this order):
 - **Development Changes**: Significant changes to Bionic's development process, such
   as changes to our Pytest configuration or our Continuous Integration ("CI").
 
-.. Upcoming Version (Not Yet Released)
-.. -----------------------------------
+Upcoming Version (Not Yet Released)
+-----------------------------------
 
 .. Record any notable changes in this section. When we update the current version,
    add a new version heading below, and then comment out the heading above until more
    changes are added. This way, the "Upcoming Version" section will be never be visible
    in the "stable" docs (corresponding to the last release) but will be visible in the
    "latest" docs (corresponding to the master branch).
+
+Development Changes
+...................
+
+- Our CI now also runs all GCS tests using the real GCS.
 
 0.9.1 (Oct 15, 2020)
 --------------------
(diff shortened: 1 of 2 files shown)


docs/release-notes.rst
```diff
@@ -65,15 +65,20 @@ For each release, we list the following types of change (in this order):
...
 .. Record any notable changes in this section. When we update the current version,
    add a new version heading below, and then comment out the heading above until more
    changes are added. This way, the "Upcoming Version" section will be never be visible
    in the "stable" docs (corresponding to the last release) but will be visible in the
    "latest" docs (corresponding to the master branch).
 
+Development Changes
+...................
+
+- Our CI now also runs all GCS tests using real GCS.
```
jqmp, line 80: nit: "*the* real GCS".

jqmp (approved):
Wow! I had no idea this was possible. Great find.

Any idea what the monetary cost would be? I assume it's pretty small, but it might be good to add a comment with a rough estimate.

namanjain:
> Wow! I had no idea this was possible. Great find.
> 
> Any idea what the monetary cost would be? I assume it's pretty small, but it might be good to add a comment with a rough estimate.

I'm not sure how to get the monetary cost for a single CI run, but the projected cost for all buckets (including our individual ones) in our project was $1.23 yesterday. This includes running the CI at least 5-6 times and any GCS testing done on any of our local machine. I think $1.23 is pretty small but since it's projected cost, I don't know how accurate that is. we might have to wait a while to get the actual cost.

jqmp:
> I'm not sure how to get the monetary cost for a single CI run, but the projected cost for all buckets (including our individual ones) in our project was $1.23 yesterday. This includes running the CI at least 5-6 times and any GCS testing done on any of our local machine. I think $1.23 is pretty small but since it's projected cost, I don't know how accurate that is. we might have to wait a while to get the actual cost.

I agree, a dollar or so per day seems pretty affordable for a multi-engineer project.

simonafk (approved):

PR #283, opened 2020-10-23
Fix redundant gcs downloads

jqmp:
This fixes the bug that @simonafk found where we download too many times from GCS, and also updates the tests to catch this type of bug in the future. There are more details in the individual commit messages.

I think this obviates the main fix in https://github.com/square/bionic/pull/281/files, but that PR still has some useful logging test changes that are probably worth keeping.

### DIFF
--- a/bionic/persistence.py
+++ b/bionic/persistence.py
@@ -155,21 +155,42 @@ class CacheAccessor:
             if entry is None:
                 return None
 
-            if entry.tier == "local":
-                return entry.artifact
-
-            elif entry.tier == "cloud":
-                return self._local_artifact_from_cloud(entry.artifact)
-
-            else:
-                raise AssertionError("Unrecognized tier: " + entry.tier)
+            return entry.artifact
 
         except InternalCacheStateError as e:
             self.raise_state_error_with_explanation(e)
 
-    def save_artifact(self, artifact):
+    def replicate_and_load_local_artifact(self):
         """
-        Saves an artifact in each cache layer that doesn't already have an exact match.
+        If any cached artifact exists, replicates it to each cache layer and returns
+        the local copy; otherwise returns None.
+        """
+
+        try:
+            entry = self._get_nearest_entry_with_artifact()
+
+            if entry is None:
+                return None
+
+            if entry.tier == "local":
+                local_artifact = entry.artifact
+
+            elif entry.tier == "cloud":
+                local_artifact = self._local_artifact_from_cloud(entry.artifact)
+
+            else:
+                raise AssertionError("Unrecognized tier: " + entry.tier)
+
+            self._save_or_reregister_artifact(local_artifact)
+            return local_artifact
+
+        except InternalCacheStateError as e:
+            self.raise_state_error_with_explanation(e)
+
+    def save_local_artifact(self, artifact):
+        """
+        Saves a local artifact in each cache layer that doesn't already have an exact
+        match.
         """
 
         try:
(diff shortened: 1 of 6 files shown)


bionic/persistence.py
```diff
@@ -149,27 +149,50 @@ def load_artifact(self):
...
+                local_artifact = entry.artifact
 
             elif entry.tier == "cloud":
-                return self._local_artifact_from_cloud(entry.artifact)
+                local_artifact = self._local_artifact_from_cloud(entry.artifact)
 
             else:
                 raise AssertionError("Unrecognized tier: " + entry.tier)
 
+            self._save_or_reregister_artifact(local_artifact)
```
simonafk, line 186: I was wondering about doing this also. I was thinking if this can cause unnecessary uploads to GCS. For example, we try to load an artifact from a previous computation; the new code would cause the disk copy to be uploaded to GCS if it is not already there; and it later turns out that we need an exact match while the artifact only has a functional match. Then, we will need to re-compute the entity anyways.

bionic/persistence.py
```diff
@@ -149,27 +149,50 @@ def load_artifact(self):
...
+
+        except InternalCacheStateError as e:
+            self.raise_state_error_with_explanation(e)
+
+    def load_local_artifact(self):
+        """
+        Returns a local artifact if any cached artifact exists.
+
+        If the nearest artifact is not local, it will be copied to a local instance and
+        registered in the cache.
```
simonafk, line 168: I think this will upload the artifact to the cloud as well? (if so, mention it in the comment)

bionic/persistence.py
```diff
@@ -149,27 +149,50 @@ def load_artifact(self):
...
+
+            if entry is None:
+                return None
+
+            return entry.artifact
+
+        except InternalCacheStateError as e:
+            self.raise_state_error_with_explanation(e)
+
+    def load_local_artifact(self):
```
simonafk, line 163: I think this method is doing more than just `load`. Naming is hard. Wondering if having something like `sync` in the name would make things clearer?

bionic/persistence.py
jqmp: Yes, good point! I wrote the comment before adding the copy-to-cloud functionality. I'll fix it.

bionic/persistence.py
jqmp: Yes, I think that's a good idea.

bionic/persistence.py
jqmp: Good question. Can this happen? I haven't thought this all the way through, but we only call `load_local_artifact` when we're ready to load the data into memory, so I think the versioning check for this task should have already happened.

Also, two other things have to be true for this to happen:

1. The artifact wasn't already in GCS, which means we computed it with GCS caching off and then ran the flow again with GCS caching on. That can definitely happen, but probably not very often.
2. The user made a versioning mistake.

Given all these qualifications, this change seems pretty safe to me.

simonafk (approved):

jqmp:
Note to self: this should have release notes too!

jqmp:
@namanjain's recent change to CI revealed that these new tests don't work right when parallel execution is enabled. I think the rest of the changes are still good and worth merging, so I've added a hack to disable the new asserts in parallel mode and a TODO to come back and fix them, which I plan to do next.

namanjain (approved):
The main code looks good, but the patching code looks messy. But as per the Slack discussion, if the global patching code is going away soon, I think it's okay.

PR #284, opened 2020-10-23
Remove the FakeCloudStore class

jqmp:
Now that we have FakeGcsFs, we might as well use that instead of
maintaining two different fake GCS abstractions.

### DIFF
--- a/tests/test_flow/test_persistence_fuzz.py
+++ b/tests/test_flow/test_persistence_fuzz.py
@@ -5,7 +5,6 @@ from textwrap import dedent
 from random import Random
 
 from bionic.exception import CodeVersioningError
-from bionic.persistence import FakeCloudStore
 from bionic import interpret
 import bionic as bn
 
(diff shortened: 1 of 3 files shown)


tests/test_flow/test_persistence_fuzz.py
```diff
@@ -386,10 +385,8 @@ def _expected_called_names(
 
 
 @pytest.fixture(scope="function")
-def fuzzer(builder, make_list, tmp_path):
-    fake_cloud_store = FakeCloudStore(str(tmp_path / "BNTESTDATA_FAKE_CLOUD"))
-    builder.set("core__persistent_cache__cloud_store", fake_cloud_store)
-    return Fuzzer(builder, make_list)
+def fuzzer(fake_gcs_builder, make_list, tmp_path):
```
namanjain, line 388: Instead of doing this, what do you think about adding a marker that basically says don't run with real gcs? If that marker is present for a test, then we skip if the test also has `real-gcp` marker.

jqmp: I thought about this but ended up leaning against it. It doesn't seem to be possible to directly attach a mark to a fixture, so we'd have to do it by parametrizing the fixture (with a single parameter value). I don't love the proliferation of extra parameters, and of extra test instances that get created and then filtered out. I also didn't love the idea of adding more cases to our top-level filtering to solve this problem. That said, I'm not crazy about this solution either and I could probably be persuaded either way.

BTW, it's interesting to note that some of the limitations in pytest's parametrization are similar to those of Bionic's multiplicity. For example, in both frameworks it's frustrating that you can't dynamically set the number of parameters/cases based on the value of another fixture/entity. I'm really looking forward to solving this in Bionic (with descriptors).

jqmp:
I think the tests are failing because our fake GCS stuff doesn't work in parallel mode. I know @simonafk is working on this, so I'm going to leave this PR for now and come back when he's done.

jqmp:
Thanks to @simonafk's work, the tests are passing now!

namanjain (approved):

PR #285, opened 2020-10-26
Fix sphinx-autobuild

jqmp:
This handles some non-backwards-compatible changes to `sphinx-autobuild`. More details in commit messages.


jqmp:
BTW, I just tested this configuration on the previous version of `sphinx-autobuild` (0.7.1, from 2017) and it seems to work.

namanjain (approved):

PR #287, opened 2020-10-29
injectable gcs fs, allow fake gcs with parallel

simonafk:
Allow GCS filesystem instances to be injectable and stored in the flow.
For testing, injecting instead of global patching allows fake GCS to be
shared for multiprocessing.

### DIFF
--- a/bionic/aip/task.py
+++ b/bionic/aip/task.py
@@ -2,13 +2,13 @@
 Data model for task running on AI platform
 """
 import logging
+from typing import Callable, Optional, Any
+
 import attr
-from typing import Callable, Optional
 
 from bionic.aip.client import get_aip_client
 from bionic.aip.future import Future
 from bionic.deps.optdep import import_optional_dependency
-from bionic.gcs import get_gcs_fs_without_warnings
 
 
 @attr.s(auto_attribs=True, frozen=True)
@@ -53,6 +53,7 @@ class Task:
     function: Callable
     config: Config
     task_config: TaskConfig
+    gcs_fs: Any  # AbstractFileSystem-like object, not necessarily a subclass.
 
     def job_id(self):
         return f"{self.config.uuid}_{self.name}"
@@ -106,8 +107,7 @@ class Task:
         path = self.inputs_uri()
         logging.info(f"Staging task {self.name} at {path}")
 
-        gcs_fs = get_gcs_fs_without_warnings()
-        with gcs_fs.open(path, "wb") as f:
+        with self.gcs_fs.open(path, "wb") as f:
             cloudpickle.dump(self, f)
 
     def submit(self) -> Future:
@@ -126,4 +126,6 @@ class Task:
         request.execute()
         url = f'https://console.cloud.google.com/ai-platform/jobs/{spec["jobId"]}'
         logging.info(f"Started task on AI Platform: {url}")
-        return Future(self.config.project_name, self.job_id(), self.output_uri())
+        return Future(
+            self.gcs_fs, self.config.project_name, self.job_id(), self.output_uri()
+        )
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -8,7 +8,6 @@ import pandas as pd
 from pandas import testing as pdt
 
 import bionic as bn
-from bionic.gcs import get_gcs_fs_without_warnings
 
 
 # TODO This name is cumbersome; maybe one of these shorter names?
@@ -230,25 +229,6 @@ def longest_regex_prefix_match(regex, string, flags=0):
     match = re.search(regex[:max_succ_ix], string, flags=flags)
     assert match is not None
     return match
-
-
-def gcs_fs_wipe_path(url):
-    assert "BNTESTDATA" in url
-    fs = get_gcs_fs_without_warnings()
-    fs.rm(url, recursive=True)
-
-
-def gcs_fs_path_exists(url):
-    fs = get_gcs_fs_without_warnings()
-    return fs.exists(url)
-
-
-def gcs_fs_download(url, path):
-    fs = get_gcs_fs_without_warnings()
-    if fs.isdir(url):
-        fs.get(url, str(path), recursive=True)
-    else:
-        fs.get_file(url, str(path))
 
 
 def local_wipe_path(path_str):
--- a/tests/test_flow/conftest.py
+++ b/tests/test_flow/conftest.py
@@ -1,22 +1,17 @@
 import getpass
 import random
 from multiprocessing.managers import SyncManager
+from typing import Optional
 
 import pytest
 
 import bionic as bn
-from .fakes import FakeGcsFs, instrument_gcs_fs, run_in_fake_gcp
+from bionic.gcs import get_gcs_fs_without_warnings
+from .fakes import FakeGcsFs, fake_aip_client
 from ..helpers import (
     SimpleCounter,
     ResettingCallCounter,
-    gcs_fs_wipe_path,
-    gcs_fs_path_exists,
 )
-
-
-@pytest.fixture
-def fake_gcs_fs():
-    return FakeGcsFs()
 
 
 # Parameterizing a fixture adds the parameter in the test name at the end,
@@ -29,7 +24,7 @@ def fake_gcs_fs():
         pytest.param("parallel", marks=pytest.mark.parallel),
     ],
 )
-def parallel_execution_enabled(request):
+def parallel_execution_enabled(request) -> bool:
     return request.param == "parallel"
 
 
@@ -42,21 +37,8 @@ def parallel_execution_enabled(request):
         pytest.param("real-gcp", marks=pytest.mark.real_gcp),
     ],
 )
-def use_fake_gcp(request, fake_gcs_fs, caplog):
-    if request.param == "fake-gcp":
-        with run_in_fake_gcp(fake_gcs_fs, caplog):
-            yield True
-    else:
-        yield False
-
-
-# This replaces the global GCS filesystem with an instrumented version. Note that we
-# depend on `use_fake_gcp`, so we can end up wrapping either the real or the fake
-# filesystem.
-@pytest.fixture
-def instrumented_gcs_fs(use_fake_gcp, make_list):
-    with instrument_gcs_fs(make_list) as inst_gcs_fs:
-        yield inst_gcs_fs
+def use_fake_gcp(request) -> bool:
+    return request.param == "fake-gcp"
 
 
 # We provide this at the top level because we want everyone using FlowBuilder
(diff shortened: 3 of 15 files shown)


tests/test_flow/test_cache_api.py
```diff
@@ -77,24 +77,25 @@ def _validate_entry(self, entry):
...
         # are derelativized right away when we load the metadata YAML.
-        metadata_str = read_bytes_from_url(entry.metadata_url).decode("utf-8")
+        metadata_str = read_bytes_from_url(entry.metadata_url, self.gcs_fs).decode(
+            "utf-8"
+        )
         assert entry.entity in metadata_str
 
 
-def read_bytes_from_url(url):
+def read_bytes_from_url(url, gcs_fs):
```
jqmp, line 86: Oof, it's unfortunate that it's necessary to pass in `gcs_fs` here. Long-term, it seems like it would be nice if we could pass around a single object that knows how to access files for each protocol.

@namanjain Do you think it could make sense to have some sort of `MultiplexFileSystem` that wraps the others and dispatches each method to the right implementation based on the protocol? I realize that `fsspec` already has its own registry, but (a) it's global and (b) even if you could pass it around, you still have to do the work of extracting the protocol and looking up the filesystem before you can do your actual operation.

E.g., instead of this:

    protocol = protocol_from_url(url)
    fs = fsspec.filesystem(protocol)
    return fs.cat_file(url)

we could just do this:

    return mux_fs.cat_file(url)

tests/test_flow/test_copy.py
```diff
@@ -50,6 +51,40 @@ def dask_df():
     return builder.build()
 
 
+@pytest.fixture
```
jqmp, line 54: This is an interesting special case! I agree that we might as well handle this with patching. However, there a couple things about your implementation I don't understand:

1. Why do we need to return a context manager instead of doing the patching inside the fixture like we (used to) do elsewhere? I.e.,

    do_patching()
    yield None
    undo_patching()

2. Do we still need separate methods for `_upload_to_gcs` and `upload_to_gcs`? Could we accomplish the same thing with just a global `gcs_fs` object (like `_cached_gcs_fs`) that we patch out?

tests/helpers.py
```diff
@@ -232,25 +231,6 @@ def longest_regex_prefix_match(regex, string, flags=0):
     return match
 
 
-def gcs_fs_wipe_path(url):
-    assert "BNTESTDATA" in url
```
jqmp, line 236: I like that we're deleting these methods, but it's a bit a shame to be losing these asserts. Should we be worried about the risk of deleting the wrong files on GCS?

jqmp (commented):
Overall this looks good. Aside from some nits and possible simplifications, the only thing that worries me is the loss of the "BNTESTDATA" asserts. I'm not sure if it's a big deal or not, but they did make me feel better. What do you think?

bionic/aip/task.py
```diff
@@ -106,7 +107,7 @@ def _stage(self):
         path = self.inputs_uri()
         logging.info(f"Staging task {self.name} at {path}")
 
-        gcs_fs = get_gcs_fs_without_warnings()
+        gcs_fs = self.gcs_fs
```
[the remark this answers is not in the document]
simonafk: Good catch

tests/test_flow/conftest.py
```diff
@@ -163,7 +177,7 @@ def gcs_url_stem(request, use_fake_gcp):
 
 
 @pytest.fixture
-def session_tmp_gcs_url_prefix(gcs_url_stem, use_fake_gcp):
+def session_tmp_gcs_url_prefix(gcs_url_stem, gcs_fs):
     """
     Sets up and tears down a temporary "directory" on GCS to be shared by all
```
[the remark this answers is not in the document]
simonafk: Yeah the scope was changed so that it supports both fake and real GCS.

tests/test_flow/fakes.py
```diff
@@ -118,7 +116,7 @@ def glob(self, url):
         url = url[:-4]
         # Glob doesn't return the protocol, so we strip "gs://".
         urls = []
-        for file_url in self._files_by_url:
+        for file_url in self._files_by_url.keys():
```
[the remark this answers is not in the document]
jqmp: Ah, got it! That makes sense.

tests/test_flow/conftest.py
```diff
@@ -133,14 +133,28 @@ def gcs_builder(builder, tmp_gcs_url_prefix):
     builder.set("core__persistent_cache__gcs__object_path", object_path)
     builder.set("core__persistent_cache__gcs__enabled", True)
 
+    if use_fake_gcp:
```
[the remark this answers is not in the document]
jqmp: Ah, I see. That does sound like a good reason for the if statement. I'm still not sure about the assert statement, but it's not hurting anything.

tests/test_flow/conftest.py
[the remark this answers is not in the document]
jqmp: Does this mean we do the setup/teardown for each test? Is that expensive?

Also, I'm not following why the scope needs to change to support both real and fake GCS. What's the problem with using session scope? Since `use_fake_gcp` is parameterized, won't pytest still create two separate instances of this fixture?

tests/test_flow/fakes.py
```diff
@@ -205,28 +201,27 @@ class InstrumentedFilesystem:
     ("get") operations.
     """
 
-    def __init__(self, wrapped_fs, make_list):
+    def __init__(
+        self,
+        wrapped_fs,
+        shared_downloaded_list,
```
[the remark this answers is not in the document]
jqmp: Huh. I'm confused by how this would happen, but maybe I need to think more about this. When is the serialization of the `make_list` function happening?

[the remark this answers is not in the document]
simonafk: Hmm, I do not remember what issues I encountered earlier, but switching back to passing `make_list` works now.

tests/test_flow/conftest.py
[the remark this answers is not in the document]
simonafk: Modified this fixture to support real gcs only. This fixture now runs only if the `--bucket` is passed.

jqmp (approved):
LGTM!

tests/test_flow/test_cache_api.py
namanjain: I think that makes sense to use this mux fs in test code. But we should be careful about what we store in that mux fs since we will probably send it across to subprocesses for parallel processing.

jqmp: I agree, we'd want the mux FS to be easily transmittable to other processes. But is that an argument against using it for non-test code? It seems like the alternative would be to use two separate fsspec filesystems -- local and GCS -- and both of those would need to be transmitted as well, so grouping them together doesn't seem like an added cost. I had imagined that we could use this mux FS as a single interface for accessing the disk and GCS.

namanjain: Yeah, it doesn't seem too bad for now. I was a bit worried that future fs we use might not be serializable but I guess we can have custom pickle code in mux fs for non-pickleable fs.

jqmp: It sounds like fsspec has [serialisability](https://filesystem-spec.readthedocs.io/en/latest/features.html#serialisability) as an explicit goal, so hopefully this won't be a problem with any popular filesystems.

namanjain: TIL and that's awesome!

tests/test_flow/fakes.py
```diff
@@ -188,15 +192,7 @@ def create_aip_job(body, parent):
         "state": "SUCCEEDED"
     }
 
-    cached_gcs_fs = gcs._cached_gcs_fs
```
namanjain, line 191: Can you also get rid of `gcs._cached_gcs_fs` now?

tests/test_flow/conftest.py
```diff
@@ -121,7 +103,26 @@ def _make_list():
...
+
+    return _make_dict
+
+
+@pytest.fixture
+def gcs_fs(use_fake_gcp, make_dict):
+    if use_fake_gcp:
+        return FakeGcsFs(make_dict)
+    else:
+        return get_gcs_fs_without_warnings()
```
namanjain, line 121: Are you using this gcs fs instance anywhere? If not, consider calling this fixture `fake_gcs_fs` and use it only when gcs is faked.

tests/test_flow/test_persistence_gcs.py
```diff
@@ -7,33 +7,36 @@
...
-
 
 # This is detected by pytest and applied to all the tests in this module.
 pytestmark = pytest.mark.needs_gcs
 
 
-@pytest.fixture(scope="function")
-def preset_gcs_builder(gcs_builder):
+@pytest.fixture
+def instrumented_gcs_fs(gcs_fs, use_fake_gcp, make_list):
```
namanjain, line 30: What is `use_fake_gcp` used for? Is it still needed?

tests/helpers.py
simonafk: Left out by accident. `BNTESTDATA` check has been added back.

tests/test_flow/conftest.py
```diff
@@ -175,20 +189,18 @@ def session_tmp_gcs_url_prefix(gcs_url_stem, use_fake_gcp):
     gs_url = gcs_url_stem + "/" + path_str + "/"
     # This emits a stderr warning because the URL doesn't exist.  That's
```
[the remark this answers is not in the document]
simonafk: Hmm, I do not see any warnings when running pytest with `--bucket`.

tests/test_flow/test_persistence_gcs.py
simonafk: Good catch

tests/test_flow/conftest.py
simonafk: Yes, we are using `gcs_fs` in places runs on both real and fake gcs e.g. `instrumented_gcs_fs`.

tests/test_flow/conftest.py
```diff
@@ -133,69 +134,104 @@ def gcs_builder(builder, tmp_gcs_url_prefix):
...
     Sets up and tears down a temporary "directory" on GCS to be shared by all
-    of our tests.
+    of our tests. Not applicable for fake GCS.
     """
 
+    if real_gcs_url_stem is None:
+        yield None
+        return
+
+    assert real_gcs_url_stem is not None
```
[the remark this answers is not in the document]
simonafk: Good catch

tests/test_flow/fakes.py
simonafk: That would remove the caching of `fsspec.filesystem("gcs")`? Not exactly sure if we should do this.

PR #290, opened 2020-11-04
Janek/rename bootstrap

jqmp:
This renames the `Bootstrap` class to `ExecutionCore` (first commit), and then initializes this object with a default instance instead of `None` (second commit), which helps us avoid having a bunch of null checks in our code. This default value is referred to as the "bootstrap core" -- i.e., the core used for the bootstrapping process, which computes the real core.

### DIFF
--- a/bionic/deriver.py
+++ b/bionic/deriver.py
@@ -8,7 +8,7 @@ import warnings
 
 import attr
 
-from .datatypes import ResultGroup, DescriptorMetadata, TaskKey
+from .datatypes import ResultGroup, DescriptorMetadata, TaskKey, VersioningPolicy
 from .descriptors.parsing import entity_dnode_from_descriptor
 from .descriptors import ast
 from .deps.optdep import import_optional_dependency
@@ -63,9 +63,12 @@ class EntityDeriver:
         # Tracks whether we've pre-validated the base descriptors in this flow.
         self._base_prevalidation_is_complete = False
 
-        # The "core state" needs to be complete before we can compute user-defined
-        # entities.
-        self._core = None
+        # We need an ExecutionCore object to do any derivation, but creating the core
+        # itself requires derivation. We'll use this default core to get us through the
+        # initial bootstrapping phase, and then replace it with a properly configured
+        # core.
+        self._core_is_final = False
+        self._core = BOOTSTRAP_CORE
 
     # TODO We should adjust the wording of the docstring below or refactor this a
     # little. It's not necessary for *the user* to call this method, but it is necessary
@@ -77,7 +80,7 @@ class EntityDeriver:
         necessary but allows errors to surface earlier.
         """
         self._register_static_providers()
-        self._set_up_core()
+        self._set_up_final_core()
         self._prevalidate_base_dnodes()
 
     def derive(self, dnode):
@@ -301,12 +304,12 @@ class EntityDeriver:
         # behavior downstream.
         self._followup_dnode_lists_by_dnode = dict(followup_dnode_lists_by_dnode)
 
-    def _set_up_core(self):
+    def _set_up_final_core(self):
         """
         Initializes some key objects needed to compute user-defined entities.
         """
 
-        if self._core is not None:
+        if self._core_is_final:
             return
 
         self._core = ExecutionCore(
@@ -320,7 +323,9 @@ class EntityDeriver:
             should_persist_default=self._compute_core_entity(
                 "core__persist_by_default"
             ),
+            task_key_logging_level=logging.INFO,
         )
+        self._core_is_final = True
 
     def _prevalidate_base_dnodes(self):
         """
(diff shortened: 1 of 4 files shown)


bionic/deriver.py
```diff
@@ -301,28 +304,28 @@ def register_provider(provider):
...
         self._followup_dnode_lists_by_dnode = dict(followup_dnode_lists_by_dnode)
 
-    def _set_up_bootstrap(self):
+    def _set_up_core(self):
         """
         Initializes some key objects needed to compute user-defined entities.
         """
 
-        if self._bootstrap is not None:
+        if self._core_is_finalized:
```
namanjain, line 312: If we are using `finalize` to describe this variable, should we also use `finalize` for this method instead of `set_up`?

Also, since we have a default core, should this be a property of `ExecutionCore` instead?

namanjain (approved):
This definitely looks cleaner than before, thanks!

bionic/deriver.py
```diff
@@ -301,28 +304,28 @@ def register_provider(provider):
         # behavior downstream.
         self._followup_dnode_lists_by_dnode = dict(followup_dnode_lists_by_dnode)
 
-    def _set_up_bootstrap(self):
+    def _set_up_core(self):
```
simonafk, line 307: How about using the term `initialize` instead of `set_up` or `finalize`?

bionic/deriver.py
```diff
@@ -63,9 +63,12 @@ def __init__(self, flow_config, flow_instance_uuid):
...
         self._base_prevalidation_is_complete = False
 
-        # The "bootstrap state" needs to be complete before we can compute user-defined
-        # entities.
-        self._bootstrap = None
+        # We need an ExecutionCore object to do any derivation, but creating the core
+        # itself requires derivation. We'll use this default core to get us through the
+        # initial bootstrapping phase, and then replace it with a properly configured
+        # core.
+        self._core_is_finalized = False
```
simonafk, line 70: What about having an `is_initialized()` method in `ExecutorCore`?

bionic/deriver.py
```diff
@@ -301,28 +304,28 @@ def register_provider(provider):
...
 
-        self._bootstrap = Bootstrap(
-            persistent_cache=self._bootstrap_singleton_entity("core__persistent_cache"),
-            versioning_policy=self._bootstrap_singleton_entity(
-                "core__versioning_policy"
-            ),
-            aip_executor=self._bootstrap_singleton_entity("core__aip_executor"),
-            process_executor=self._bootstrap_singleton_entity("core__process_executor"),
-            should_memoize_default=self._bootstrap_singleton_entity(
+        self._core = ExecutionCore(
```
simonafk, line 315: It seems to me that we are doing a 2-step initialization? How about having an `initialize` method in `ExecutorCore`? That would make `ExecutorCore` not immutable, but we can add checks that verify the class is initialized before certain properties can be returned.

bionic/deriver.py
jqmp: I agree, we can improve this name. What do you think about `set_up_final_core`?

I'm not so sure about making `is_finalized` a property of `ExecutionCore` though. I originally had it this way, but I moved it out. My thinking is, we need to put this boolean somewhere, but putting it in `EntityDeriver` hides it from the other classes that use `ExecutionCore`. In particular, the fact that the deriver has two stages and switches cores doesn't seem like a detail that needs to be exposed to the logic in `flow_execution` and `task_execution`.

@simonafk For the same reason, having an `initialize` method that mutates the core doesn't seem very attractive to me. Currently the core is mutable from the POV of the `EntityDeriver` but immutable from the POV of the execution code; with this change it would be mutable everywhere. WDYT?

bionic/deriver.py
jqmp: The reason I prefer `finalize` is that `initialize` suggests the core isn't usable if it's not initialized. (That's how we use "initialize" elsewhere.) In this case, there is a usable core, just not the best one. I'm open to other terms though.

bionic/deriver.py
namanjain: `set_up_final_core` reads cleaner to me than just `set_up_core`. But in that case, wdyt about renaming this variable `_final_core_is_set_up`? I think that if that doesn't sound clean either, then maybe we aren't using the right verbs.

jqmp: How about `_core_is_final`?

bionic/deriver.py
```diff
@@ -684,9 +680,13 @@ def _compute_result_group_for_dnode(self, dnode):
 
 
 @attr.s(frozen=True)
-class Bootstrap:
+class ExecutionCore:
```
simonafk, line 683: What do you think of `CoreEntities`? We use the term `core entities` in some places but not `execution`?

simonafk: `ExecutionCore` seems fine.

I think my concern is more about consistency in naming and comments.

jqmp: `CoreEntities` is not a bad name but I prefer to avoid using plurals in class names, since a variable called `_core_entities` would look like a collection. On the other hand, it's hard to think of a singular noun that makes sense here. "Core" was the best one I could think of, and I added "Execution" because that's what it's used for. (We don't use the term "execution" consistently but so far I think it's the best term for what happens in `flow_execution.py` and `task_execution.py`.) Note also that not every built-in entity appears in the `ExecutionCore`, and one of the core fields (`task_key_logging_level`) is not an entity at all.

Our use of "core" and "internal" is indeed a bit confusing. I think we currently tend to use "internal entity" to describe any built-in entity, but they were originally named as `core__...` and it's hard to change user-facing names, so we haven't done that (yet).

After writing all this, given that we're stuck with the user-facing term "core" for the time being, I wonder if re-using the term in our code is unnecessarily confusing. A similar term would be "nucleus" -- it sounds a bit weird but at least it doesn't have any other connotations. Other possibilities could be "kernel" or "infrastructure", but they're both suggestive of other concepts.

What do you all think?

jqmp: I tried renaming "core" to "nucleus" but ended up liking it less -- passing around a "nucleus" feels weirder and less intuitive to me than a "core". For now I'm leaving it as "core", but I'm open to revisiting it.

PR #291, opened 2020-11-06
Extract temporary in-memory caching to central object

jqmp:
This PR introduces an ExecutionContext class that bundles together a few commonly-used-together objects, and then moves our temporary result caching off of the individual TaskRunnerEntries and onto the ExecutionContext. More details in the commit messages.

### DIFF
--- a/bionic/core/flow_execution.py
+++ b/bionic/core/flow_execution.py
@@ -7,6 +7,8 @@ from concurrent.futures import wait, FIRST_COMPLETED
 import logging
 import time
 
+import attr
+
 from .task_execution import (
     EntryLevel,
     EntryPriority,
@@ -54,11 +56,9 @@ class TaskCompletionRunner:
     enum, indicating whether we need to and/or can do more work on it.
     """
 
-    def __init__(self, core, flow_instance_uuid, task_key_logger):
+    def __init__(self, context):
         # These are needed to complete entries.
-        self._core = core
-        self._flow_instance_uuid = flow_instance_uuid
-        self.task_key_logger = task_key_logger
+        self._context = context
 
         # These are used for caching and tracking.
         self._entries_by_task_key = {}
@@ -68,16 +68,16 @@ class TaskCompletionRunner:
 
     @property
     def _parallel_execution_enabled(self):
-        return self._core.process_executor is not None
+        return self._context.core.process_executor is not None
 
     @property
     def _aip_execution_enabled(self):
-        return self._core.aip_executor is not None
+        return self._context.core.aip_executor is not None
 
     def run(self, states):
         try:
             if self._parallel_execution_enabled:
-                self._core.process_executor.start_logging()
+                self._context.core.process_executor.start_logging()
 
             for state in states:
                 entry = self._get_or_create_entry_for_state(state)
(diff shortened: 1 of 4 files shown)


bionic/core/flow_execution.py
```diff
@@ -588,5 +587,22 @@ def log_computed_aip(self, task_key):
         self._log("Computed   %s using AIP", task_key)
 
 
+# TODO Consider introducing the term "query" to refer to the scope of one get() call.
```
namanjain, line 594: I think `Context` works well for this scope.

bionic/core/flow_execution.py
```diff
@@ -599,10 +603,32 @@ class ExecutionContext:
     flow_instance_uuid = attr.ib()
     core = attr.ib()
     task_key_logger = attr.ib()
+    # This is used for temporarily saving results for the duration of one execution.
+    # Currently we only do this for entities where both persistence and memoization are
+    # disabled.
+    temp_result_cache = attr.ib()
```
namanjain, line 609: Given that this cache lives inside `ExecutionContext`, I think it's clear that this is a temporary cache. IMO, you can drop `temp`.

bionic/core/flow_execution.py
```diff
@@ -599,10 +603,32 @@ class ExecutionContext:
...
 
     def evolve(self, **kwargs):
         return attr.evolve(self, **kwargs)
 
 
+class MemoryResultCache:
```
namanjain, line 615: It would be cool to use this for our "flow scoped" in memory caching as well.

namanjain (approved):
Left a few minor comments, LGTM otherwise!

simonafk (approved):

bionic/core/flow_execution.py
```diff
@@ -278,14 +278,17 @@ def _process_active_entry(self, entry):
                 remote_subgraph.get_stripped_state(target_entry.state)
                 for target_entry in target_entries
             ]
-            new_core = self._core.evolve(aip_executor=None, process_executor=None)
-            new_task_completion_runner = TaskCompletionRunner(
+            new_core = self._context.core.evolve(
```
simonafk, line 281: nit: comments on why we are removing the executors this would be nice.

bionic/core/flow_execution.py
jqmp: I think `Context` is a good name for the kind of object it is, but I'm not sure it makes the scope clear. Also, if we have other classes with the same scope, it would be nice if they all had the prefix. (E.g., `TaskCompletionRunner`, `TaskRunnerEntry`, and `ExecutionContext` all have the same scope, so it would be nice if they all started with `Execution` or `Query`. Likewise, `Flow`, `EntityDeriver`, `FlowConfig`, and `TaskState` also share a scope and could also have the same prefix. (`Flow`?))

bionic/core/flow_execution.py
jqmp: Agreed. I see you added it to one of your PRs too.

bionic/core/flow_execution.py
jqmp: I guess it doesn't seem super obvious to me, especially since the other memory cache is handled in a different way. For example, in the line

    self.context.temp_result_cache.save(result)

I think the `temp_` makes it clearer that this is distinct from the main memory cache.

jqmp:
I was planning a followup to this PR that needed this cache to be centralized, but that followup didn't work out the way I hoped. I'm currently working on a different direction that may not require this. I'm on the fence about merging this on its own; I think it's probably still the right direction, but I'm not 100% sure it's helpful to do it if we don't centralize some of the TaskState fields as well. WDYT?

PR #292, opened 2020-11-09
Revert "Restrict cattrs version on Python 3.6"

simonafk:
This reverts commit 9866f0e0a4eda1eeee5e534813a0847baf22f143.

The commit above was added because the latest version of cattrs dropped
support for Python 3.6. cattrs now uses python_requires so that Python
3.6 will fall back to an older version. Hence, the commit above is no
longer necessary. It also breaks Dockerfile script.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,10 +17,7 @@ with open("README.md") as readme_file:
 
 requirements = [
     "attrs>=20.1",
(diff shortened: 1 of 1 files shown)


jqmp (approved):
LGTM!

PR #294, opened 2020-11-11
Check instances for uniqueness in GC tests

jqmp:
I added a check to our tests to make sure that when we load a
(non-memoized) value from disk, the original value is no longer in
memory. (I hope to change our caching logic soon and don't want to
accidentally lose this property.)

### DIFF
--- a/tests/test_flow/test_persistence.py
+++ b/tests/test_flow/test_persistence.py
@@ -1361,6 +1361,11 @@ def make_tracked_class():
     """
     Creates a "tracked" class which keeps a count of how many of its instances are in
     memory.
+
+    The class also asserts that each instance in memory is has a unique value; this
+    guarantees that we don't (for example) read an instance from disk while the
+    original instance is still in memory.
+
     """
 
     def _make_tracked_class():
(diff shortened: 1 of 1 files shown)


tests/test_flow/test_persistence.py
```diff
@@ -1378,14 +1383,22 @@ def read(self, path):
 
         class Tracked:
             n_instances_in_memory = 0
```
simonafk, line 1385: Do we still need this variable since you can use the set to get the count?

simonafk (approved):

namanjain (approved):

tests/test_flow/test_persistence.py
jqmp: Yeah, I was wondering about whether to do that. I didn't want to convert `n_instances_in_memory` to a function, but having a property on a class is a little awkward too. (Looks like you can do it like this: https://stackoverflow.com/questions/128573/using-property-on-classmethods)

I don't feel strongly about it either way; what do you all think?

PR #301, opened 2020-11-30
Group tuple nodes visually

jqmp:
This commit updates our DAG visualization to group together any nodes
generated by a common function (i.e., a tuple-generating function). This
means the nodes will be visually aligned and have the same color. I
think this clarifies the relationship between these nodes.

(We already do this for nodes that have the same descriptor but
different case keys; this just extends the grouping logic to nodes that
share entity names.)

Before:

image

After:

image

(`train_frame` and `test_frame` are now grouped together.)


bionic/dagviz.py
```diff
@@ -91,14 +91,39 @@ def dot_from_graph(graph, vertical=False, curvy_lines=False, name=None):
...
-        node_lists_by_cluster[descriptor].append(node)
-
-    descriptors = list(set(graph.nodes[node]["descriptor"] for node in graph.nodes()))
-    color_strs_by_descriptor = hpluv_color_dict(
-        descriptors, saturation=99, lightness=90
+        entity_names = list(node.dnode.all_entity_names())
+        # If this node has no entity names at all, we'll pretend it has a single entity
+        # name. (This is unlikely to happen in practice, but could happen for a
+        # descriptor like the empty tuple `(,)`.)
+        if len(entity_names) == "":
```
namanjain, line 103: I think this should be `len(entity_names) == 0`

bionic/dagviz.py
```diff
@@ -91,14 +91,39 @@ def dot_from_graph(graph, vertical=False, curvy_lines=False, name=None):
...
+                first_cluster.extend(other_cluster)
+                node_clusters_by_entity_name[other_entity_name] = first_cluster
+
+        first_cluster.append(node)
+
+    # Now we deduplicate the clusters and arrange them in a deterministic order.
+    unique_node_clusters_by_obj_id = {
+        id(cluster): cluster for cluster in node_clusters_by_entity_name.values()
+    }
+    sorted_node_clusters = sorted(unique_node_clusters_by_obj_id.values(), key=min)
```
namanjain, line 121: Why are we sorting the clusters now whereas we were not sorting them before?

namanjain (approved):

bionic/dagviz.py
jqmp: Yup! Good catch.

bionic/dagviz.py
jqmp: We want the assignment of colors to clusters to be deterministic. Previously we clustered by descriptor and then assigned a color to each descriptor; now we assign each cluster an arbitrary index and then assign colors to the indices. Sorting the clusters first (using each cluster's first descriptor as a key) means we always assign the same index to each cluster.

I'll add a comment to make this clearer.

PR #302, opened 2020-12-01
better parallel execution fail logging

simonafk:
Use logger.exception to log parallel/aip job failure(s). Stack trace(s)
are printed and this would help debugging.

### DIFF
--- a/bionic/core/flow_execution.py
+++ b/bionic/core/flow_execution.py
@@ -303,7 +303,10 @@ class TaskCompletionRunner:
                 )
 
                 def done_callback(callback_future):
-                    if not callback_future.cancelled():
+                    if (
+                        not callback_future.cancelled()
+                        and callback_future.exception() is None
+                    ):
                         for target_entry in target_entries:
                             self._context.task_key_logger.log_computed_aip(
                                 target_entry.state.task_key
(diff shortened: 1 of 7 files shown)


bionic/core/flow_execution.py
```diff
@@ -407,47 +410,54 @@ def _mark_entry_active(self, entry):
 
     def _wait_on_in_progress_entries(self):
         "Waits on any in-progress entry to finish."
-        futures = set(entry.future for entry in self._in_progress_entries)
-        finished_futures, _ = wait(futures, return_when=FIRST_COMPLETED)
-        try:
-            for finished_future in finished_futures:
+        future_to_entry = {entry.future: entry for entry in self._in_progress_entries}
```
jqmp, line 413: Nit: the convention for dicts here would be `entries_by_future`.

bionic/aip/state.py
```diff
@@ -2,7 +2,9 @@
 
 
 class AipError(Exception):
-    pass
+    def __init__(self, job_id, error_message):
+        self.job_id = job_id
```
jqmp, line 6: When this exception is converted to a string, I think it will be rendered as a tuple: `(job_id, error_message)`. You probably want to provide a human-readable message, either by calling the superclass constructor (like `CodeVersioningError` does) or by implementing your own `__str__` method.

tests/test_flow/conftest.py
```diff
@@ -258,20 +258,32 @@ def __init__(self, caplog):
...
 
     def expect_all(self, *expected_messages):
-        actual_messages = self._pop_messages()
+        actual_messages = self._pop_messages(include_exception=False)
         assert set(actual_messages) == set(expected_messages)
         self._caplog.clear()
 
-    def expect_regex(self, *expected_patterns):
-        actual_messages = self._pop_messages()
+    def expect_regex(self, *expected_patterns, include_exception=False):
```
jqmp, line 265: Is there any reason not to always include exceptions?

tests/test_flow/fakes.py
```diff
@@ -164,14 +148,13 @@ def _run_aip_job(self, body):
...
-        # AIP execution does not send back logs to local bionic instance. Hence,
-        # all logs are filtered here.
-        with temporarily_filter_all_logs():
-            try:
-                run_aip(input_uri, self._gcs_fs)
-                return "SUCCEEDED", None
-            except Exception as e:
-                return "FAILED", str(e)
+        try:
+            # TODO: filter out logs. AIP execution does not send back logs to
```
jqmp, line 152: Why this change?

tests/test_flow/conftest.py
simonafk: Good point. Changed to always include exceptions.

bionic/aip/state.py
simonafk: Reverted changes to this class.

tests/test_flow/fakes.py
simonafk: It may not be working correctly as expected. In addition, the Syncmanager Barrier was also not working correctly; all the entities in `test_parallel_fail`, including the ones that do not throw exceptions, were failing. Hence, I have changed the fake AIP to use a process executor instead of a thread executor.

bionic/core/flow_execution.py
simonafk: Good point. Renamed variable.

tests/test_flow/fakes.py
```diff
@@ -122,21 +121,6 @@ def pipe(self, url, content_bytes):
...
-    logger = logging.getLogger()
-    filter_all_logs = FilterAllLogs()
-    try:
-        logger.addFilter(filter_all_logs)
-        yield
-    finally:
-        logger.removeFilter(filter_all_logs)
-
-
 class FakeAipClient:
```
namanjain, line 124: This is a lot easier to read now!

bionic/core/flow_execution.py
```diff
@@ -302,7 +302,10 @@ def _process_active_entry(self, entry):
                 )
 
                 def done_callback(callback_future):
-                    if not callback_future.cancelled():
+                    if (
+                        not callback_future.cancelled()
+                        and not callback_future.exception()
```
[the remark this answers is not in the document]
jqmp: +1 to `is None`, if that's what we're checking. That's what we tend to do elsewhere in Bionic.

bionic/core/flow_execution.py
```diff
@@ -407,47 +410,56 @@ def _mark_entry_active(self, entry):
...
+                )
+                finished_futures, _ = wait(
+                    entries_by_future.keys(), return_when=ALL_COMPLETED
+                )
+
+                # There may be errors in the other tasks. Ensure that we log all
+                # of the other errors as well.
+                for future in finished_futures:
+                    try:
+                        future.result()
```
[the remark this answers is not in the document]
simonafk: If the results are stored in GCS / disk, then the entities should not be recomputed.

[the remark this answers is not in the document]
namanjain: That is the ideal behavior and is true if we make a call again from a new flow. But if you use the same flow, I'm pretty sure that the task state's cache accessor will contain an empty (local or cloud) entry and will think that the results are not cached.

tests/test_flow/fakes.py
```diff
@@ -150,28 +134,18 @@ def _create_aip_job(self, body, parent):
...
 
-        # AIP execution does not send back logs to local bionic instance. Hence,
-        # all logs are filtered here.
-        with temporarily_filter_all_logs():
-            try:
-                run_aip(input_uri, self._gcs_fs)
-                return "SUCCEEDED", None
-            except Exception as e:
-                return "FAILED", str(e)
+        self._jobs[body["jobId"]] = ProcessPoolExecutor(max_workers=1).submit(
```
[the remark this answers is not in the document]
simonafk: I am trying to keep this simple. Also, executors cannot be stored in `FakeAipClient` as executors are not serializable.

bionic/core/flow_execution.py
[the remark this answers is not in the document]
simonafk: Fixed to use `is None`.

bionic/core/flow_execution.py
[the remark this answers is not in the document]
simonafk: Modified to sync after getting the results.

tests/test_flow/test_executor.py
```diff
@@ -114,7 +114,7 @@ def test_parallel_fail(aip_builder, make_counter, multiprocessing_manager, log_c
 
     # The barrier ensures that all the entity functions do not complete unless
     # all of them are started.
-    barrier = multiprocessing_manager.Barrier(4, timeout=120)
+    barrier = multiprocessing_manager.Barrier(4, timeout=240)
```
jqmp, line 117: From our conversation, it sounds like we only have a high timeout for debugging purposes, and we should never hit this normally. Maybe we can add a comment explaining this?

tests/test_flow/fakes.py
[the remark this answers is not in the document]
jqmp: Just documenting what we all discussed in a hangout:

1. If you create the AIP client entity with a function (instead of using `builder.set`) and disable persistence, then it shouldn't need to be parallelizable.
2. Let's try to check whether creating the ProcessPoolExecutor is expensive or not. If so, we can look into caching it.

jqmp (approved):

bionic/core/flow_execution.py
```diff
@@ -407,47 +410,58 @@ def _mark_entry_active(self, entry):
...
-                            f"""
-                        Encountered another exception while doing remote
-                        computation: {e!r}.
+        entries_by_future = {entry.future: entry for entry in self._in_progress_entries}
+        finished_futures, _ = wait(
+            entries_by_future.keys(), return_when=FIRST_COMPLETED
+        )
+        for finished_future in finished_futures:
+            try:
+                self._sync_remotely_computed_task_keys(finished_future.result())
```
jqmp, line 419: It might be good to include the word `complete` in this method name -- the fact that we are transitioning the entries from `IN_PROGRESS` to `COMPLETED` is important.

simonafk: I can't think of a good way to rename this method without having a very long name. E.g. `_sync_remotely_computed_task_keys_and_mark_complete_if_possible`. or `_sync_remotely_computed_and_mark_complete_if_possible`

jqmp: In this case, it will always be possible to mark them complete, right? So we could write `_sync_and_complete_remotely_computed_task_keys`. ("mark completed" would be more precise but this seems like an okay compromise to me.)

PR #303, opened 2020-12-02
Refactor random tests

jqmp:
This moves some code out of ModelFlowHarness and into ModelBinding; now all the model and user-code logic is in the binding class, while the harness class is only responsible for coordinating between the model and the real flow.

(I also looked at moving all the model state out into a separate `ModelFlow` class, but it didn't seem like a big improvement.)


namanjain (approved):
It's too bad that `ModelFlow` didn't turn out that well, but this change still has a cleaner separation of responsibilities compared to before.

Issue #296, opened 2020-11-16
Uncached values should be recomputed each time they're used


jqmp:
Currently, when an entity has both persistence and memoization disabled, we store its value in memory for the duration the `Flow.get()` call (as documented [here](https://bionic.readthedocs.io/en/stable/concepts.html#disabling-in-memory-caching)). I think we should change this behavior: instead of storing it in memory, we should recompute the value each time it's used.

I believe the current behavior was chosen because it was easier to implement. However, it seems more intuitive to me that if memoization is disabled, then we don't store the value in memory any longer than necessary. This also matches the intuition of data scientists I've spoken with. (If you're at Square, you can search Slack for the string "it sounds like everyone here would prefer to have it recompute".) Finally, it's slightly confusing that turning off persistence will change the amount of memoization.

This would be a breaking change. Although I doubt many users depend on the current behavior, we should make sure to check in with users before finalizing the change.

jqmp:
[This PR](https://github.com/square/bionic/pull/295) has a concrete example. In this test case, `x` is non-memoized and non-persisted and is used by two other entities (`x_squared` and `y_x_cubed`). With the current behavior, we store its value in memory until the `get` call is finished, so `x` is only computed once per call. After this change, `x` would be computed twice.

PR #304, opened 2020-12-03
Add generic eviction code

jqmp:
This generalizes our "vacate" code to make it safe to use on any
temporarily memoized value, not just drafts. Now temporarily memoized
values can be evicted from memory and recomputed later if necessarily.

(This commit doesn't change any behavior; it just makes it possible to
introduce changes later.)

Motivation
----------

There are two motivations for this:

1. I want to split out the persistent saving and loading operations into
separate tasks, so that instead of a persistent entity being computed
like this:

    <x>           -> x
    (draft value) -> (final value)

it will be computed like this:

    <x>           -> x/artifact -> x
    (draft value) -> (artifact) -> (final value)

The catch is, if `x` is not memoized, we want it to be loaded from disk
each time it's needed; that is, we want the loading function `x/artifact
-> x` to be recomputed each time. That means we need a way to evict `x`
from memory as soon as it gets used (but not any sooner!).

2. Recomputing uncached values each time they're used (described
[here](https://github.com/square/bionic/issues/296)).

Implementation
--------------

Both of these goals require us to be able to: compute an entry value;
hold it in memory long enough for a dependency to access it; delete it;
and then recompute it again if necessary. (Also, if the value is a
tuple, we need to make sure to keep it around just long enough for all
of its followup tasks to run. If it weren't for this extra constraint,
we could solve this by just computing these uncached values (and their
uncached dependencies) lazily each time; but that would mean any task
that produces a pair of values would always get run twice.)

Previously the only values we evicted from memory were draft values.
This was easy to do because we know that draft values are only ever
accessed by their followup tasks, so once the followups are done, it's
safe to delete the draft value forever. However, now we want
to delete values that might be accessed later, so we need to be able to
recompute them on demand. This required the following changes:

1. We need a mechanism to delete requirements that are no longer needed.
We delete a requirement if (a) the requirement's source entry is fully
completed (`CACHED`), or (b) the requirement is marked "transient" and
its destination is fully completed. ("Transient" requirements are for
things like followup tasks, where we need an entry to be completed but
there's no other entry waiting for it.)

2. We now have one extra priority level: some followup tasks are not
just `HIGH` but `TOP` priority. The idea here is that if you have a
tuple, it's more urgent to break it into smaller pieces than it is to
process those individual pieces. This approach helps us get those pieces
out of memory as early as possible.

3. We need to actually delete each temporarily-memoized value from
memory at the right time. This can be either right after a child task
uses the value, or right after computing a value that no one is waiting
for (which can happen with followup tasks). Note that we don't do this
eviction for actual entity values, because that would be a breaking
change; we still just do this for draft values. We use a new
`should_memoize_for_query` field on `DescriptorMetadata` to determine
which values we should do this for.

### DIFF
--- a/bionic/core/task_execution.py
+++ b/bionic/core/task_execution.py
@@ -18,11 +18,12 @@ in the correct order; thus, the TaskRunnerEntry mostly contains data pertaining 
 task's relationship to other tasks.
 """
 
-import attr
 import copy
-import logging
-
 from enum import auto, Enum, IntEnum
+import logging
+from typing import Optional
+
+import attr
 
 from ..datatypes import Artifact, Result
 from ..exception import (
@@ -73,7 +74,7 @@ class TaskRunnerEntry:
 
         self.incoming_reqs = set()
         self.outgoing_reqs = set()
-        self.priority = EntryPriority.DEFAULT
+        self.priority = EntryPriority.NORMAL
 
     @property
     def stage(self):
@@ -206,6 +207,7 @@ class TaskRunnerEntry:
         # exactly the same value.)
         elif state.should_memoize:
             state._result = result
+
         else:
             self.context.temp_result_cache.save(result)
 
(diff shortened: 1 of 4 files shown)


bionic/core/task_execution.py
```diff
@@ -313,16 +278,21 @@ class EntryRequirement:
...
+    its ``src_entry`` has reached the (final) ``CACHED`` level. Once a requirement is
+    discarded, it's permissible for ``dst_entry`` to revert to a previous level. If a
+    requirement is not transient and has no source, it is assumed to be permanent.
     """
 
-    src_entry = attr.ib()
-    dst_entry = attr.ib()
-    level = attr.ib()
+    dst_entry: TaskRunnerEntry = attr.ib()
+    level: "EntryLevel" = attr.ib()
```
simonafk, line 293: If this indicates the requirements of `dst_entry`, what about adding a `dst_` prefix?

jqmp: I think this could make sense. It might be better to do it in a separate PR though. Let's discuss it in our meeting today.

bionic/core/task_execution.py
```diff
@@ -313,16 +278,21 @@ class EntryRequirement:
...
+    src_entry: Optional[TaskRunnerEntry] = attr.ib()
+    is_transient: bool = attr.ib()
```
namanjain, line 295: This might be only me, but I find `transient` hard to understand since it isn't clear when it disappears. It seems like you are using `transient` to tell if only `dst` cares about the requirement. When it is not transient, it means that either `src` cares about it, or the `dst` was requested (so deriver cares about it). How do you feel about encoding these two properties separately instead of combining them in one property (which is `is_transient`)?

namanjain (approved):
The changes looks good logically. It's really exciting that you could generalize it well enough that it can be used to further decompose tasks into micro-tasks.

But I find the code a little confusing with the use of transient. It's another one of those which even though I understand it now, I'll need to read the whole flow when I look at it again.

bionic/core/task_execution.py
```diff
@@ -313,16 +278,21 @@ class EntryRequirement:
...
-    A requirement indicates that one entry (`src_entry`) can't make any further progress
-    until another entry (`dst_entry`) has reached a certain level of progress (`level`).
+    A requirement indicates that we need a particular entry (``dst_entry``) to reach a
+    certain level of progress (``level``). This could be because it's blocking some
+    other entry (``src_entry``), or for some external or a priori reason.
 
-    Note: `src_entry` can also be `None`, indicating some sort of external or a priori
-    requirement. On the other hand, `dst_entry` cannot be `None`.
+    An entry can be marked ``is_transient``, meaning it can safely be discarded once
+    ``dst_entry`` has reached the target ``level``. Otherwise, it also be discarded if
```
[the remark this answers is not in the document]
jqmp: Yes, "otherwise" was meant to mean "if it's not transient". I'll make that clearer.

bionic/core/task_execution.py
jqmp: As we discussed in a hangout, I'll try replacing this with some kind of `valid_until` enum.

jqmp: I replaced `is_transient` with an `expiration` enum field.

namanjain: Thanks! I find the code a lot easier to read now.

PR #305, opened 2020-12-03
better aip job naming

simonafk:


jqmp (approved):
LGTM!

jqmp:
Now that we've added new logging tests here, is there anything else from https://github.com/square/bionic/pull/281 that we want to keep?

simonafk:
> Now that we've added new logging tests here, is there anything else from #281 that we want to keep?

Closed #281. Most of the changes has been imported here and in other PRs.

PR #306, opened 2020-12-10
Speed up fake AIP tests

jqmp:
These tests currently take 30 seconds because we have a fixed 10-second
polling period, which makes sense for real AIP but not for our fake
version. I've made this period configurable and set it to 0.1 seconds
for fake AIP.

### DIFF
--- a/bionic/aip/task.py
+++ b/bionic/aip/task.py
@@ -34,16 +34,26 @@ class Config:
     Attributes
     ----------
     uuid: str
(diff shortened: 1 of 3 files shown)


jqmp:
@simonafk does this make sense, or do you think we should do this a different way?

simonafk (approved):

simonafk:
> @simonafk does this make sense, or do you think we should do this a different way?

Looks good to me

Issue #307, opened 2020-12-14
Set-valued entities are not hashed determistically


jqmp:
After computing any entity value, Bionic serializes the value and then hashes it to check if any downstream entities need to be recomputed. However, if the value is a Python set, this hash is not deterministic: we serialize the set by pickling it, which amounts to just iterating through the set and pickling each value, and iteration order on sets is non-deterministic.

We should probably have a special protocol for sets (and frozensets) which does something deterministic, if that's possible. (I'm not sure if it's always possible, since some sets can't be sorted.)

PR #308, opened 2020-12-14
Add deterministic protocol for sets

jqmp:
This mostly fixes https://github.com/square/bionic/issues/307 by adding
a special protocol for sets (and frozensets) that sorts them before
hashing them.

Unfortunately this doesn't work for other kinds of objects that
themselves contain sets. Short of writing our own pickle-like algorithm
I'm not sure there's a good way to handle those.


bionic/protocols.py
```diff
@@ -599,6 +599,69 @@ def read(self, path):
...
+
+    def validate(self, value):
+        assert isinstance(value, (set, frozenset))
+
+    def write(self, value, path):
+        original_type = type(value)
+        try:
+            sorted_contents = sorted(value)
+        except TypeError as e:
+            message = f"""
```
namanjain, line 639: If sorting on the set fails, another alternative is to convert all elements to bytes using `pickle.dumps` and sort the bytes. This would require some extra pickling and unpickling logic, but we can be sure that any picklable set can use `PicklableSetProtocol`.

bionic/protocols.py
```diff
@@ -599,6 +599,69 @@ def read(self, path):
         return single_element(path.iterdir())
 
 
+# This protocol allows us to hash sets deterministically.
+# It would be preferable if PicklableProtocol could install a special handler for sets,
```
namanjain, line 603: I'm not 100% sure if this will work, but is it possible to use [`dispatch_table`](https://python.readthedocs.io/en/stable/library/pickle.html#pickle.Pickler.dispatch_table) to install a handler for sets?

jqmp: Unfortunately not. I don't think either of the override mechanisms works for built-in types. (The pure Python implementation of `pickle` has a separate undocumented dispatch table for built-in types that you can modify, but in the faster C implementation (which we use) the behavior for sets is hardcoded.)

bionic/protocols.py
jqmp: That's a great idea. I can think of two ways to do this (that don't involve redundant pickling):

1. Pickle all the elements, sort the resulting collection of bytes objects, then pickle that sorted collection.
2. Create a directory and pickle each element to a file in the directory. Then somehow sort the files (by hash or by contents) at some point during saving or loading.

1 is simpler but requires making an in-memory copy of the whole thing. WDYT?

bionic/protocols.py
namanjain: I saw the implementation too and now see what you mean by it. Technically, we can subclass the pickler in Bionic and modify `self.dispatch` variable to override set behavior, but that seems risky given that variable isn't documented. I guess this is fine then.

bionic/protocols.py
namanjain: I think the extra space in 1 is okay since we are not going to keep the in-memory copy around for long. My hesitation with 2 is that it might get tricky creating digest hash for these files (since we'll have to somehow sort them).

jqmp: I'm always reluctant to make a change that could potentially double the peak memory usage of a flow, but I agree that 2 is tricky. I'll try and see if there's a simple-ish way to do 2.

bionic/protocols.py
jqmp: Yeah, and the subclassing approach only works for the pure Python implementation, so all of our pickling would get slower.

bionic/protocols.py
jqmp: I tried number 2 and it didn't seem too bad. Let me know what you think.

namanjain: I didn't realize that our tokenization function accepts a folder too. This was easier than I expected. But this will also convert `frozenset` to `set`. I think we should pickle and store the type of the set too. What do you think?

jqmp: Good catch! That should have a test!

(Actually I originally considered having sets be deserialized as frozensets, because no one should be mutating them anyway, but I thought it was also potentially confusing. Could still be convinced though.)

jqmp: I fixed this and parametrized one of the tests to check frozensets as well.

namanjain (approved):
LGTM pending a minor comment!

bionic/protocols.py
```diff
@@ -599,6 +600,86 @@ def read(self, path):
...
+                pickle.dump(item, file_, protocol=self._pickle_protocol_version)
+            # After creating the file, we hash its contents and use that as the new
+            # name. However, it's possible for two objects in a set to generate the same
+            # hash (see `test_set_with_duplicate_pickle_output` for an example) so we
+            # need to add a unique suffix as well.
+            item_file_hash = hexdigest_from_path(item_file_path)
+            suffix = 0
+            while True:
+                new_item_file_path = items_path / f"{item_file_hash}_{suffix}.pkl"
+                if not new_item_file_path.exists():
```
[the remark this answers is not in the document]
jqmp: This can happen when two objects have the same pickle output, as described in the comment above. That's why we have a loop: so we can keep trying new paths until we find one that doesn't exist yet.

I'll try to make the comment clearer.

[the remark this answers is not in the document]
jqmp: I ended up rewriting the code in a way that's hopefully clearer.

simonafk (approved):

PR #311, opened 2020-12-15
Improve error messages from log regex tests

jqmp:
When `LogChecker.expect_regex` fails, it now prints a list of the
messages and regexes that failed to match; this makes failures much
easier to debug.

### DIFF
--- a/tests/test_flow/test_persistence_aip.py
+++ b/tests/test_flow/test_persistence_aip.py
@@ -67,7 +67,7 @@ def test_aip_jobs(aip_builder, log_checker):
     log_checker.expect_regex(
         r"Staging AI Platform task .* at gs://.*bionic_y1.*",
         r"Started AI Platform task: https://console.cloud.google.com/ai-platform/jobs/.*bionic_y1.*",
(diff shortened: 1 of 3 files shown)


tests/test_flow/test_persistence_aip.py
```diff
@@ -103,7 +103,6 @@ def x_plus_one(x):
 
     log_checker.expect_regex(
         r"Staging AI Platform task .* at gs://.*bionic_x_plus_one.*",
-        r"Submitting AI Platform task on .*x_plus_one.*x=1.*",
```
simonafk, line 106: Was this regex incorrect?

jqmp: Nope, it was correct, just mostly redundant. It was targeting the same message as the one on line 108, and the new implementation of `expect_regex` doesn't allow two regexes to match the same messag.

simonafk: Oh, good catch. It may have been a duplicate.

simonafk (approved):

simonafk:
Originally I wanted to keep the log regex checking simple, and verify that sufficient information is provided to the logger.

I think the changes here can help with debugging if the test breaks.

PR #312, opened 2020-12-15
Cache object hashes within CodeHasher

namanjain:
This change keeps track of all the object hashes for a CodeHasher
instance and avoids hashing the same object multiple times using it.
We scope the object hashes to a CodeHasher instance instead of globally
to avoid using a stale hash in case an object change by the time we
hash it again.

### DIFF
--- a/bionic/code_hasher.py
+++ b/bionic/code_hasher.py
@@ -61,90 +61,108 @@ class CodeHasher:
     """
 
     def __init__(self, suppress_warnings):
-        self._hash = hashlib.new("md5")
         self._suppress_warnings = suppress_warnings
         # This is used to detect circular references.
         self._object_depths_by_id = {}
+        # This caches the hashed values for objects we have already
+        # hashed.
+        self._obj_hash_digests_by_id = {}
+        # This stores the objects that we have hashed. If we don't
+        # store the objects, the objects can get garbage-collected,
+        # and new objects can reuse the same object id as the
+        # garbage-collected objects. Doing so ensures that the ids
+        # for the hashes we cache are not reused.
+        # See this stack overflow post for more details:
+        # https://stackoverflow.com/questions/35173479/why-do-different-methods-of-same-object-have-the-same-id
+        self._encountered_objects_by_id = {}
 
     @classmethod
     def hash(cls, obj, suppress_warnings=False):
         hasher = cls(suppress_warnings)
-        hasher._check_and_ingest(obj=obj)
-        return hasher._hash.hexdigest()
+        return hasher.hash_for_obj(obj)
 
-    def hexdigest(self):
-        return self._hash.hexdigest()
+    def hash_for_obj(self, obj):
+        return str(self._check_and_hash(obj))
 
-    def update(self, obj):
-        return self._check_and_ingest(obj)
-
-    def _ingest_raw_prefix_and_bytes(self, type_prefix, obj_bytes=b""):
-        self._hash.update(type_prefix.value)
-        self._hash.update(get_size_as_bytes(obj_bytes))
-        self._hash.update(PREFIX_SEPARATOR)
-        self._hash.update(obj_bytes)
-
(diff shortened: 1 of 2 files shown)


bionic/code_hasher.py
```diff
@@ -93,67 +82,75 @@ def _ingest_raw_prefix_and_bytes(self, type_prefix, obj_bytes=b""):
...
         """
-        Checks for circular references before calling the _ingest
+        Checks for circular references before calling the _update_hash
         method, which does the actual encoding.
         """
         obj_id = id(obj)
 
+        # If the hash is already cached, return the cached value.
+        if obj_id in self._hashes_by_id:
+            hash_obj = self._hashes_by_id[obj_id]
```
jqmp, line 94: I'm having trouble finding where we add entries to `self._hashes_by_id`.

bionic/code_hasher.py
```diff
@@ -93,67 +82,75 @@ def _ingest_raw_prefix_and_bytes(self, type_prefix, obj_bytes=b""):
         # The fix for this issue can also be applied to dedup code
         # references.
 
-    def _check_and_ingest(self, obj, code_context=None):
+    def _check_and_hash(self, obj, code_context=None):
         """
-        Checks for circular references before calling the _ingest
+        Checks for circular references before calling the _update_hash
```
jqmp, line 87: Can we update the docstrings to reflect what is being returned and passed around? It's hard for me to follow what state these hash objects are supposed to be in before and after each method.

bionic/code_hasher.py
```diff
@@ -62,28 +61,18 @@ class CodeHasher:
     """
 
     def __init__(self, suppress_warnings):
-        self._hash = hashlib.new("md5")
         self._suppress_warnings = suppress_warnings
         # This is used to detect circular references.
         self._object_depths_by_id = {}
+        self._hashes_by_id = {}
```
jqmp, line 67: It makes me a little nervous to be storing mutable hash objects in this cache. Would it make sense to store the final hash bytes instead?

bionic/code_hasher.py
namanjain: This is why we have code reviews!

bionic/code_hasher.py
```diff
@@ -291,9 +327,17 @@ class TypePrefix(Enum):
     INTERNAL_ROUTINE = b"AO"
     BUILTIN = b"AP"
     CLASS = b"AQ"
+    HASH = b"AR"
```
jqmp, line 330: I like the use of this new `HASH` category.

bionic/code_hasher.py
namanjain: I did that at first, but then I realized I was using `hash_obj.digest()` to add to other hash objects, and `hash_obj.hexdigest()` to return the final result. But I think I can just use `str(hash_obj.digest())` instead of `hexdigest` to return the final result instead.

bionic/code_hasher.py
```diff
@@ -93,67 +82,75 @@ def _ingest_raw_prefix_and_bytes(self, type_prefix, obj_bytes=b""):
...
-            self._ingest_raw_prefix_and_bytes(
-                type_prefix=TypePrefix.BYTEARRAY, obj_bytes=obj
-            )
+            add_to_hash(hash_obj, type_prefix=TypePrefix.BYTEARRAY, obj_bytes=obj)
 
         elif obj is None:
-            self._ingest_raw_prefix_and_bytes(
-                type_prefix=TypePrefix.NONE, obj_bytes=b"None"
-            )
+            add_to_hash(hash_obj, type_prefix=TypePrefix.NONE, obj_bytes=b"None")
```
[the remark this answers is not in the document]
namanjain: Not really. This is probably left around from the initial impl. I'll get rid of `b"None"`.

jqmp (approved):

PR #313, opened 2020-12-15
Remove TODOs

namanjain:
We use `func.__code__.co_code` to detect any changes to the function
code and hash of references to detect any changes to the code the
function uses. We don't need to hash the same reference multiple times
to detect changes to references. This change dedups the references to
avoid unneeded work for detecting changes.

### DIFF
--- a/bionic/code_hasher.py
+++ b/bionic/code_hasher.py
@@ -37,8 +37,6 @@ PREFIX_SEPARATOR = b"$"
 
 
 # List of things we should do before releasing Smart Caching:
-# - dedup references
-# - caching individual object hashes for a CodeHasher run
 # - hash classes
 # - Throw an exception for unhandled bytecode instruction type
 # - investigate if we can hash module or package versions


bionic/code_hasher.py
```diff
@@ -85,6 +85,14 @@ def _ingest_raw_prefix_and_bytes(self, type_prefix, obj_bytes=b""):
         self._hash.update(PREFIX_SEPARATOR)
         self._hash.update(obj_bytes)
 
+        # TODO: It is not safe to use just the id of an object since
```
jqmp, line 88: Hmmm, nice observation. What if we keep a reference to each object for the duration of the hashing process? That way their lifetimes will all overlap and they should each have a unique memory address.

jqmp (commented):
This will change our hash values and break cache compatibility when assist/auto versioning is enabled, right? Can/should we extend our compatibility tests to cover that?

namanjain:
> This will change our hash values and break cache compatibility when assist/auto versioning is enabled, right? Can/should we extend our compatibility tests to cover that?

Good call. I opened #315 for tests since I'm going to have a lot of tests and it would be nice to have the tests independent of this change.

bionic/code_hasher.py
namanjain: I'll try this out and report back if it works.

namanjain: Yup, this worked for code hasher. I already merged the change as part of #312.

tests/test_code_references.py
```diff
@@ -315,36 +310,26 @@ def predict(resize):
         plt.show,
         ReferenceProxy("int"),
         ReferenceProxy("len"),
+        # These are repeated since we are deduping using object ids.
```
namanjain, line 313: What if we dedup using hash when available for a reference? That might catch some of these duplicated references. Any reference that is not hashable can still be deduped using the id like what I'm doing here. What do you all think?

jqmp: Hmm... if the hasher is already maintaining an internal cache already-hashed objects, how important is it to remove duplicate references at all?

namanjain: Not a lot tbh. I don't mind just killing this PR (or merge this PR with only removing the TODO).

jqmp (approved):
Well, that made the review easier!

PR #315, opened 2020-12-17
Add compatibility tests for auto vesioning mode

namanjain:


jqmp (approved):
Looks pretty good, aside from the test failures.

I think in the past we've avoided the problems with wrappers by calling `counter.mark()` directly from the function instead wrapping it with `@counter`. (E.g., see `test_persistence.py::test_versioning`.) However, if you can figure out a way to transparently handle wrappers, that would be much better.

simonafk (approved):

tests/test_flow/test_persistence_compatibility.py
```diff
@@ -25,12 +26,18 @@ def older_serialized_cache_harness(make_counter, tmp_path):
 # To renegerate cache, run the following command from bionic/ dir
 #   `python -m tests.test_flow.generate_test_compatibility_cache`
 def test_caching_compatibility(older_serialized_cache_harness):
-    flow = older_serialized_cache_harness.flow
-    assert flow.get("xy") == 6
-    assert flow.get("yz") == 12
-    assert flow.get("xy_plus_yz") == 18
+    # Auto versioned flow requires Python 3.7 or higher.
+    if sys.version_info < (3, 7):
+        flows = [older_serialized_cache_harness.manual_flow]
```
simonafk, line 31: wondering, why not just test both manual and auto flows?

simonafk (approved):

tests/test_flow/generate_test_compatibility_cache.py
```diff
@@ -40,40 +40,48 @@ def __init__(self, cache_dir, make_counter):
...
+            xy_plus_yz_counter.mark()
             return xy + yz
 
-        self.flow = builder.build()
         self.xy_counter = xy_counter
         self.yz_counter = yz_counter
         self.xy_plus_yz_counter = xy_plus_yz_counter
 
+        self.manual_flow = builder.build()
+        builder.set("core__versioning_mode", "auto")
```
simonafk, line 66: Wondering, does the `auto` and `manual` modes make a difference to the tests here?

tests/test_flow/test_persistence_compatibility.py
namanjain: Mostly for the sake of parity. I tried parameterize the test based on the value of a fixture but unfortunately, pytest doesn't allow for dynamic parameterization.

tests/test_flow/generate_test_compatibility_cache.py
namanjain: A flow with `auto` mode will also add the bytecode of the function to its functional hash. Whereas manual mode wouldn't. Adding this test for auto mode will check for cache compatibility for auto versioning mode as well. Similar to `manual` mode cache compatibility check, the detection of bytecode hashing logic change is not exhaustive. But it should at least catch major changes to the hashing logic.

tests/test_flow/test_persistence_compatibility.py
```diff
@@ -25,12 +26,18 @@ def older_serialized_cache_harness(make_counter, tmp_path):
 # To renegerate cache, run the following command from bionic/ dir
 #   `python -m tests.test_flow.generate_test_compatibility_cache`
 def test_caching_compatibility(older_serialized_cache_harness):
-    flow = older_serialized_cache_harness.flow
-    assert flow.get("xy") == 6
-    assert flow.get("yz") == 12
-    assert flow.get("xy_plus_yz") == 18
+    # Auto versioned flow requires Python 3.7 or higher.
```
jqmp, line 29: Can we explain why 3.7+ is required? This makes it seem like auto versioning doesn't work in Python 3.6.

PR #316, opened 2020-12-17
Only get references inside modules for references with qualified names

namanjain:


tests/test_code_references.py
```diff
@@ -203,114 +208,123 @@ def x(my_class):
...
+        if op_type == "ceil":
+            func = math.ceil
+        else:
+            func = math.floor
+
+        return func(val)
+
+    # TODO: We should also return math.ceil as a reference. It isn't
+    # returned now because it is stored in a local variable. Maybe we
+    # should return the values of local variables as they are assigned.
```
jqmp, line 223: Nice catch. I'm still not sure why we don't use `set_tos` for `STORE_FAST` to track these references. Is there a drawback?

tests/test_code_references.py
```diff
@@ -203,114 +208,123 @@ def x(my_class):
...
-        hasher._ingest_raw_prefix_and_bytes,
-        TypePrefix.BYTES,
-        "isinstance",
-        "bytearray",
-        hasher._ingest_raw_prefix_and_bytes,
-        TypePrefix.BYTEARRAY,
-        hasher._ingest_raw_prefix_and_bytes,
-        TypePrefix.NONE,
-        "isinstance",
+    # Example code of a ml flow. Taken from below article with minor changes:
```
jqmp, line 228: I like the realism of this example, but one weakness it has is that there's no control flow. For example, it wouldn't surface the `math.ceil` issue you found above because it doesn't have any if statements -- right?

tests/test_code_references.py
namanjain: The only issue I can see is that we add the reference multiple times. But given that we are going to dedup the references anyway, I think we should add to `tos` for `STORE_FAST` too. I'll make the change in this PR.

tests/test_code_references.py
namanjain: Right. I will add a control flow to find the prediction model.

jqmp (approved):
LGTM!

PR #317, opened 2020-12-17
Hash proxy and string refs differently

namanjain:


bionic/code_references.py
```diff
@@ -252,11 +247,17 @@ def set_tos(t, placeholder=False):
     return refs
 
 
-class PlaceholderTos:
+class ReferenceProxy:
```
jqmp, line 250: Would it make sense to just use `@attr.s` for this?

bionic/code_references.py
```diff
@@ -252,11 +247,17 @@ def set_tos(t, placeholder=False):
...
     """
-    Wraps the placeholder strings for TOS objects to differentiate the
-    object from actual string objects.
+    Wraps the proxy of references to differentiate the proxy reference
```
jqmp, line 252: Maybe we could briefly explain what a reference proxy is and why it exists?

bionic/code_references.py
```diff
@@ -171,7 +166,7 @@ def set_tos(t, placeholder=False):
                     # This can happen if the variable does not exist, or if LOAD_NAME
                     # is trying to access a local frame argument. If we cannot find the
                     # variable, we return its name instead.
-                    set_tos(op.argval, placeholder=True)
+                    set_tos(op.argval, proxy=True)
```
jqmp, line 169: Could these just be written as `set_tos(ReferenceProxy(op.argval))`?

tests/test_code_references.py
```diff
@@ -213,104 +217,109 @@ def test_complex_function():
     hasher = CodeHasher(False)
     # TODO: This makes me realize that maybe we should dedup the results.
     assert get_references(hasher._ingest) == [
-        "isinstance",
-        "bytes",
+        ReferenceProxy("isinstance"),
```
jqmp, line 220: I really like the extra clarity here.

jqmp (approved):

PR #318, opened 2020-12-18
Introduce artifact descriptors

jqmp:
This PR refactors refactors our task structure to make saving and loading into separate tasks. There should be no behavior changes (except for some slight changes in logging).

There are three commits (see their messages for more details):

1. Moving a few methods around to make the next steps easier.
2. Defining a new type of descriptor.
3. Changing Bionic's internals to use the new descriptor.


jqmp:
There are two failing tests here. I think one of them (`test_parallel_fail`) may be a flake that we've seen elsewhere. The other one (`test_flow_handles_delete_gracefully[parallel]`) looks like a real bug. I think my changes have invalidated some of the assumptions in our persistent cache flushing logic in `_get_or_create_entry_for_state`, and this PR might require some more work.

bionic/core/flow_execution.py
```diff
@@ -146,12 +146,12 @@ def _process_active_entry(self, entry):
         if self._mark_entry_completed_if_possible(entry):
             return
 
-        # If this entry is persistable, we may be able to load it from the persistent
-        # cache, which would immediately get us to the CACHED level.
-        if entry.state.should_persist:
-            entry.state.attempt_to_access_persistent_cached_value()
+        # If this entry produces an artifact, we may be able to load it from the
+        # persistent cache, which would immediately get us to the CACHED level.
+        if entry.state.yields_artifact:
```
simonafk, line 151: What about `produces_artifact` to match comments?

bionic/core/task_execution.py
```diff
@@ -324,15 +314,15 @@ class EntryLevel(IntEnum):
...
 
-    3. PRIMED: If the TaskState's value is persistable, this is equivalent to CACHED;
+    3. PRIMED: If the TaskState's value is an artifact, this is equivalent to CACHED;
         otherwise it's equivalent to INITIALIZED. This abstract definition is useful
         because it guarantees two things:
 
         a.  The task's provenance digest is available, which means any downstream tasks
             can have their values loaded from the persisted cache. For a task with
-            non-persistable output, its provenance digest depends only on its
+            non-artifact output, its provenance digest depends only on its
```
simonafk, line 323: I think the term `non-persistable` is clearer than `non-artifact`.

bionic/core/flow_execution.py
simonafk: In `task_execution.py`, we talk about the difference in behavior between persistable and non-persistable output. To make things even more clearer, what about something along the lines of `produces_persistable_artifact` or `output_is_persistable` ?

PR #319, opened 2020-12-22
Hash class code in CodeHasher

namanjain:
With this change, Bionic now hashes classes with entity values as part
of content hash to invalidate cache. There are still some gaps with
this approach, like not hashing class properties. I'll add support for
class properties in a future PR, which will give me the opportunity to
get feedback on the approach as well.

### DIFF
--- a/bionic/code_hasher.py
+++ b/bionic/code_hasher.py
@@ -37,12 +37,12 @@ PREFIX_SEPARATOR = b"$"
 
 
 # List of things we should do before releasing Smart Caching:
-# - hash classes
 # - Throw an exception for unhandled bytecode instruction type
 # - investigate if we can hash module or package versions
 # - verify that we hash all Python constant types
 # - version.suppress_bytecode_warnings TODO
 # - skip and warn for referenced code objects
+# - add support for Enums, class properties, and attrs.Attribute
 
 
 class CodeHasher:
@@ -247,13 +247,32 @@ class CodeHasher:
             self._update_hash_for_code(hash_accumulator, obj, code_context)
 
         elif inspect.isclass(obj):
-            # TODO: Hashing classes is next on our roadmap. Let's hash
-            # the name for now.
-            add_to_hash(
-                hash_accumulator,
-                type_prefix=TypePrefix.CLASS,
-                obj_bytes=obj.__name__.encode(),
-            )
+            # TODO: See if we can get the version of the module and
+            # hash the version as well.
+            if is_internal_class(obj):
+                add_to_hash(hash_accumulator, type_prefix=TypePrefix.INTERNAL_CLASS)
+                class_name = "%s.%s" % (obj.__module__, obj.__name__)
+                add_to_hash(
+                    hash_accumulator,
+                    type_prefix=TypePrefix.HASH,
+                    obj_bytes=self._check_and_hash(class_name),
+                )
+            else:
+                add_to_hash(hash_accumulator, type_prefix=TypePrefix.CLASS)
+                members_to_hash = [
+                    m_value
+                    for (m_name, m_value) in inspect.getmembers(obj)
+                    if not (
+                        inspect.ismethoddescriptor(m_value)
+                        or inspect.isgetsetdescriptor(m_value)
+                        or m_name in {"__class__", "__dict__", "__members__"}
+                    )
+                ]
+                add_to_hash(
+                    hash_accumulator,
+                    type_prefix=TypePrefix.HASH,
+                    obj_bytes=self._check_and_hash(members_to_hash),
+                )
 
         else:
             # TODO: Verify that we hash all Python constant types.
@@ -344,6 +363,7 @@ class TypePrefix(Enum):
     CLASS = b"AQ"
     REF_PROXY = b"AR"
     HASH = b"AS"
+    INTERNAL_CLASS = b"AT"
     DEFAULT = b"ZZ"
 
 
@@ -357,3 +377,15 @@ def add_to_hash(hash_accumulator, type_prefix, obj_bytes=b""):
     hash_accumulator.update(get_size_as_bytes(obj_bytes))
     hash_accumulator.update(PREFIX_SEPARATOR)
     hash_accumulator.update(obj_bytes)
+
+
+def is_internal_class(class_obj):
+    assert inspect.isclass(class_obj)
+
+    try:
+        return is_internal_file(inspect.getfile(class_obj))
+    except TypeError:
+        # inspect.getfile throws TypeError when the class_obj is a
+        # built-in class. See the documentation for more info.
+        # https://docs.python.org/3/library/inspect.html#inspect.getfile
+        return False
--- a/bionic/code_hasher.py
+++ b/bionic/code_hasher.py
@@ -22,6 +22,7 @@ CACHE_SCHEMA_VERSION to update cache scheme.
 from enum import Enum
 import hashlib
 import inspect
+import types
 import warnings
 
 from .code_references import (
@@ -266,6 +267,12 @@ class CodeHasher:
                         inspect.ismethoddescriptor(m_value)
                         or inspect.isgetsetdescriptor(m_value)
                         or m_name in {"__class__", "__dict__", "__members__"}
+                        # TODO Remove this when hashing properties.
+                        or isinstance(m_value, property)
+                        # TODO These should be handled the same way as properties.
+                        or isinstance(m_value, types.DynamicClassAttribute)
+                        # TODO Remove this when hashing attr classes.
+                        or m_name == "__attrs_attrs__"
                     )
                 ]
                 add_to_hash(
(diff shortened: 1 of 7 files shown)


bionic/code_hasher.py
```diff
@@ -249,12 +274,25 @@ def _update_hash(self, hash_accumulator, obj, code_context):
...
-            # the name for now.
+            add_to_hash(hash_accumulator, type_prefix=TypePrefix.CLASS)
+            user_defined_members = []
+            for (m_name, m_value) in inspect.getmembers(obj):
+                is_user_defined = (
+                    # This still misses some types, like class properties.
+                    # There might be more, but we don't want to warn for them
+                    # since it is hard to differentiate between them and Python
+                    # class wrapper properties.
+                    type(m_value) in self.BUILTIN_TYPES
```
namanjain, line 285: Note to self, hash Enum variables as well

docs/release-notes.rst
```diff
@@ -83,6 +83,9 @@ Improvements
 - When "assisted" or "fully automatic" :ref:`versioning <automatic-versioning>`
   is enabled, Bionic now inspects not just entity functions, but also any
   functions they reference.
+- When "assisted" or "fully automatic" :ref:`versioning <automatic-versioning>`
```
jqmp, line 86: Could we just merge this with the bullet above?

bionic/code_hasher.py
```diff
@@ -249,12 +274,25 @@ def _update_hash(self, hash_accumulator, obj, code_context):
             self._update_hash_for_code(hash_accumulator, obj, code_context)
 
         elif inspect.isclass(obj):
-            # TODO: Hashing classes is next on our roadmap. Let's hash
-            # the name for now.
+            add_to_hash(hash_accumulator, type_prefix=TypePrefix.CLASS)
```
jqmp, line 277: Will we do this for classes defined in libraries, like `pandas.DataFrame`? Should we?

bionic/core/task_execution.py
```diff
@@ -751,13 +752,35 @@ def _local_artifact_from_value(self, value):
...
+        protocol = self.desc_metadata.protocol
+        code_versioning_policy = self.func_attrs.code_versioning_policy
+        code_version = code_versioning_policy.version
+        versioning_policy = context.core.versioning_policy
+        try:
+            bytecode_hash = CodeHasher.hash(
+                type(value),
+                code_versioning_policy.suppress_bytecode_warnings
+                or versioning_policy.ignore_bytecode_exceptions
+                or not code_version.includes_bytecode,
```
jqmp, line 772: Hmm... if `not code_version.includes_bytecode`, should we be doing this hash at all?

bionic/core/task_execution.py
```diff
@@ -751,13 +752,35 @@ def _local_artifact_from_value(self, value):
...
 
-        value_hash = protocol.tokenize_file(value_path)
+        value_hash = self._generate_value_hash(value, value_path, context)
         return Artifact(
             url=url_from_path(value_path),
             content_hash=value_hash,
         )
         return value_path
 
+    def _generate_value_hash(self, value, value_path, context):
```
jqmp, line 762: Should we do this for all values, or just pickled ones? For example, if we serialize a dataframe object to a Parquet file, it will always get deserialized as a standard Pandas dataframe, so it's not clear that the details of the original class matter.

bionic/core/task_execution.py
```diff
@@ -751,13 +752,35 @@ def _local_artifact_from_value(self, value):
...
+            )
+        except Exception:
+            bytecode_hash = None
+            if (
+                not versioning_policy.ignore_bytecode_exceptions
+                and code_version.includes_bytecode
+            ):
+                raise
+
+        return protocol.tokenize_file(value_path) + protocol.tokenize(bytecode_hash)
```
jqmp, line 782: `protocol.tokenize` is really only intended to be applied to valid values for that protocol. Since `bytecode_hash` is already a hash, could we just append it directly to the tokenized file? (Or make this hashing logic the responsibility of the protocol?)

jqmp (commented):
I have some questions about the details, but the idea and tests look good!

bionic/code_hasher.py
```diff
@@ -249,12 +274,25 @@ def _update_hash(self, hash_accumulator, obj, code_context):
...
+            user_defined_members = []
+            for (m_name, m_value) in inspect.getmembers(obj):
```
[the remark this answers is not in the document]
namanjain: So I started with this approach in the design doc/one pager so I thought people must have at least seen the approach. But after opening the PR, I kinda reached the same conclusion as you and posted this in our dev slack channel thread. I'll give this one a shot and see if it's doable.

bionic/code_hasher.py
namanjain: I don't think it's too bad to hash them if we are whitelisting hashable properties. Any of the more complex types like methods will be filtered out in the `inspect.isroutine` condition.

But since we are now considering blacklisting non-hashable properties instead, it might be worth skipping classes like we do routines. I'll try it when I'm addressing the [other comment](https://github.com/square/bionic/pull/319#discussion_r549373846).

bionic/core/task_execution.py
namanjain: I'm not 100% sure what the right thing to do here is either. Let's discuss this when we come back to work.

bionic/core/task_execution.py
namanjain: I like this idea!

bionic/core/task_execution.py
namanjain: Taking your [previous comment](https://github.com/square/bionic/pull/319#discussion_r549390370) into consideration, I think it makes sense to make this the responsibility of the protocol.

namanjain: I moved the logic into the protocol and appended it to the tokenized file.

bionic/core/task_execution.py
namanjain: I'm skipping it for now, but open to discussion on this.

bionic/code_hasher.py
namanjain: Done!

bionic/core/task_execution.py
```diff
@@ -751,13 +751,34 @@ def _local_artifact_from_value(self, value):
                 )
             ) from e
 
-        value_hash = protocol.tokenize_file(value_path)
+        value_hash = protocol.tokenize_file(
```
jqmp, line 754: nit: Aesthetically, it feels to me like hash strings should have a fixed, moderate length, so I'd be tempted to create a new hash from these two values instead of concatenating them. (Although there's no functional benefit except for making the metadata YAML a tiny bit cleaner-looking.) WDYT?

bionic/protocols.py
```diff
@@ -156,6 +160,13 @@ def tokenize_file(self, path):
         """Like ``tokenize``, but operates on a serialized value."""
         return tokenization.tokenize(path, read_hashable_bytes_from_file_or_dir)
 
+    def hash_value_type(self, value_type, suppress_warnings):
+        """
+        Converts the type of a value into a unique hash which can be used to
+        invalidate cache.
```
jqmp, line 166: nit: "invalidate *the* cache"

But also, I don't think this provides enough context on what's happening here. The key thing seems to me that this gives the protocol an opportunity to add data to the hash which doesn't show up in the serialized output.

Finally, if the framing above makes sense, WDYT about just making this something like just `get_extra_value_hash(self, value, suppress_warnings)` that takes the value itself?

docs/concepts.rst
```diff
@@ -333,10 +333,7 @@ won’t detect changes of the following types:
    global variables that have simple types (int, string, bool, bytes, etc.), but
    it will ignore changes for other types of variables.
 
-3. **Python Classes**: Bionic won’t detect changes to the code of Python classes.
```
jqmp, line 336: Hmmm. Offline I agreed that we don't need to generate warnings for things we don't handle yet, like `attrs` attributes. But if we're going to say in the document that we hash classes, then it does seem appropriate to emit warnings for stuff we don't support yet. Maybe we should say something like "we detect some but not all changes" here.

docs/release-notes.rst
```diff
@@ -81,8 +81,8 @@ Improvements
   <bionic.outputs>` decorator), those entities and the function itself are now
   all visualized with the same color.
 - When "assisted" or "fully automatic" :ref:`versioning <automatic-versioning>`
-  is enabled, Bionic now inspects not just entity functions, but also any
-  functions they reference.
+  is enabled, Bionic now inspects not just entity functions and values, but also
+  any functions they reference and the class type of the entity values.
```
jqmp, line 85: nit: "the classes of entity values".

jqmp (approved):

bionic/core/task_execution.py
namanjain: Yeah, that works for me. I'll hash them using `hash_simple_obj_to_hex`.

PR #320, opened 2021-01-07
Fix artifact deletion bug

jqmp:
We had a bug where Bionic could crash (throw an exception) if we ran in
parallel mode after an artifact was manually deleted. This commit fixes
the bug and adds a couple of tests for this situation.

The problem was this: our system is supposed to detect artifact deletion
by clearing each TaskState's cache state when we create a new
TaskRunnerEntry for it. However, when we execute a task remotely, we
send over a whole subgraph that may include tasks that we never created
entries for, so we may have stale data for those tasks and think they
still have artifacts when they don't. When we try to actually run the
subgraph remotely, we fail to load those artifacts and then try to
re-run those tasks, but since we stripped off the Task object, that
fails.

To fix this, I updated the logic to explicitly identify and check the
task entries that need to be cached. This mostly involved refactoring
the subgraph code: instead of looking at which TaskStates are already
cahced (which could be stale), we identify the states that logically
need to be cached, and then let the runner code figure out if they
actually are.

I also made two other small changes:

1. We now handle the case where some of the outputs ("target states") of
a subgraph are already completed. This can happen if one function
generates multiple artifacts, and then just one of them is manually
deleted.

2. I tweaked the definition of "cached". Previously it was possible for
a persistable task to be "cached" if it had a memoized result but not
persisted artifact; this could only happen if the artifact was deleted,
and this case breaks the subgraph assumptions above. Now persistable
tasks are only "cached" if they have an artifact.

### DIFF
--- a/bionic/core/flow_execution.py
+++ b/bionic/core/flow_execution.py
@@ -168,7 +168,7 @@ class TaskCompletionRunner:
             self._aip_execution_enabled or self._parallel_execution_enabled
         ) and entry.state.should_persist
         if entry_may_be_computable_remotely:
-            remote_subgraph = RemoteSubgraph(entry.state, self._context)
+            remote_subgraph = RemoteSubgraph(entry.state)
             if self._aip_execution_enabled:
                 aip_task_configs = remote_subgraph.distinct_aip_task_configs
             else:
@@ -180,7 +180,7 @@ class TaskCompletionRunner:
                 descriptions_str = "; ".join(
                     f"function outputting {state.task_key.dnode.to_descriptor()!r} "
                     f"requires {state.func_attrs.aip_task_config}"
-                    for state in remote_subgraph.stripped_states_with_aip_task_configs
+                    for state in remote_subgraph.states_with_aip_task_configs
                 )
                 message = f"""
                 Multiple functions need to be run together (since some are not
@@ -196,11 +196,11 @@ class TaskCompletionRunner:
                     # as an AIP-decorated derived node.
                     non_serializable_descriptors = [
                         state.task_key.dnode.to_descriptor()
-                        for state in remote_subgraph.non_serializable_stripped_states
+                        for state in remote_subgraph.non_serializable_states
                     ]
                     aip_descriptors = [
                         state.task_key.dnode.to_descriptor()
-                        for state in remote_subgraph.stripped_states_with_aip_task_configs
+                        for state in remote_subgraph.states_with_aip_task_configs
                     ]
                     message = f"""
                     Found impossible configuration:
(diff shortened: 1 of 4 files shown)


bionic/core/flow_execution.py
```diff
@@ -224,23 +224,41 @@ def _process_active_entry(self, entry):
...
+                for external_dep_state in remote_subgraph.external_dependency_states
+            ]
+            assert entry not in dep_entries
+
+            # All of our dependency entries will need to be cached before we can run
+            # this subgraph. (One might think that they should already be cached because
+            # they are persistable and they are ancestors of the active entry, so we
+            # should have cached all of them by now. However, it's possible that some
+            # of their artifacts have been deleted since then, so we need to make sure
+            # to actuall instantiate their entries and confirm that they're still
```
simonafk, line 249: nit: typo: actually

bionic/core/task_execution.py
```diff
@@ -782,97 +785,126 @@ class RemoteSubgraph:
...
 
-            if stripped_state.is_cached_persistently:
+            # Strip out data cached in memory -- we can't necessarily pickle it, so
+            # we need to get rid of it before trying to transmit this state to
+            # another process.
+            stripped_state._result = None
+
+            # These fields are picklable, but only needed for cache setup and
+            # initialization.
+            if stripped_state.is_initialized:
```
simonafk, line 875: Wondering, what scenario would the state be not serialized?

namanjain (approved):
The change looks good overall and I think it should fix the problem.

But reviewing this PR definitely made me realize how complicated this code has become. I think the ideas we discussed before off-GitHub around changing the way persistence works and maybe having a more short lived cache accessor would help simplify this logic. What do you think about leaving a TODO for it?

bionic/core/flow_execution.py
```diff
@@ -279,12 +297,15 @@ def _process_active_entry(self, entry):
...
+                if target_entry.stage == EntryStage.ACTIVE:
+                    active_target_entries.append(target_entry)
+                else:
+                    assert target_entry.stage == EntryStage.COMPLETED
 
-            stripped_target_states = [
-                remote_subgraph.get_stripped_state(target_entry.state)
-                for target_entry in target_entries
-            ]
+            stripped_active_target_states = remote_subgraph.strip_states(
```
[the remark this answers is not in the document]
jqmp: Yeah, that sounds like a good idea to me. (I would like to find a name that makes it sound less like we're mutating the states, but that's orthogonal -- I agree "prepare" is a better verb than "strip".)

bionic/core/task_execution.py
jqmp: I'm assuming you meant "in what scenario would the state not be initialized?" That's a great question and I had to poke around for a while to answer it. At first I thought this was impossible and this `if` statement was obsolete, because of this comment in `flow_execution`:

```
# We want every task state to be initialized before we try running it
# remotely. We'll just require this of the target states, since all the
# other ones are ancestors of them.
```

However, it turns out this comment is wrong (which I found out by adding `assert stripped_state.is_initialized` to the line you're asking about). In fact, we can have non-persistable followup tasks that are not initialized yet. For example:

```
@builder
@bn.persist(False)
@bn.outputs("a", "b")
def _():
    return 1, 2

@builder
def c(a):
    return a + 1

builder.build().get("c")
```

In this case, `b` is a follow-up of `a, b`, so it will appear in the subgraph; but it is not persistable, so it won't appear in `target_states` and won't get initialized.

I will fix that comment -- thanks!

(Also, as @namanjain pointed out, we probably don't need the `case_key` attribute at all more, so this line of code is obsolete for other reasons. I'm also not 100% sure we still need `every task state to be initialized before we try running it remotely`, but it doesn't hurt to do it, so I'll just add a `TODO` to come back and check if it's still necessary.)

jqmp:
> But reviewing this PR definitely made me realize how complicated this code has become. I think the ideas we discussed before off-GitHub around changing the way persistence works and maybe having a more short lived cache accessor would help simplify this logic. What do you think about leaving a TODO for it?

Yes, I agree. I'll add that TODO. I started a branch to do just that, but put it on hold when I noticed this pre-existing bug. I'm actually wondering if we can get rid of the stateful cache accessor altogether once we move to a short-lived artifact cache.

PR #321, opened 2021-01-12
Don't add extra value hash for manual and assist versioning mode

namanjain:
The previous change introduced a bug where we hashed classes for the
values even for manual and assisted versioning mode. That seemed to
have been caused from a git conflict, but this change fixes that bug.

### DIFF
--- a/tests/test_flow/test_persistence.py
+++ b/tests/test_flow/test_persistence.py
@@ -902,6 +902,73 @@ def test_versioning_auto_class_changes(builder, make_counter):
     assert call_counter.times_called() == 1
 
 
+@pytest.mark.parametrize("versioning_mode", ["manual", "assist"])
+def test_versioning_class_changes(builder, make_counter, versioning_mode):
+    call_counter = make_counter()
+
+    # PicklableProtocol cannot pickle IntWrapper class because it is
+    # defined inside the test function. We instead use this custom
+    # protocol that wraps the PicklableProtocol to be able to serialize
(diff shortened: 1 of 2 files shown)


tests/test_flow/test_persistence.py
```diff
@@ -902,6 +902,69 @@ def wrapped_f(x, y):  # noqa: F811
     assert call_counter.times_called() == 1
 
 
+def test_versioning_manual_class_changes(builder, make_counter):
```
namanjain, line 905: Note: I ran this test on master to verify the bug and the fix. It fails on master whereas it passes with the changes in this PR.

jqmp: Should we test the assisted mode as well?

jqmp (approved):

tests/test_flow/test_persistence.py
```diff
@@ -902,6 +902,73 @@ def wrapped_f(x, y):  # noqa: F811
...
+    class IntWrapper:  # noqa: F811
+        def __init__(self, value):
+            self.value = value
+
+        def __str__(self):
+            return f"value={self.value}"
+
+    @builder  # noqa: F811
+    @IntWrapperProtocol()
+    @bn.version(2)
```
jqmp, line 960: Suppose we used the same version here (`@version(1)`), and we had `versioning_mode == "assist"`. In that case we'd want this to throw an exception, right? Will that happen?

PR #323, opened 2021-01-21
aip docker build

simonafk:
Add a mechanism for building docker images automatically from the
current Python environment.

### DIFF
--- a/bionic/aip/docker_image_builder.py
+++ b/bionic/aip/docker_image_builder.py
@@ -1,0 +1,144 @@
+"""
+Builds a docker image for Google AI Platform execution using the current Python
+environment.
+"""
+import pathlib
+import subprocess
+from concurrent.futures import Future
+from concurrent.futures.thread import ThreadPoolExecutor
+from textwrap import dedent
+
+from bionic.deps.optdep import import_optional_dependency
+
+import hashlib
+import sys
+import tempfile
+import re
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+_cached_docker_module = None
+_cached_docker_client = None
+
+
+def get_docker_module():
+    global _cached_docker_module
+
+    if _cached_docker_module is None:
+        _cached_docker_module = import_optional_dependency(
+            "docker", purpose="Build Docker images"
+        )
+
+    return _cached_docker_module
+
+
+def get_docker_client():
+    global _cached_docker_client
+
+    if _cached_docker_client is None:
+        docker = get_docker_module()
+        logger.info("Initializing Docker client ...")
+        _cached_docker_client = docker.from_env()
+
+    return _cached_docker_client
+
+
+def fix_pip_requirements(pip_requirements: str) -> str:
+    # Pip freeze may contain entries with editable installs pointing to remote
+    # git repositories. This can happen when doing Bionic development. Docker
+    # service is not able to access repositories using the git+git protocol.
+    # Hence, any entries containing git+git is converted to use git+https.
+    #
+    # Example entry:
+    # -e git+git@github.com:square/bionic.git@f13f5405e928d92b553d2cbee41084eecccf7de3#egg=bionic
+    #
+    # Converted entry:
+    # -e git+https://github.com/square/bionic.git@f13f5405e928d92b553d2cbee41084eecccf7de3#egg=bionic
+
+    def fix_line(line: str) -> str:
+        if line.startswith("-e git+git"):
+            return re.sub(r"-e ([^@]*)@", "-e git+https://", line.replace(":", "/"))
+        else:
+            return line
+
+    return "\n".join([fix_line(line) for line in pip_requirements.split("\n")])
+
+
+def get_pip_freeze() -> str:
+    return subprocess.run(
+        ["pip", "freeze"], capture_output=True, check=True, encoding="utf-8"
+    ).stdout
+
+
+def get_pip_requirements() -> str:
+    return fix_pip_requirements(get_pip_freeze())
+
+
+def get_image_uri(project_name: str, pip_requirements: str) -> str:
+    m = hashlib.sha256()
+    m.update(pip_requirements.encode("utf-8"))
+    m.update(str(sys.version_info).encode("utf-8"))
+
+    image_tag = f"bionic_{m.hexdigest()}"
+
+    return f"gcr.io/{project_name}/bionic:{image_tag}"
+
+
+def build_image(
+    docker_client,
+    pip_requirements: str,
+    image_uri: str,
+):
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        tmp_path = pathlib.Path(tmp_dir)
+
+        (tmp_path / "requirements.txt").write_text(pip_requirements)
+
+        container_image = f"python:{sys.version_info[0]}.{sys.version_info[1]}"
+
+        (tmp_path / "Dockerfile").write_text(
+            dedent(
+                f"""
+                    FROM {container_image}
+                    COPY requirements.txt requirements.txt
+                    RUN pip install -r requirements.txt
+                """
+            )
+        )
+
+        logger.info(f"Building {image_uri} using {container_image}")
+        image, _ = docker_client.images.build(path=tmp_dir, tag=f"{image_uri}")
+
+        logger.info(f"Pushing {image_uri}")
+        for line in docker_client.images.push(f"{image_uri}", stream=True, decode=True):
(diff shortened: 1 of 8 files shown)


bionic/aip/docker_image_builder.py
```diff
@@ -0,0 +1,129 @@
...
+        docker_client.images.build(path=tmp_dir, tag=f"{image_uri}")
+
+        logger.info(f"Pushing {image_uri}")
+        for line in docker_client.images.push(f"{image_uri}", stream=True, decode=True):
+            logger.debug(line)
+
+        logger.info(f"Uploaded {image_uri}")
+
+
+def build_image_if_not_exists(project_name) -> str:
```
jqmp, line 98: nit: `if_not_exists` doesn't sound grammatical. WDYT about `retrieve_or_build_image_uri`?

bionic/executor.py
```diff
@@ -30,18 +32,39 @@ class AipExecutor:
 
     NON_ALPHANUMERIC_PATTERN = re.compile(r"\W")
 
-    def __init__(self, gcs_fs, aip_client, aip_config):
+    def __init__(
+        self,
+        gcs_fs,
+        aip_client,
+        aip_config: Config,
+        docker_build_enabled: bool,
```
jqmp, line 40: WDYT about replacing these two arguments with a single `docker_image_uri_func` argument, which is a function that returns the URI?

bionic/flow.py
```diff
@@ -1900,6 +1900,7 @@ def core__process_executor(
 
     builder.assign("core__aip_execution__enabled", False, persist=False)
     builder.assign("core__aip_execution__gcp_project_name", None, persist=False)
+    builder.assign("core__aip_execution__docker_build_enabled", False, persist=False)
```
jqmp, line 1903: WDYT about removing this entity and making the build process happen automatically if no image name is provided?

tests/test_flow/test_persistence_aip.py
```diff
@@ -107,3 +107,22 @@ def x_plus_one(x):
         r"Submitting AI Platform task .*\(name='x_plus_one'\).*CaseKey\(x=1\).*",
         r".*error while doing remote computation for x_plus_one\(x=1\).*AipError.*",
     )
+
+
+# This test may not work if the current Bionic git branch is not pushed to the
```
jqmp, line 112: Can we validate this somehow so that the test doesn't create misleading failures?

tests/test_flow/test_persistence_aip.py
```diff
@@ -107,3 +107,22 @@ def x_plus_one(x):
...
+@pytest.mark.no_parallel
+def test_aip_with_docker_build(aip_builder):
+    builder = aip_builder
+    builder.set("core__aip_execution__docker_build_enabled", True)
+    builder.set("core__aip_execution__docker_image_name", None)
+
+    @builder
+    @bn.aip_task_config("n1-standard-4")
+    def x():
+        return 1
```
jqmp, line 124: Is there anything we can do here to validate that the environment we're running in matches the local one?

simonafk: Good suggestion. Maybe we can run a `pip freeze` and verify that the output is the same as the local environment.

bionic/flow.py
simonafk: Good question. I was initially thinking that if `core__aip_execution__docker_image_uri` is specified and `core__aip_execution__docker_build_enabled` is true, then maybe we can build a docker image with the tag specified in `core__aip_execution__docker_image_uri`.

bionic/aip/docker_image_builder.py
simonafk: Hmm, what about `build_image_if_missing` ?

Bionic does not retrieve the image, but only verify if it exists.

bionic/aip/docker_image_builder.py
```diff
@@ -0,0 +1,129 @@
...
+    if _cached_docker_client is None:
+        docker = get_docker_module()
+        logger.info("Initializing Docker client ...")
+        _cached_docker_client = docker.from_env()
+
+    return _cached_docker_client
+
+
+def fix_pip_requirements(pip_requirements) -> str:
+    # There may be entries with editable installs that point to git
```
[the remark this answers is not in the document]
simonafk: +1 adding a test

tests/test_flow/test_persistence_aip.py
simonafk: This entity can return the output of `pip freeze`. Pip entry for bionic is excluded because it does not have access to github.

tests/test_flow/test_persistence_aip.py
simonafk: Added a `@pytest.mark.skipif` check to see if the current git branch is committed to the remote repo.

I am making an assumption that this test is run only when doing bionic development.

bionic/flow.py
```diff
@@ -1986,13 +1978,24 @@ def core__aip_executor(
                 core__persistent_cache__gcs__fs is None.
             """
             raise AssertionError(oneline(error_message))
+
+        if core__aip_execution__docker_image_uri is None:
+            docker_image_uri_func = build_image_if_missing_async(
```
jqmp, line 1983: Is it possible for `build_image_if_missing_async` to hang, or just take a long time? If so, it might be confusing to the user, since it won't be obvious from the log messages that Bionic is waiting for this thread to finish. WDYT about wrapping this function like this?

    docker_image_uri_future = build_image_if_missing_async( core__aip_execution__config.project_name)
    def docker_image_uri_func():
        if not docker_image_uri_future.done():
            logger.info("Waiting for Docker image to finish building...")
        return docker_image_uri_future.result

tests/test_flow/test_persistence_aip.py
```diff
@@ -107,3 +115,67 @@ def x_plus_one(x):
...
+
+def git_repo_is_clean_and_pushed():
+    repo = Repo(os.getcwd(), search_parent_directories=True)
+    return (
+        not repo.is_dirty()
+        and len(repo.git.branch("-r", "--contains", repo.head.ref.object.hexsha)) > 0
+    )
+
+
+@pytest.mark.skipif(
```
jqmp, line 150: How expensive is this check?

tests/test_flow/test_persistence_aip.py
```diff
@@ -107,3 +115,67 @@ def x_plus_one(x):
...
+    not git_repo_is_clean_and_pushed(),
+    reason="current commit needs to be available in remote repo for docker access",
+)
+@pytest.mark.real_gcp_only
+@pytest.mark.no_parallel
+def test_aip_with_docker_build(aip_builder, gcp_project):
```
jqmp, line 156: Is there any way to test just the image building part, or to test it together with a fake AIP? (Not necessarily in this PR -- just curious how hard it would be.)

jqmp (approved):
I have a few questions still, but overall LGTM!

tests/test_flow/test_persistence_aip.py
simonafk: This checks if there are uncommitted changes and if the current commit hash is in the remote repo. Both operations should be pretty fast; the latter is just a hash check.

tests/test_flow/test_persistence_aip.py
simonafk: I think it should be possible to do it on github. There are github actions that can install docker on the test environment. My concern is that it would take some amount of time downloading and installing all the python packages and consuming github quota.

bionic/aip/docker_image_builder.py
```diff
@@ -0,0 +1,144 @@
...
+    # Hence, any entries containing git+git is converted to use git+https.
+    #
+    # Example entry:
+    # -e git+git@github.com:square/bionic.git@f13f5405e928d92b553d2cbee41084eecccf7de3#egg=bionic
+    #
+    # Converted entry:
+    # -e git+https://github.com/square/bionic.git@f13f5405e928d92b553d2cbee41084eecccf7de3#egg=bionic
+
+    def fix_line(line: str) -> str:
+        if line.startswith("-e git+git"):
```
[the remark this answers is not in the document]
simonafk: In addition to `git`, `pip freeze` lines can also contain other VCS like Mercurial. I think it would be complicated and require a lot of effort to handle every possible combination. Only `git` is handled for now since Bionic is hosted in github.

Maybe we can add a warning whenever a line using VCS is detected?

PR #325, opened 2021-02-03
Warn for unknown bytecode instructions

namanjain:


bionic/code_references.py
```diff
@@ -246,3 +255,133 @@ class ReferenceProxy:
     """
 
     val = attr.ib()
+
+
+"""
+List of all the opcode instructions that we have evaluted for finding code
+references. As of now, these are all the instructions listed in dis upto
+Python 3.8.
```
simonafk, line 263: Should we also support up to 3.9?

simonafk:
Looking at the test failures for 3.6 and 3.7
```
AssertionError: assert 'BREAK_LOOP' in {'BEFORE_ASYNC_WITH', 'BEGIN_FINALLY', 'BINARY_ADD', 'BINARY_AND', 'BINARY_FLOOR_DIVIDE', 'BINARY_LSHIFT', ...}
```

Test failure on 3.8 seems to be due to flaky test.

bionic/code_references.py
namanjain: IIRC, Bionic uses a few libraries that don't support 3.9 yet. I think we are dependent on them to either formally support it, or try and see if our usage of the libraries are supported with 3.9.

namanjain:
> Looking at the test failures for 3.6 and 3.7
> 
> ```
> AssertionError: assert 'BREAK_LOOP' in {'BEFORE_ASYNC_WITH', 'BEGIN_FINALLY', 'BINARY_ADD', 'BINARY_AND', 'BINARY_FLOOR_DIVIDE', 'BINARY_LSHIFT', ...}
> ```
> 
> Test failure on 3.8 seems to be due to flaky test.

Fixed the tests. Thanks!

bionic/code_references.py
```diff
@@ -151,6 +151,17 @@ def set_tos(t):
 
     for op in dis.get_instructions(code):
         try:
+            if op.opname not in SUPPORTED_INSTRUCTIONS:
+                message = f"""
+                Bytecode operation {op.opname} not found in the list of Bionic
+                supported bytecode operations. You must be using an unsupported
```
simonafk, line 157: How does `may` sound vs `must`?

namanjain: Done

bionic/code_references.py
```diff
@@ -151,6 +151,17 @@ def set_tos(t):
 
     for op in dis.get_instructions(code):
         try:
+            if op.opname not in SUPPORTED_INSTRUCTIONS:
+                message = f"""
+                Bytecode operation {op.opname} not found in the list of Bionic
```
jqmp, line 156: I don't think the "list of supported bytecode instructions" is a concept that users should have to think about. Either (a) they are using a newer version of Python that we don't fully support, or (b) we screwed up and missed an instruction. Can we look at the Python version, figure out which case it is, and then give the appropriate message?

bionic/code_references.py
jqmp: It looks like the only hard dependency that doesn't explicitly support 3.9 is `pyrsistent`, and I suspect it's unlikely to break in 3.9.

bionic/code_references.py
```diff
@@ -246,3 +255,138 @@ class ReferenceProxy:
     """
 
     val = attr.ib()
+
+
+"""
+List of all the opcode instructions that we have evaluted for finding code
```
jqmp, line 261: s/evaluted/evaluated/
s/upto/up to/

jqmp (approved):

bionic/code_references.py
namanjain: I'll leave a different TODO for supporting Python 3.9 in case that brings up some unknown issues.

PR #326, opened 2021-02-04
rename aip decorator

simonafk:
rename from aip_task_config to run_in_aip


namanjain (approved):
I don't think anyone is using this decorator yet and the rename is sensible.

PR #327, opened 2021-02-05
Remove TODO to hash module version

namanjain:
There is no sure way to get version for kinds of module in Python. You
can find versions for top level modules from `__version__` member
filed, but that version field doesn't exist for all modules. Since we
can't hash module versions consistently, this change does not hash the
version to be consistent and to avoid confusing behavior for the user.

### DIFF
--- a/bionic/code_hasher.py
+++ b/bionic/code_hasher.py
@@ -39,7 +39,6 @@ PREFIX_SEPARATOR = b"$"
 
 # List of things we should do before releasing Smart Caching:
 # - Throw an exception for unhandled bytecode instruction type
-# - investigate if we can hash module or package versions
 # - verify that we hash all Python constant types
 # - version.suppress_bytecode_warnings TODO
 # - skip and warn for referenced code objects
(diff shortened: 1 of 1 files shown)


jqmp (approved):
Can we add a comment at the module-handling section explaining why we made this choice?

I'm not opposed to this choice, but it's not obvious to me that it would be confusing to the user to use the version when specified. Is the problem that we might have a sub-module (like `from bionic import core`) and we don't want to have to import the parent module to check if it has a version?

namanjain:
> Can we add a comment at the module-handling section explaining why we made this choice?
> 
> I'm not opposed to this choice, but it's not obvious to me that it would be confusing to the user to use the version when specified. Is the problem that we might have a sub-module (like `from bionic import core`) and we don't want to have to import the parent module to check if it has a version?

Right. If we have a sub-module like you specified, I don't think we should import the parent module to get the version (and I don't think there is even an easy way to do it anyway). IMO, users won't think about what is the top level module and what is not and it would be confusing to see Bionic invalidate cache for some imports and not invalidate it for others.

I can add a comment explaining all of this.

PR #328, opened 2021-02-05
Skip code object references for hashing

namanjain:
We find references for a function from it's code object using the code
context created from the function. These references won't necessarily
share the same context with the function.

This change stops sending the code context to hash the references. But
without a code context, we cannot hash a code object efficiently (since
we can't find it's references). For any references that are code
objects, this change also treats them as a complex type and emits a
warning instead.

### DIFF
--- a/bionic/code_hasher.py
+++ b/bionic/code_hasher.py
@@ -39,7 +39,6 @@ PREFIX_SEPARATOR = b"$"
 
 # List of things we should do before releasing Smart Caching:
 # - verify that we hash all Python constant types
-# - skip and warn for referenced code objects
 # - add support for class properties and attr.Attribute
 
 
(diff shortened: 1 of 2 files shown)


bionic/code_hasher.py
```diff
@@ -340,15 +318,48 @@ def _update_hash_for_code(self, hash_accumulator, code, code_context):
         )
 
         references = get_referenced_objects(code, code_context, self._suppress_warnings)
-        # TODO: We should skip any referenced code objects since
-        # they can't be analyzed with the current code's
-        # code_context.
+        references = self._filter_referenced_code_objects(hash_accumulator, references)
```
jqmp, line 321: Maybe we should have a comment explaining why we need to do this?

jqmp (approved):

PR #329, opened 2021-02-05
Hash class properties in CodeHasher

namanjain:


tests/test_code_hasher.py
```diff
@@ -192,6 +192,30 @@ class ClassWithClassMethod2:
         def v(cls):
             return 20
 
+    class ClassWithProperty:
```
jqmp, line 195: Should these two classes get added to `values`?

bionic/code_hasher.py
```diff
@@ -219,6 +219,19 @@ def _update_hash(self, hash_accumulator, obj, code_context):
...
+        elif isinstance(obj, (property, types.DynamicClassAttribute)):
+            if isinstance(obj, list):
+                type_prefix = TypePrefix.PROPERTY
+            else:
+                type_prefix = TypePrefix.DYNAMIC_CLASS_ATTR
+            # A property is identified using fget, fset, fdel, and doc.
+            members = [obj.fget, obj.fset, obj.fdel, obj.__doc__]
+            add_to_hash(
+                hash_accumulator,
+                type_prefix=TypePrefix.HASH,
```
jqmp, line 231: Should this be `type_prefix=type_prefix`?

bionic/code_hasher.py
```diff
@@ -219,6 +219,19 @@ def _update_hash(self, hash_accumulator, obj, code_context):
                 obj_bytes=self._check_and_hash(obj.value, code_context),
             )
 
+        elif isinstance(obj, (property, types.DynamicClassAttribute)):
+            if isinstance(obj, list):
+                type_prefix = TypePrefix.PROPERTY
+            else:
+                type_prefix = TypePrefix.DYNAMIC_CLASS_ATTR
+            # A property is identified using fget, fset, fdel, and doc.
+            members = [obj.fget, obj.fset, obj.fdel, obj.__doc__]
```
jqmp, line 228: Hmm, do we hash docstrings in other contexts?

bionic/code_hasher.py
```diff
@@ -219,6 +219,19 @@ def _update_hash(self, hash_accumulator, obj, code_context):
                 obj_bytes=self._check_and_hash(obj.value, code_context),
             )
 
+        elif isinstance(obj, (property, types.DynamicClassAttribute)):
+            if isinstance(obj, list):
```
jqmp, line 223: What's the relationship between being a property and being a list?

bionic/code_hasher.py
```diff
@@ -219,6 +219,19 @@ def _update_hash(self, hash_accumulator, obj, code_context):
                 obj_bytes=self._check_and_hash(obj.value, code_context),
             )
 
+        elif isinstance(obj, (property, types.DynamicClassAttribute)):
```
jqmp, line 222: When would we encounter a `types.DynamicClassAttribute`? Is there a test case for this?

bionic/code_hasher.py
namanjain: Yes 🤦🏽

bionic/code_hasher.py
namanjain: We do. It's part of `co_consts` for routines and we get it as a `member` for class types.

bionic/code_hasher.py
namanjain: This should have been a property 🤦🏽 . I'll see if I can add a test where property is replaced with dynamic class attr.

bionic/code_hasher.py
namanjain: `Enum`s use it and we hash `TypePrefix` that tests for it. But related to [this comment](https://github.com/square/bionic/pull/329#discussion_r571264494), I'll see how hard it is to create a `DynamicClassAttribute` and test it directly (like property tests).

tests/test_code_hasher.py
namanjain: Yes 🤦🏽

bionic/code_hasher.py
namanjain: ah with the current pattern, this should be `TypePrefix.Hash`. But I'm missing `add_to_hash(hash_accumulator, type_prefix=type_prefix)` before this.

bionic/code_hasher.py
namanjain: I added a test for it and it fails without fixing this bug. Good catch!

jqmp (approved):

PR #330, opened 2021-02-05
Don't suppress bytecode warnings by default

namanjain:
I also considered skipping `ResettingCallCounter` in `CodeHasher`
logic. But I think that it's cleaner to have a bunch of version
decorators with `suppress_bytecode_warnings=True` than to bleed
test behavior into the real hashing logic.

### DIFF
--- a/bionic/__init__.py
+++ b/bionic/__init__.py
@@ -1,6 +1,7 @@
 from .flow import Flow, FlowBuilder  # noqa: F401
 from .decorators import (  # noqa: F401
     version,
+    version_no_warnings,
     output,
     outputs,
     docs,
--- a/bionic/code_hasher.py
+++ b/bionic/code_hasher.py
@@ -39,7 +39,6 @@ PREFIX_SEPARATOR = b"$"
 
 # List of things we should do before releasing Smart Caching:
 # - verify that we hash all Python constant types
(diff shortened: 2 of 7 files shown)


jqmp (approved):
I agree that `suppress_bytecode_warnings=True` is better than having a special case in our hashing logic. But it does suggest that it would be useful to have a way to exclude a class from bytecode analysis. For example, it would be great if we could attach `@bn.version` to arbitrary functions and classes.

In the meantime, WDYT about defining a `@version_no_warnings` decorator for our persistence tests?

namanjain:
> I agree that `suppress_bytecode_warnings=True` is better than having a special case in our hashing logic. But it does suggest that it would be useful to have a way to exclude a class from bytecode analysis. For example, it would be great if we could attach `@bn.version` to arbitrary functions and classes.
> 
> In the meantime, WDYT about defining a `@version_no_warnings` decorator for our persistence tests?

Yeah, I agree that it would be great to register arbitrary classes and functions to be excluded from bytecode analysis. But I'm not sure I can get it done by next week (let's talk off GitHub about this) since there are a few options to do it and exploring those options might take time. I can leave a new TODO and link it to your comment.

I like the new decorator (which we can keep undocumented) that we only use for persistence tests.

PR #332, opened 2021-02-06
Hash all constant types and more

namanjain:
For Smart Caching, we want to hash all kind of constants that exists in
`func.__code__.co_consts` variable. We already hash most of them, but
with this change, we should be hashing all of the constants. I couldn't
find an official source that lists the type of constants, but I looked
at CPython impl to see the different constant kinds. Particularly, I
looked at the function that validates constant in ast which should list
all constant types. But I also found another function that creates a key
for constants (doesn't have all the constants though).

`validate_constant` in AST: https://github.com/python/cpython/blob/master/Python/ast.c#L120
`_PyCode_ConstantKey`: https://github.com/python/cpython/blob/86e322f1412be6c5a6c8fa8e73af68b99c42bb18/Objects/codeobject.c#L784

This change also hashes a couple of more types that are not constants
but can be easily hashed.


tests/test_flow/test_persistence.py
```diff
@@ -1069,11 +1069,13 @@ def ref_func():
 def test_versioning_auto_version_options(builder):
     builder.set("core__versioning_mode", "auto")
 
-    complex_number = complex(1.0, 0.0)
+    import pandas as pd
+
+    df = pd.Series([1, 2, 3])
```
jqmp, line 1074: Doesn't `df` generally mean "dataframe"?

bionic/code_hasher.py
```diff
@@ -162,11 +174,13 @@ def _update_hash(self, hash_accumulator, obj, code_context):
...
             )
 
-        elif isinstance(obj, (list, set, tuple)):
+        elif isinstance(obj, (list, set, tuple, frozenset)):
             if isinstance(obj, list):
                 type_prefix = TypePrefix.LIST
             elif isinstance(obj, set):
                 type_prefix = TypePrefix.SET
+            elif isinstance(obj, frozenset):
+                type_prefix = TypePrefix.FROZENSET
```
jqmp, line 183: Hmm... does the fact that frozensets (and sets) have non-deterministic order present any problems here?

jqmp (approved):

tests/test_flow/test_persistence.py
namanjain: You're right. I was creating a dataframe first but ended up with a series. I'll rename it.

bionic/code_hasher.py
namanjain: Ah yeah, we wanted to do it for both this and protocols but I think we forgot about this part. I have an idea on how to fix it. I can create an individual hash for each element and sort the hashes before adding it to the `hash_accumulator`. That way, the order would be guaranteed.

I'll do it in a follow up task to not block this TODO.

PR #333, opened 2021-02-06
Stable hash for set and frozenset

namanjain:


bionic/code_hasher.py
```diff
@@ -196,6 +192,25 @@ def _update_hash(self, hash_accumulator, obj, code_context):
                     obj_bytes=self._check_and_hash(elem, code_context),
                 )
 
+        elif isinstance(obj, (set, frozenset)):
+            if isinstance(obj, set):
+                type_prefix = TypePrefix.SET
+            else:
+                type_prefix = TypePrefix.FROZENSET
+            obj_bytes = str(len(obj)).encode()
```
jqmp, line 200: Wouldn't `obj_len_bytes` make more sense?

bionic/code_hasher.py
```diff
@@ -196,6 +192,25 @@ def _update_hash(self, hash_accumulator, obj, code_context):
...
+            add_to_hash(
+                hash_accumulator,
+                type_prefix=type_prefix,
+                obj_bytes=obj_bytes,
+            )
+            elem_hashes = [self._check_and_hash(elem, code_context) for elem in obj]
```
jqmp, line 206: You can also just use `sorted(self._check_and_hash(elem, code_context) for elem in obj)` here.

Also, could we add a comment explaining why this is necessary?

jqmp (approved):

Issue #331, opened 2021-02-06
Confusing error message when non-iterable input used with `outputs`


jqmp:
Sometimes users pass a single argument to `@bn.outputs`, expecting it to work like `@bn.output`:

    builder = bn.FlowBuilder('test')

    @builder
    @bn.outputs('value')
    def _():
        return 7
    
    flow = builder.build()
    
    flow.get('value')

This fails because `@outputs` expects the decorated function to return a sequence with one value. The correct thing would be to either return the tuple `7,`, or use `@bn.output` instead of `@bn.outputs`. However, the error message could be more helpful:

    EntityValueError: Value received for '<value>,' is not valid for TupleProtocol due to <class 'AssertionError'>: 'int' object is not iterable

The message is correct, but it would be better to state what was expected (a sequence of length 1). We could even go farther and say something like "did you mean to use `@output` instead of `@outputs`?". (Also, we should probably replace the `<class 'AssertionError'>` with just `AssertionError`.

PR #334, opened 2021-02-06
Clarify TupleProtocol error message

db7894:
Addressing the confusing error message issue in #331 - added a _raise_validation_exception message specific to TupleProtocol with a clearer message. Added test_non_sequence_output to test_outputs.py for thoroughness, and changed function name `three_outputs` in `test_non_sequence_outputs` to `non_sequence_output` since that function just returns a single int.

### DIFF
--- a/bionic/protocols.py
+++ b/bionic/protocols.py
@@ -83,7 +83,7 @@ class BaseProtocol:
     def _raise_validation_exception(self, descriptor_text, exception):
         message = f"""
         Value received for {descriptor_text} is not valid for
-        {self.__class__.__name__} due to {exception.__class__}:
+        {self.__class__.__name__} due to {exception.__class__.__name__}:
         {exception}
         """
         raise EntityValueError(oneline(message)) from exception
@@ -876,8 +876,16 @@ class TupleProtocol(BaseProtocol):
     def validate(self, value):
         try:
             items = tuple(value)
-        except TypeError as e:
-            raise AssertionError(str(e))
+        except TypeError:
+            message = f"""
+            Expected a sequence with length {self._expected_length};
+            instead, got non-sequence value {value!r}.
+            """
+            if self._expected_length == 1:
+                message += f"""
+                {"Did you mean to use @output instead of @outputs?"}
+                """
+            raise AssertionError(oneline(message))
         if len(items) != self._expected_length:
             message = f"""
             Expected a sequence with length {self._expected_length};
--- a/docs/release-notes.rst
+++ b/docs/release-notes.rst
@@ -107,6 +107,11 @@ Improvements
   serialized non-deterministically, which could lead to downstream values being
   spuriously recomputed. (Unfortunately this fix doesn't help with other objects that
   happen to contain sets.)
+- Clarified an `error message <https://github.com/square/bionic/issues/331>`__ that
+  occurs when users pass a single argument to @bn.outputs instead of using @bn.output.
+  TupleProtocol's validate method now provides more information about what is expected
+  and, in the case when a user passes a single argument, asks if the user meant to
+  use @bn.output instead.
 
 Bug Fixes
 .........
--- a/tests/test_flow/test_outputs.py
+++ b/tests/test_flow/test_outputs.py
@@ -126,9 +126,21 @@ def test_wrong_number_of_outputs(builder):
 def test_non_sequence_outputs(builder):
     @builder
     @bn.outputs("a", "b")
-    def three_outputs():
+    def non_sequence_output():
         return 1
 
     flow = builder.build()
     with pytest.raises(EntityValueError):
         flow.get("a")
+
+
+def test_non_sequence_outputs_message(builder):
+    @builder
+    @bn.outputs("a")
+    def non_sequence_output():
+        return 7
+
+    flow = builder.build()
+    with pytest.raises(EntityValueError) as e:
+        flow.get("a")
+        assert "did you mean to use @output instead of @outputs?" in e.value


tests/test_flow/test_outputs.py
```diff
@@ -123,10 +123,21 @@ def three_outputs():
...
+
+    flow = builder.build()
+    with pytest.raises(EntityValueError):
+        flow.get("a")
+
+
 def test_non_sequence_outputs(builder):
     @builder
     @bn.outputs("a", "b")
-    def three_outputs():
```
jqmp, line 129: Ha, nice fix.

bionic/protocols.py
```diff
@@ -885,6 +885,14 @@ def validate(self, value):
             """
             raise AssertionError(oneline(message))
 
+    def _raise_validation_exception(self, descriptor_text, exception):
+        message = f"""
+        Expected a sequence. Value is invalid for
+        {self.__class__.__name__} due to {exception.__class__.__name__}:
+        {exception}. Did you mean to use @output instead of @outputs?
```
jqmp, line 892: What do you think about, rather than overriding this method, changing line 880 to produce a message more like what's on line 883? That way, `validate` is responsible for the details of what we were expecting, and `_raise_validation_expection` is just responsible for providing context like `{descriptor_text}`.

In particular, it looks like if we throw the exception on line 886, the resulting error message will say `Expected a sequence` twice, and also say `Did you mean to use @output instead of @outputs?`, which doesn't make sense for that particular error.

tests/test_flow/test_outputs.py
```diff
@@ -123,10 +123,21 @@ def three_outputs():
...
 
 
+def test_non_sequence_output(builder):
+    @builder
+    @bn.outputs("a")
+    def non_sequence_output():
+        return 7
+
+    flow = builder.build()
+    with pytest.raises(EntityValueError):
```
jqmp, line 133: Since we're changing the error message here, we might want to check that the exception has (part of) the message we expect. (You can see an example of this in `tests/test_flow/test_api.py::test_get_modes_not_persisted`.)

tests/test_flow/test_outputs.py
```diff
@@ -123,10 +123,21 @@ def three_outputs():
         flow.get("a")
 
 
+def test_non_sequence_output(builder):
```
jqmp, line 126: nit: I feel like `test_non_sequence_output` suggests that we're testing `@output` rather than `@outputs`.

jqmp (commented):
This is looking great. I think you may want to make the core change in a slightly different place, as suggested in the comments, and I have a few nits about the tests, but otherwise I think this will be very helpful. (I've seen multiple people get tripped up by this error.)

bionic/protocols.py
db7894: Right, that's a lot cleaner - I didn't realize that bit of the error message came from 880. Fixed this in the new commit.

tests/test_flow/test_outputs.py
db7894: Addressed your comment about checking the exception message and changed the name to `test_non_sequence_outputs_message` since there's already a `test_non_sequence_outputs`.

bionic/protocols.py
db7894: Right, that's a lot cleaner - I didn't realize that bit was coming from the TypeError section in `validate` at first. I moved the clarification there like you suggested.

tests/test_flow/test_outputs.py
db7894: Changed this to `test_non_sequence_outputs_message` since there's already a `test_non_sequence_outputs`, also addressing the comment above.

bionic/protocols.py
```diff
@@ -877,7 +877,11 @@ def validate(self, value):
         try:
             items = tuple(value)
         except TypeError as e:
-            raise AssertionError(str(e))
+            message = f"""
+            Expected a sequence with length {self._expected_length};
+            did you mean to use @output instead of @outputs?
```
jqmp, line 882: Like below, let's also include some description of what we got instead, like `instead, got non-sequence value {value!r}`, to make it easier to debug.

Oh, also, I should have clarified this before, but the ` did you mean to use @output instead of @outputs?` message only makes sense if `self._expected_length` is `1`, right?

jqmp (commented):
Thanks for these changes! One more thing I should have mentioned before: we should add an entry to `docs/release-notes.rst` (under "Upcoming Version"/"Improvements") as well. (This step should really be in our Contributing to Bionic documentation. I'll try to add this soon.)

bionic/protocols.py
db7894: Ah yes, that was silly of me not to notice! I've amended my commit to fix this.

docs/release-notes.rst
```diff
@@ -93,6 +93,11 @@ Improvements
   serialized non-deterministically, which could lead to downstream values being
   spuriously recomputed. (Unfortunately this fix doesn't help with other objects that
   happen to contain sets.)
+- Clarified an `error message <https://github.com/square/bionic/issues/331>`__ that
+  occurs when users pass a single argument to @bn.outputs instead of using @bn.output.
```
jqmp, line 97: nit: we generally write decorators, etc., in monospace, like this:

    ``@bn.outputs``

jqmp (approved):
Looks good!

jqmp:
Hey @db7894, when you have a chance, could you rebase this PR on the latest master and push again? Hopefully that will get the CI tests to run.

PR #337, opened 2021-02-10
derive aip project id from environment

simonafk:


jqmp (approved):
Makes sense to me. I'm sure you already know this, but let's update the documentation PR to reflect this too.

PR #338, opened 2021-02-10
Use fsspec filesystem in LocalFilesystem

namanjain:
This is part of an ongoing effort to move all file operations to fsspec
in Bionic. There are still a lot of places that don't use fsspec and we
should migrate them too, but this is the first step towards using "file"
fsspec filesystem.

### DIFF
--- a/bionic/persistence.py
+++ b/bionic/persistence.py
@@ -5,20 +5,17 @@ point of entry is the PersistentCache, which encapsulates this functionality.
 
 import attr
 import cattr
+import fsspec
 import os
-import shutil
-import tempfile
 from typing import List, Optional, Tuple
 import yaml
 import warnings
 from uuid import uuid4
 from pathlib import Path
+from urllib.parse import unquote
 
 from .datatypes import CodeFingerprint, Artifact
-from .utils.files import (
-    ensure_dir_exists,
-    ensure_parent_dir_exists,
-)
+from .utils.files import ensure_parent_dir_exists
 from .utils.misc import hash_simple_obj_to_hex, oneline
 from .utils.urls import (
     bucket_and_object_names_from_gs_url,
@@ -664,9 +661,8 @@ class LocalStore:
         self._artifact_root_path = root_path / "artifacts"
 
         inventory_root_path = root_path / "inventory"
-        tmp_root_path = root_path / "tmp"
         self.inventory = Inventory(
-            "local disk", "local", LocalFilesystem(inventory_root_path, tmp_root_path)
+            "local disk", "local", LocalFilesystem(inventory_root_path)
         )
 
     def generate_unique_dir_path(self, provenance):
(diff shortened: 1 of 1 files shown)


bionic/persistence.py
```diff
@@ -754,43 +750,35 @@ class LocalFilesystem:
...
-        try:
-            working_path = working_dir / "tmp_file"
-            working_path.write_bytes(content_bytes)
-
-            working_path.rename(path)
+        ensure_parent_dir_exists(path_from_url(url))
 
-        finally:
-            shutil.rmtree(str(working_dir))
+        with self._fs.transaction:
```
simonafk, line 771: Wondering, why is it necessary to use a `transaction` here?

namanjain: We write the file to a temporary location and then move it to the cache location to avoid data corruption in case the write somehow failed or stopped. We can now do the same thing using `transaction` instead.

bionic/persistence.py
```diff
@@ -754,43 +750,35 @@ class LocalFilesystem:
...
-        if not path_prefix.is_dir():
-            return []
-
-        return [
-            url_from_path(path_prefix / sub_path)
-            for sub_path in path_prefix.glob("**/*")
-        ]
+        # This endpoint is a glob **/* search.
+        glob_url = unquote(url_prefix) + "**/*"
+        return ["file://" + url for url in self._fs.glob(glob_url)]
```
simonafk, line 763: Wondering, we cannot use `url_from_path` here?

bionic/persistence.py
```diff
@@ -754,43 +750,35 @@ class LocalFilesystem:
     to local disk.
     """
 
-    def __init__(self, inventory_dir, tmp_dir):
+    def __init__(self, inventory_dir):
         self.root_url = url_from_path(inventory_dir)
-        self.tmp_root_path = tmp_dir
+        self._fs = fsspec.filesystem("file")
```
simonafk, line 755: I am assuming that fsspec for "file" filesystem is able to work with urls?

simonafk (approved):

bionic/persistence.py
```diff
@@ -754,43 +750,35 @@ class LocalFilesystem:
...
 
     def exists(self, url):
-        return path_from_url(url).exists()
+        return self._fs.exists(unquote(url))
```
jqmp, line 758: I notice we're unquoting all the URLs that we pass to the filesystem, but we don't do this for the GCS filesystem. Do these two filesystems make different assumptions about their inputs?

jqmp:
Love that we're moving towards using fsspec everywhere!

bionic/persistence.py
namanjain: We'll have to convert urls to paths first but that should work too!

bionic/persistence.py
namanjain: Yes, `file` is the protocol for the `LocalFileSystem` that we use here.

bionic/persistence.py
namanjain: Yeah, gcsfs unquotes by itself. But local fs doesn't, it takes the quotes literally. So if a file has space, like "file name", `path.to_uri()` will replace the space character with `%20` and without unquoting, Bionic will treat the filename as "file%20name".

PR #340, opened 2021-02-23
Revert "Use fsspec filesystem in LocalFilesystem"

simonafk:
This reverts commit 3a7280a3b79ba629d3482fd3b057888385a2ccd1.

Use of fsspec transaction caused Read The Docs build to fail.

### DIFF
--- a/bionic/persistence.py
+++ b/bionic/persistence.py
@@ -5,17 +5,20 @@ point of entry is the PersistentCache, which encapsulates this functionality.
 
 import attr
 import cattr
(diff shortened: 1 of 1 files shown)


namanjain (approved):
The change is pretty straightforward revert, LGTM. Can you add the reason why we are reverting this change, either as part of the commit message or in the PR description?

simonafk:
Yes, already added the reason to the commit message. It looks like I have to update the PR comments manually.

PR #341, opened 2021-02-25
Add `should_persist` flag to SVG mouse hover tooltip

IDl0T:
It's great to have more info in the SVG tooltip. (`flow.render_dag()` specifically)

### DIFF
--- a/bionic/dagviz.py
+++ b/bionic/dagviz.py
@@ -154,8 +154,13 @@ def dot_from_graph(graph, vertical=False, curvy_lines=False, name=None):
                 shape="box",
             )
             if doc:
-                tooltip = rewrap_docstring(doc)
-                dot_node.set("tooltip", tooltip)
+                tooltip = (
+                    rewrap_docstring(doc)
+                    + f"\n\nPersisted: {graph.nodes[node]['should_persist']}"
+                )
+            else:
+                tooltip = f"Persisted: {graph.nodes[node]['should_persist']}"
+            dot_node.set("tooltip", tooltip)
             subdot.add_node(dot_node)
 
         dot.add_subgraph(subdot)
--- a/bionic/deriver.py
+++ b/bionic/deriver.py
@@ -152,6 +152,7 @@ class EntityDeriver:
             tasks = self._get_or_create_dinfo_for_dnode(dnode).tasks
             descriptor = dnode.to_descriptor()
             doc = self._obtain_metadata_for_dnode(dnode).doc
+            should_persist = self._obtain_metadata_for_dnode(dnode).should_persist
             should_collapse = should_collapse_dnode(dnode)
 
             for task_ix, task in enumerate(
@@ -171,6 +172,7 @@ class EntityDeriver:
                     case_key=task.key.case_key,
                     task_ix=task_ix,
                     doc=doc,
+                    should_persist=should_persist,
                 )
 
                 for dep_state in state.dep_states:
--- a/docs/release-notes.rst
+++ b/docs/release-notes.rst
@@ -68,6 +68,12 @@ For each release, we list the following types of change (in this order):
 .. Upcoming Version (Not Yet Released)
 .. -----------------------------------
 
+Improvements
+............
+
+- Added a line to the SVG DAG visualization mouse hover tooltips indicating whether 
+  an entity is persisted.
+
 .. Record any notable changes in this section. When we update the current version,
    add a new version heading below, and then comment out the heading above until more
    changes are added. This way, the "Upcoming Version" section will be never be visible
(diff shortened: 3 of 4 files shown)


bionic/dagviz.py
```diff
@@ -155,6 +155,7 @@ def doc_from_node(node):
             )
             if doc:
                 tooltip = rewrap_docstring(doc)
+                tooltip += ", persist = " + str(graph.nodes[node].get("should_persist", ""))
```
jqmp, line 158: The `doc` string will probably be in complete sentences, so attaching a comma to the end may not look right. Maybe something like this?

    rewrap_docstring(doc) + f"\n\n Persisted: {should_persist}"

Also, what should we do when `doc` is not present?

bionic/deriver.py
```diff
@@ -480,7 +482,7 @@ def _obtain_metadata_for_dnode(self, dnode):
                 self._core.should_memoize_for_query_if_uncached
                 and not (should_memoize or should_persist)
             )
-
+
```
jqmp, line 485: Let's not add this extra whitespace.

jqmp (commented):
This is looking good! Aside from the formatting changes I've suggested, we just need a few more things to make this PR complete:

1. You'll probably need to update the tests to make sure they pass. I'm guessing `test_dot_tooltips` will need updating?
2. Let's add a note to `docs/release-notes.rst` describing this change.
3. I notice you have an extra merge commit. Could you rebase your branch to get rid of it?

Thanks for making this PR! I think it'll be very helpful.

IDl0T:
@jqmp I have made some changes according to your suggestions. The only problem might be the weird commit history. I tried to rebase but somehow I felt my HEAD pointer isn't working. `git show HEAD^` gave the intended commit while `git reset HEAD^` and `git rebase -i HEAD^` led me to wild places. I might need to review some git tutorials! :grimacing:

bionic/dagviz.py
```diff
@@ -154,7 +154,11 @@ def doc_from_node(node):
                 shape="box",
             )
             if doc:
-                tooltip = rewrap_docstring(doc)
+                tooltip = (rewrap_docstring(doc) + 
+                           f"\n\n Persisted: {graph.nodes[node].get('should_persist', '')}")
+                dot_node.set("tooltip", tooltip)
```
jqmp, line 159: Do we need to duplicate this line in both branches of the if statements?

bionic/dagviz.py
```diff
@@ -154,7 +154,11 @@ def doc_from_node(node):
                 shape="box",
             )
             if doc:
-                tooltip = rewrap_docstring(doc)
+                tooltip = (rewrap_docstring(doc) + 
+                           f"\n\n Persisted: {graph.nodes[node].get('should_persist', '')}")
```
jqmp, line 158: Is it possible for the `should_persist` field to be missing? If so, when? If not, why do we need to call `get` with a default value?

jqmp, line 158: Why the extra space between `\n\n` and `Persisted`?

docs/release-notes.rst
```diff
@@ -65,8 +65,13 @@ For each release, we list the following types of change (in this order):
...
 
-.. Upcoming Version (Not Yet Released)
-.. -----------------------------------
+Upcoming Version (Not Yet Released)
+-----------------------------------
+
+Improvements
+............
+
+- Add a `should_persist` flag to the SVG mouse hover tooltip of `flow.render_dag()`
```
jqmp, line 74: These notes are intended for the user, so we should describe the improvement in terms of what they'll see. So in this case we might write "Added a line to the SVG DAG visualization tooltips indicating whether an entity is persisted." (Instead of mentioning `should_persist`, which is an internal variable name.)

jqmp (commented):
I'm not sure what your local repository looks like, but I wonder if the following would work for you?

    git rebase -i origin/master

Hopefully this will show you just your two commits. Then you can replace `pick` with `squash` on the second one to squash them together.

FWIW, I usually find it easier to work on [feature branches](https://www.atlassian.com/git/tutorials/comparing-workflows/feature-branch-workflow) rather than on `master`.

docs/release-notes.rst
IDl0T: That makes a lot of sense. Thanks for explaining all these things!

IDl0T:
@jqmp Huge thanks for you explaining all these things and being so kind and patient! With your advice, I've successfully resolved all issues mentioned and learned a lot about developing. Let's see if this PR looks good or not.

jqmp (approved):
This looks good! Thanks for bearing with all my comments. I'm going to approve it, but I also need to make sure the automated tests run before we merge it. I'm not sure why they weren't triggered automatically, but I'll look into it.

jqmp:
Looks like the failing tests here are not related to this PR, but are caused by not having access to GitHub secrets in PR-triggered CI actions. That should be fixed in https://github.com/square/bionic/pull/350. Meanwhile I'm going to go ahead and merge this PR.

PR #342, opened 2021-02-27
Suppress import warnings

jqmp:
Bionic was wasn't respecting the `suppress_warnings` flag when it
encountered a dynamic import during automatic versioning. This fixes it.

### DIFF
--- a/bionic/code_references.py
+++ b/bionic/code_references.py
@@ -206,18 +206,21 @@ def get_referenced_objects(code, context, suppress_warnings=False):
                 # module using the import statement. If a user is importing
                 # modules inside a function, they probably don't want to import
                 # the module until the function execution time.
-                message = f"""
-                Entity function in file {code.co_filename} imports the
-                '{op.argval}' module at line {lineno};
-                Bionic will not be able to automatically detect any changes to this
-                module.
-                To enable automatic detection of changes, import the module at the
-                global level (outside the function) instead.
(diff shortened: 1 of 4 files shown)


bionic/code_references.py
```diff
@@ -167,11 +179,7 @@ def set_tos(t):
                     Please raise a new issue at
                     https://github.com/square/bionic/issues to let us know.
                     """
-                message += """
-                You can also suppress this warning by removing the
-                `suppress_bytecode_warnings` override from the
-                `@version` decorator on the corresponding function.
-                """
+                message += make_suppression_advice(code.co_name)
```
namanjain, line 182: This will give you the function name of whatever `code` is, which can be a non-bionic entity function. The advice suggests the user to change the `@version` decorator on the function which won't work for non-bionic entity functions. If you still want to show the function name, you will have to plumb the top level function name here.

namanjain (commented):
Nice catch! I left a comment about the function name where we suggest changing the `@version` decorator field, but it LGTM otherwise.

bionic/code_references.py
jqmp: I updated the message to make it clear that the named function is not necessarily the same as the top-level function. @namanjain WDYT?

namanjain: LGTM

namanjain (approved):

PR #343, opened 2021-02-27
Always emit versioning warnings when appropriate

jqmp:
It looks like we were defaulting to emitting versioning warnings for
functions decorated with `@version`, but not for undecorated functions.
This commit should make us always default to emitting warnings (unless
auto/assist versioning is not enabled).

To do this, I changed the `@version` decorator to defer to the defaults
of the CodeVersion and CodeVersioningPolicy constructors instead of
trying to apply its own defaults; that way they should always stay in
sync.

I also had to apply @version_no_warnings` to a bunch more functions in
our tests.

@namanjain could you confirm that I haven't done something dumb here?

### DIFF
--- a/bionic/datatypes.py
+++ b/bionic/datatypes.py
@@ -327,7 +327,7 @@ class CodeVersioningPolicy:
 
     version: CodeVersion = attr.ib()
     suppress_bytecode_warnings: bool = attr.ib(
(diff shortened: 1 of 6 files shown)


namanjain (approved):
Nope, nothing stupid. This looks great! Thanks for fixing it.

PR #347, opened 2021-03-08
hasher fix str encoding

simonafk:
utf-8 string encoding may cause errors. Avoid the error by switching to
a different encoding.


damienrj (approved):
This looks like this should solve the problem we were having

jqmp:
Looks reasonable, but do we need to bump the cache schema version? Some strings will hash to different values with this new encoding, right?

simonafk:
> Looks reasonable, but do we need to bump the cache schema version? Some strings will hash to different values with this new encoding, right?

Makes sense. Updated schema version.

jqmp (approved):
Looks good, except for the failing test (did you delete the test fixtures?) -- and don't forget the release notes!

simonafk:
> Looks good, except for the failing test (did you delete the test fixtures?) -- and don't forget the release notes!

Ah. I deleted some old test data files, but did not add the new ones. Added them.

PR #348, opened 2021-03-09
skip git check if aip test is skipped

simonafk:
test_aip_with_docker_build checks if the current git commit is
pushed to the remote repo. This change would skip the check
if aip is not enabled.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,9 @@
+import os
+
 import pytest
+from git import Repo
+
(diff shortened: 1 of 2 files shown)


jqmp (approved):
Makes sense. Thanks for adding this fix! Hopefully CI runs will work on external PRs now.

jqmp:
I'm going to go ahead and merge this to see if it fixes https://github.com/square/bionic/pull/344.

PR #349, opened 2021-03-13
Ignore __module__ field on classes when hashing

jqmp:
When computing the hash a class, we now ignore the `__module__` field.
This field usually contains the name of the module the class was defined
in, but if that module was run directly from the command line then it
contains `__main__` instead. That means we can get different values
depending on whether we run a python module directly or load it from
somewhere else (like a Jupyter notebook), causing our smart caching to
detect a spurious change.

### DIFF
--- a/bionic/code_hasher.py
+++ b/bionic/code_hasher.py
@@ -418,7 +418,9 @@ class CodeHasher:
             for (m_name, m_value) in inspect.getmembers(obj)
             if not (
                 inspect.isgetsetdescriptor(m_value)
-                or m_name in {"__class__", "__dict__", "__members__"}
+                # We ignore __module__ because it can vary depending on whether
+                # that class's module was run as __main__.
+                or m_name in {"__class__", "__dict__", "__members__", "__module__"}
             )
         ]
         add_to_hash(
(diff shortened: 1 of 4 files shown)


jqmp:
@namanjain WDYT?

bionic/code_hasher.py
```diff
@@ -418,7 +418,9 @@ def _update_hash_for_members_of_obj(self, hash_accumulator, obj):
             for (m_name, m_value) in inspect.getmembers(obj)
             if not (
                 inspect.isgetsetdescriptor(m_value)
-                or m_name in {"__class__", "__dict__", "__members__"}
+                # We ignore __module__ because it can vary depending on whether
+                # that class's module was run as __main__.
+                or m_name in {"__class__", "__dict__", "__members__", "__module__"}
```
simonafk, line 423: Does that mean that, if a function is moved to a different file, but the implementation remains the same, the hash remains the same?

simonafk: Hmm, `__code__.co_filename` would change if the location of the function changes.

simonafk (approved):

bionic/code_hasher.py
jqmp: This change is only for classes, and yes, I think this means that if you move a class to a different file, you'll get the same hash. I think functions already behave this way, because we don't (AFAICT) include `__code__.co_filename` in the hash.

In general it makes sense to me that we wouldn't care which file a function/class was in when hashing it.

PR #350, opened 2021-03-14
Skip GCS tests when secrets not available

jqmp:
GitHub secrets are not available in CI actions triggered by pull
requests (presumably because PRs can be started by people outside the
project), so those actions can't run tests that require GCS. This commit
should disable those tests when the secrets aren't available.


simonafk (approved):
Makes sense

PR #351, opened 2021-03-21
S3 Migration Draft

db7894:
Drafting migration from GCS to S3


simonafk:
Interesting! From the "Draft" in the title, I am assuming you are still working on this, and making it usable on both s3 and gcs?

jqmp:
> Interesting! From the "Draft" in the title, I am assuming you are still working on this, and making it usable on both s3 and gcs?

Yes, the goal is for this to work on both S3 and GCS. I suggested to @db7894 that he create this draft PR so I could help him debug an issue he's having (which I unfortunately haven't gotten to yet).

PR #352, opened 2021-03-22
Bump version: 0.10.0 → 0.10.1

simonafk:


docs/release-notes.rst
```diff
@@ -74,10 +74,14 @@ Upcoming Version (Not Yet Released)
...
    "latest" docs (corresponding to the master branch).
 
+0.10.1 (Mar 22, 2021)
+---------------------
+
 Bug Fixes
 .........
 
-- Fixes an issue where automatic versioning erroneously treated classes as
+- Fixed an issue with hashing of strings containing invalid unicode.
```
jqmp, line 83: If we changed the cache schema version, then shouldn't we have a section on "Cache Compatibility Changes"? Also, then this would need to be 0.11, right?

jqmp (commented):
Looks good except for the cache versioning issue. Also, I just merged one last caching change -- if you could squeeze it into this release I'd appreciate it!

docs/release-notes.rst
simonafk: Good catch. Bumping to 0.11.0 instead of 0.10.1. Closing this PR and opened https://github.com/square/bionic/pull/352

PR #356, opened 2022-11-03
Exclude pyrsistent version 0.19.1 from deps

namanjain:
`Pyrsistent` 0.19.1 had a regression in 0.19.1 which was fixed in 0.19.2 
The regression was on pmap where inserted elements were unreliably
inserted. See https://github.com/tobgu/pyrsistent/issues/263

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,6 +23,7 @@ requirements = [
     "pandas",
     "pyarrow",
     "pyrsistent!=0.19.1",
+    "decorator<5",
 ]
 
 setup(
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,7 @@ setup(
     include_package_data=True,
     install_requires=requirements,
     extras_require=extras_require,
-    python_requires=">=3.6",
+    python_requires=">=3.7",
     zip_safe=False,
     keywords="bionic",
     classifiers=[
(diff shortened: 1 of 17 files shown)


setup.py
```diff
@@ -22,7 +22,8 @@
     "numpy",
     "pandas",
     "pyarrow",
-    "pyrsistent",
+    "pyrsistent!=0.19.1",
+    "decorator<5",
```
jqmp, line 26: Could we add a comment explaining the conflict here?

jqmp (approved):
Thanks for cleaning up all the 3.6 stuff! (BTW, I assume no one at Square still depends on 3.6? I think we originally supported it because we needed it for AI Platform or something.)

andrewblane (approved):
LGTM, though I think the package version needs to be bumped.

namanjain:
> Thanks for cleaning up all the 3.6 stuff! (BTW, I assume no one at Square still depends on 3.6? I think we originally supported it because we needed it for AI Platform or something.)

Yes, I asked in internal slack and 3.6 is not used anymore. Existing pipelines that still are on 3.6 will continue to use the older versions of bionic.

namanjain:
> LGTM, though I think the package version needs to be bumped.


Yes, I'll open a new PR for bumping version following the release process https://bionic.readthedocs.io/en/stable/maintaining.html?highlight=release#release-process

